PR #5, opened 2020-07-27
Add a first pipeline test

rabernat:
This PR introduces a simple test pipeline.

The test pipeline class is defined as `Pipeline(AbstractPipeline, XarrayPrefectPipelineMixin)`. Instantiating it  currently failing with the following error.

```python-traceback
>       pipeline = Pipeline(name, cache_dir, target_dir, concat_dim, files_per_chunk, url_base, nfiles)
E       TypeError: Can't instantiate abstract class Pipeline with abstract methods run
```

I suspect this is related to how we are doing inheritance / mixins.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ prefect
 xarray
 zarr
 fsspec
+netCDF4
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,4 +6,3 @@ prefect
 xarray
 zarr
 fsspec
-netCDF4
--- a/tests/test_pipeline.py
+++ b/tests/test_pipeline.py
@@ -5,9 +5,9 @@ import subprocess
 import pytest
 import numpy as np
 import xarray as xr
-import pandas as pd
 import fsspec
 import time
+import pandas as pd
 
 # classes tested here
 from pangeo_forge.pipeline import AbstractPipeline, XarrayPrefectPipelineMixin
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -6,3 +6,4 @@ prefect
 xarray
 zarr
 fsspec
+netCDF4
```
rabernat, line 9: Why did this not work? It is not getting installed.

tests/test_pipeline.py
```diff
@@ -78,3 +89,52 @@ def test_fixture_http_files(daily_xarray_dataset, netcdf_http_server):
     open_files = [fsspec.open(url).open() for url in urls]
     ds = xr.open_mfdataset(open_files, combine="nested", concat_dim="time")
     assert ds.identical(daily_xarray_dataset)
+
+
+# a pipeline to load that data
+
+
+class Pipeline(AbstractPipeline, XarrayPrefectPipelineMixin):
```
jhamman, line 97: For reasons I don't entirely understand, the order of the inheritance needs to be reversed here.

rabernat:
Changing the order in the Pipeline class helped me move forward. Now I'm getting an error

```
>   cached_sources = [self.download(source_url) for k in self.sources]
E   NameError: name 'source_url' is not defined
```

I think that the idea of mixing prefect `@task` annotations with a class is causing some issues.

jhamman:
@rabernat - I'm confused. Where is the line that is failing? In pipeline.py, I'm seeing that line as:

https://github.com/pangeo-forge/pangeo-forge/blob/68c5164608115f4b02c55d0f53f68960b39a67e9/pangeo_forge/pipeline.py#L74

rabernat:
Edited to fix typo. I was messing around with some different ways to fix this and pasted the wrong thing. The full traceback is now above.

rabernat:
Do you want me to try to refactor `pipeline.py` to make this work? It probably involves breaking the tasks out into standalone functions... Or did you have other plans?

jhamman:
@rabernat - yes, looks like we may have gone down the wrong path. Let's take a step back and try refactoring to use explicit tasks (functions) rather than class methods.

rabernat:
Ok so this PR is now up to the error that I get locally. Do you want to take it from here? If you merge this, you can basically write to the test.

PR #8, opened 2020-07-29
refactor pipelines and tasks to use functions defined outside the pipeline

jhamman:
Here's my attempt at a pipeline refactor. The main changes are:

1. Separate the `Task` and `Pipeline` definitions. The way to think of these two object types is now - tasks are low level building blocks for flows defined in the `Pipeline.flow` property.
2. Establishes a package layout of `tasks` and `pipelines`. Just a few simple example tasks/pipelines for now but we can build incrementally.


pangeo_forge/pipeline.py
```diff
@@ -1,94 +0,0 @@
...
-        pass
-
-    @property
-    @abstractmethod
-    def targets(self) -> List[str]:
-        pass
-
-    @abstractmethod
-    def run(self) -> None:
-        pass
```
jhamman, line 25: this was moved to `pipelines/base.py`

pangeo_forge/pipeline.py
```diff
@@ -1,94 +0,0 @@
...
-
-
-class XarrayPrefectPipelineMixin:
```
jhamman, line 28: this was moved to the `http_xarray_zarr.py` pipeline module.

pangeo_forge/tasks/http.py
```diff
@@ -0,0 +1,36 @@
...
+    Returns
+    -------
+    target_url : str
+        Path or url in the form of `{cache_location}/hash({source_url})`.
+    """
+    target_url = os.path.join(cache_location, str(hash(source_url)))
+
+    # there is probably a better way to do caching!
+    try:
+        fsspec.open(target_url).open()
```
rabernat, line 28: This is nothing specifically http about this. It would work with FTP or even a local path. So maybe the module name should be revised?

jhamman: Let's revisit this down the road.

PR #14, opened 2020-10-14
Refactor storage, environment, CLI

TomAugspurger:
This is a batch of changes with the goal of simplifying recipe writing.

* Handle flow environment and storage outside of the recipe files.
* Moved to `GitHub` storage. Flows are now stored as files and loaded
  into the environment.
* Moved `Environment` to the base pipeline

See https://github.com/TomAugspurger/example-pipeline/blob/main/recipe/pipeline.py
for an example pipeline run with this setup. The primary change is the
*requirement* that the top-level of the module contains a `Flow`
instance named `flow`.

Additionally, the `pangeo-forge` CLI is updated to enable running flows
with the new structure. The CLI handles

1. Linting recipe pipelines.
2. Attaching Prefect environments and Storage to `Flow`s.
3. Registering flows with prefect.


pangeo_forge/pipelines/job.yaml
```diff
@@ -0,0 +1,58 @@
...
+  template:
+    metadata:
+      labels:
+        app: prefect-dask-job
+    spec:
+      serviceAccountName: pangeo-forge
+      serviceAccount: pangeo-forge
+      containers:
+        - name: flow
+          image: pangeoforge/noaa-oisst-avhrr
```
rabernat, line 17: How do we get around the hard-coding of this image name? Can we set these things on the fly when we run the flow?

pangeo_forge/cli.py
```diff
@@ -1,12 +1,47 @@
...
+@click.group()
 @click.version_option()
 def main():
-    """Pangeo-Forge: A tool for building and publishing analysis ready datasets.
+    pass
+
+
+@click.command()
+@click.argument("pipeline", type=click.Path(exists=True))
+def lint(pipeline):
```
rabernat, line 14:
```suggestion
def check(pipeline):
```

I feel like "lint" is a very developer specific term. We want to make this tool as accessible as possible to data managers.

pangeo_forge/cli.py
```diff
@@ -1,12 +1,47 @@
...
+    """
+    Check that the pipeline definition is valid. This does not run the pipeline.
     """
-    click.secho("Hello World!", fg="green")
+    result = runpy.run_path(pipeline)
+    missing = [key for key in ["pipeline", "flow"] if key not in result]
```
rabernat, line 19: Can we have a comment to explain what this line is doing?

pangeo_forge/cli.py
```diff
@@ -1,12 +1,47 @@
...
+    if missing:
+        click.echo(f"missing {missing}")
+    pipe = result["pipeline"]
+
+    pipe.flow.sorted_tasks()
+    pipe.flow.environment
+    pipe.flow.storage
+    pipe.flow.validate()
+    print("ok!")
```
rabernat, line 28:
```suggestion
    click.echo("ok!")
```

pangeo_forge/cli.py
```diff
@@ -1,12 +1,47 @@
...
+
+
+@click.command()
+@click.argument("pipeline", type=click.Path(exists=True))
+def register(pipeline):
```
rabernat, line 33: Should there be an option to specify _where_ to register the pipeline? In my head, this is separate from the pipeline itself.

rabernat (commented):
A few more random comments.

pangeo_forge/pipelines/job.yaml
TomAugspurger: OK, we're no longer hardcoding the image here. We set the image at https://github.com/pangeo-forge/pangeo-forge/pull/14/files#diff-467822c6f6378f68bea635c429827a2caf36c7f16cb25944cc7b5146262cf35aR105. I'm hopeful that we can layer on additional packages if you just need to `pip install` one more thing.

TomAugspurger:
@cicdw a prefect bug report / question for you if you have time (sorry for the length).

I'm using https://github.com/TomAugspurger/example-pipeline/blob/main/recipe/pipeline.py as an example flow. The basic setup is

1. Recipe writer inherits from `pangeo_forge.Pipeline` and writes a `flow` method that returns a `prefect.Flow` instance.
2. We (the pangeo-forge maintainers) register the flow at https://github.com/pangeo-forge/pangeo-forge/pull/14/files#diff-cd87ca8d9d81a5d90e31149d2e2f4d9c64f110a6770b4807abffff1ee60aa659R33.

*Ideally* recipe writers don't have to worry about the `Environment` and `Storage`, I'd like to handle that for them. So they just have

```python
class Pipeline:
    @property
    def flow(self):
        with Flow(self.name) as flow:
          ...
        return flow

pipeline = Pipeline()
flow = pipeline.flow()
```

And then in our `pangeo-forge` CLI we set the `.environment` and `.storage` on the `flow`

```
    env = runpy.run_path(pipeline)
    flow = env["flow"]
    pipe = env["pipeline"]
    flow.environment = pipe.environment
    flow.storage = pipe.storage
    flow.register(project_name="pangeo-forge", labels=["gcp"]) 
```

However, when I do that the custom job / scheduler template we use (https://github.com/pangeo-forge/pangeo-forge/pull/14/files#diff-267b30d97c826b0afcae2110fe8ca4acfe6f35a6321d80f5fcc74ea9b7547fc0) doesn't take effect. The pod is run with the `default` ServiceAccount (as an aside, `serviceAccount` is the *only* thing we need to override. Would it be possible to specify that as a keyword to `DaskKubernetesExecutor`? :smile: )

The workaround isn't too bad. The recipe writer just needs to pass through the keywords themselves

```python
    @property
    def flow(self) -> Flow:
        with Flow(
            self.name, environment=self.environment, storage=self.storage
        ) as flow:
           ...
        return flow
```

But in general, should the following two be equivalent

```python
# case 1
>>> flow = Flow(..., environment=environment)
# case 2
>>> flow = Flow(...)
>>> flow.environment = environment
# followed by
>>> flow.register(....)
```

As I write those two cases, I might spot the difference with what we're doing. The module where the flow is defined never has the storage / environment set. It's done dynamically in a separate process. Is that perhaps the relevant difference? I don't know if you have access to it, but https://cloud.prefect.io/pangeo-forge/flow/a6db8513-5b40-4258-a294-046d8e2883be?version=25 is a flow run using `Flow(self.name)` where the default service account was incorrectly used.

cicdw:
Hey @TomAugspurger I think I know what's going on.  Here's the high level of what appears to be happening:

- when you register the Flow, you tell Cloud where the flow lives (that's what `Storage` is all about).  In your case, you are using `GitHub` storage, and you tell Cloud the flow lives [here](https://github.com/TomAugspurger/example-pipeline/blob/main/recipe/pipeline.py)
- when your agent goes to run the flow, it receives the info of where to retrieve the flow from, and in this case, it re-runs that script and grabs the `Flow` object out of it (which doesn't have an environment on it)

When you attach the environment to the Flow, that metadata _is_ sent to Cloud for the purposes of displaying in the UI, but it's not enough information to recreate the whole environment class you have configured. Instead, the agent relies on the `.environment` that is actually attached to the flow.  

We're refactoring the environment interface to hopefully be more streamlined and allow you to more easily configure dynamic variables; it won't 100% solve this issue, but it will hopefully make the pattern more clear.

TomAugspurger:
Great, thanks, I think that makes sense. So the Python module (as exists on GitHub) should have the environment set. I already don't like the idea of users having to have

```python
# file: recipe/pipeline.py
# ...
pipeline = Pipeline(...)
flow = pipeline.flow
```

at the bottom of their pipeline file. As an alternative, *we* could generate the module.

```python
import recipe.pipeline

pipeline = Pipeline(...)
flow = pipeline.flow
flow.environment = ...
```

I'll play around with this a bit.

TomAugspurger:
Well, it took a few tries (now on version 37) but, the latest here works with https://github.com/tomaugspurger/example-pipeline.

I got my goal of a "simple" flow definition: https://github.com/TomAugspurger/example-pipeline/blob/59d94ba561ad29de4747f6c7d44ceded2312fc87/recipe/pipeline.py#L31-L37. Users don't need to care about `.storage` or `.pipeline`. The costs to get that were

1. We dynamically generate a `run.py` at https://github.com/pangeo-forge/pangeo-forge/pull/14/files#diff-467822c6f6378f68bea635c429827a2caf36c7f16cb25944cc7b5146262cf35aR124-R142. That's what prefect gets.
2. That `run.py` needs to be committed to https://github.com/TomAugspurger/example-pipeline/blob/main/recipe/run.py. It includes a full copy of `pipeline.py`, which gives a chance for these to get out of sync. We'll need to rely on CI to ensure they don't drift.

At this point I'm 50/50 on whether generating a `run.py` is a good idea or not. We could just require that the users do that in their recipe. Quite a bit simpler. We just might have to explain "why do I need this at the bottom of my script".

pangeo_forge/pipelines/base.py
```diff
@@ -119,3 +120,23 @@ def storage(self) -> Storage:
...
+
+            {source}
+
+            # ----------------------------------------------------------------
+            pipe = {name}()
+            flow = pipe.flow
+            flow.storage = pipe.storage
+            flow.environment = pipe.environment
+            """
+        ).format(source=source, name=name)
```
rabernat, line 142: This feels a bit hacky, but I understand why it is needed. Hopefully upstream changes in prefect could eventually make it unnecessary.

TomAugspurger:
I'm going to self-merge this and push on things for a bit if that's OK. I'd like to try out some stuff around debugging / workflows that will build on top of it. It'll be easy enough to revert if people have other directions they'd like to pursue.

FYI, if we're mirroring the conda-forge model then the CLI would eventually be called `pangeo-smithy` (though we can easily develop it here if that's easier, fewer repositories. We can wait to hear from @jhamman before moving it if that's what we want to do).

rabernat:
Thanks a lot for pushing forward with this Tom. Let's discuss more at the Oct. 26 call.

PR #16, opened 2020-10-22
Update tasks for noaa-oisst-avhrr pipeline

TomAugspurger:
A few updates to the task helpers.

### DIFF
--- a/pangeo_forge/tasks/xarray.py
+++ b/pangeo_forge/tasks/xarray.py
@@ -6,7 +6,9 @@ from prefect import task
 
 
 @task
-def combine_and_write(sources: List[str], target: str, concat_dim: str) -> List[str]:
+def combine_and_write(
+    sources: List[str], target: str, append_dim: str, concat_dim: str
+) -> List[str]:
     """
     Write a batch of intermediate files to a combined zarr store.
 
@@ -16,6 +18,8 @@ def combine_and_write(sources: List[str], target: str, concat_dim: str) -> List[
         A list of URLs pointing to the intermediate files.
     target : str
         The URL for the target combined store.
+    append_dim : str
+        Name of the dimension of which datasets should be appended during write.
     concat_dim : str
         The dimension to concatenate along.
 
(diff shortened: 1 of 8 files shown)


pangeo_forge/tasks/xarray.py
```diff
@@ -1,43 +1,69 @@
+from typing import List
+
 import fsspec
 import xarray as xr
 from prefect import task
 
 
 @task
-def combine_and_write(sources, target, append_dim, concat_dim, first=True):
+def combine_and_write(sources: List[str], target: str, concat_dim: str) -> List[str]:
```
TomAugspurger, line 9: Ah... I see I've mixed together `concat_dim` and `append_dim`. I'll look into why they were separate.

pangeo_forge/utils.py
```diff
@@ -9,3 +12,17 @@ def chunked_iterable(iterable, size):
         if not chunk:
             break
         yield chunk
+
+
+@task
```
TomAugspurger, line 17: This being a task makes it play with with `mapped` inputs, see https://github.com/pangeo-forge/noaa-oisst-avhrr-feedstock/pull/3/files#diff-c1ba9cb202886813ca3e7aa5645ec022ebf0588625b8411e5ed004e81e1b707aR117-R121.

I suspect that using `mapped` inputs will be pretty common.

pangeo_forge/tasks/zarr.py
```diff
@@ -1,17 +1,23 @@
...
     """
-    Consolidate Zarr metadata
+    Consolidate the metadata the Zarr group at `target`.
 
     Parameters
     ----------
     target : str
-         Path or url of the Zarr store.
+        The URL for the (combined) Zarr group.
+    writes : list of strings, optional
```
TomAugspurger, line 17: I'm not sure how prefect runs through it's DAG, but `consolidate_metadata` didn't have a direct dependency on the previous writes, which would maybe(?) let it run before the writes.

pangeo_forge/tasks/xarray.py
```diff
@@ -1,43 +1,73 @@
...
+        A list of URLs pointing to the intermediate files.
     target : str
-        Path or url to the target location of the Zarr store.
+        The URL for the target combined store.
     append_dim : str
         Name of the dimension of which datasets should be appended during write.
     concat_dim : str
-        Name of the dimension of which datasets should be concatenated during read.
-    first : bool
-        Flag to indicate if the target dataset should be expected to already exist.
```
TomAugspurger, line 23: This change is also to help out with mapped inputs. It's not clear to me yet if it's a good one, since we're relying on the presence of empty directories to indicate that this is the first write. That seems like a recipe for a race condition.

If possible, it'd be best to have a task explicitly for writing whatever needs to be written on the first write (metadata?) so that all the actual data writes are appends. I'll look into this more later.

PR #17, opened 2020-10-22
Handle environment / storage in Pipeline

TomAugspurger:
This avoids the need for a separate `run.py` file by setting the
environment and storage for the user without telling them.

cc @rabernat (no metaclasses needed!)


TomAugspurger:
Going to continue to merge this stuff. I'll have a demo ready in example pipelines later in the day, so we can evaluate where things are at..

PR #27, opened 2020-12-02
Total refactor

rabernat:
This is the beginning of a complete rewrite of pangeo forge, learning from our initial steps.

The core of the new code is a `Recipe` class, which is used like this:

```python
r = recipe.NetCDFtoZarrSequentialRecipe(
    input_urls=netcdf_local_paths,
    sequence_dim="time",
    inputs_per_chunk=1,
    nitems_per_input=daily_xarray_dataset.attrs["items_per_file"],
    target=tmp_target,
    input_cache=tmp_cache,
)

# manual execution of a recipe
r.prepare()
for input_key in r.iter_inputs():
    r.cache_input(input_key)
for chunk_key in r.iter_chunks():
    r.store_chunk(chunk_key)
r.finalize()
```

The idea is that we will then implement a `RecipeExecutor` class which runs these recipes. This will work because the methods `prepare`, `cache_input`, `store_chunk` and `finalize` are not regular methods. The are actually properties that return serializable functions.

There is significant work to do here, but I feel pretty good about where this is going...


rabernat:
Black and flake8 both pass on my local env but are failing in this action. Not what I want to be debugging right now.

pangeo_forge/recipe.py
```diff
@@ -0,0 +1,320 @@
...
+                # actually not necessary if we use decode_times=False
+                self.initialize_target(ds)
+
+            self.expand_target_dim(self.sequence_dim, self.sequence_len())
+
+        return _prepare
+
+
+@dataclass
+class StandardSequentialRecipe(
```
jhamman, line 313: Let's think about a more informative name here. Once concern I have is that we'll get to Xarray->Zarr focused. While I'm quite happy to support that workflow as the primary and initial implementation, I want to make sure we leave room for a Rasterio->COG workflow or Pandas->Parquet workflow.

pangeo_forge/storage.py
```diff
@@ -0,0 +1,65 @@
+import os
+from contextlib import contextmanager
+from dataclasses import dataclass
+
+import fsspec
+
+
+@dataclass
+class Target:
```
jhamman, line 9: Maybe this should be called a `ZarrTarget`?

pangeo_forge/storage.py
```diff
@@ -0,0 +1,65 @@
...
+        class.
+    prefix : str
+        A path prepended to all paths.
+    """
+
+    fs: fsspec.AbstractFileSystem
+    prefix: str = ""
+
+    def _full_path(self, path):
+        return os.path.join(self.prefix, _hash_path(path))
```
jhamman, line 50: For what its worth, I've found that hashing paths like this can make it difficult to debug failed workflows. Maybe you can explain a bit more why we need to hash all paths like this?

pangeo_forge/recipe.py
```diff
@@ -0,0 +1,320 @@
...
+        dsets = [self.open_input(i) for i in inputs]
+        # CONCAT DELETES ENCODING!!!
+        ds = xr.concat(dsets, self.sequence_dim, **self.xarray_concat_kwargs)
+        logger.debug(f"{ds}")
+
+        # do we really want to just delete all encoding?
+        # for v in ds.variables:
+        #    ds[v].encoding = {}
+
+        # TODO: maybe do some chunking here?
```
jhamman, line 173: can we make these options? I think that yes, we generally want to remove netcdf encoding before writing to zarr. But there are probably cases where this isn't true.

jhamman (commented):
I'm still catching up here so I've mostly asked questions to improve my understanding. Generally, I'm a fan of the redesign with my only concerns being how we leave space for extension to pipelines that may not use xarray or zarr.

rabernat:
Thanks a lot Joe! I'll first reply to some of your questions and then update my PR in response to your comments.

pangeo_forge/recipe.py
```diff
@@ -0,0 +1,320 @@
...
+from typing import Callable, Iterable
+
+import fsspec
+import xarray as xr
+import zarr
+
+from .storage import InputCache, Target
+from .utils import chunked_iterable, fix_scalar_attr_encoding
+
+# logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
rabernat: It does work. I'll put it back

pangeo_forge/recipe.py
```diff
@@ -0,0 +1,320 @@
...
+    #
+    #     def _finalize():
+    #         pass
+    #     return _finalize
+
+
+# Notes about dataclasses:
+# - https://www.python.org/dev/peps/pep-0557/#inheritance
+# - https://stackoverflow.com/questions/51575931/class-inheritance-in-python-3-7-dataclasses
+# This means that, for now, I can't get default arguments to work.
```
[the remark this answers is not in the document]
rabernat: Dataclasses reduce the amount of boilerplate we have to write / maintain. None of these classes needs init methods.

I believe we can fix the default arguments problem by tweaking the mixin order. This requires me to understand python's method resolution order 🤯.

pangeo_forge/storage.py
rabernat: Maybe `MapperTarget`? There is nothing zarr specific about it...

pangeo_forge/storage.py
rabernat: I agree. My thinking was: I want there to be unique mapping between inputs and paths in the cache. Hashing achieves this. The input paths may be urls with lots of forbidden characters in them. But I'll play with some alternatives that are more readable / debuggable.

pangeo_forge/recipe.py
rabernat: Yes agreed.

rabernat:
In retrospect, I feel like this approach of creating dozens of mixins and multiple inheritance is premature complexification / abstractifiction.  I recently read this [blog post](https://programmingisterrible.com/post/176657481103/repeat-yourself-do-more-than-one-thing-and) and felt like it was speaking directly to me! 😆 

Now I think that what we should do is create a basic working recipe class that implements the methods executors expect. As we define new recipes based that don't fit this mold, we should slowly refactor this class to make it more generalizable (rather than trying to generalize everything from day 1.)

Working on this now.

jhamman:
> In retrospect, I feel like this approach of creating dozens of mixins and multiple inheritance is premature complexification / abstractifiction.

Couldn't agree more. I think I went through this same realization a few months ago. I think for now, the executors may be enough of an abstraction to allow us to move forward. If we find that recipes are frequently sharing elements, we can pull those out one by one.

rabernat:
Just a note: in https://github.com/pangeo-data/rechunker/pull/77 I am working on an update to rechunker that intersects with this.

joshmoore:
> https://github.com/pangeo-forge/pangeo-forge/pull/27#pullrequestreview-548527179 my only concerns being how we leave space for extension to pipelines that may not use xarray or zarr.

This was my first impression as well. I've not yet been able to make multiscales representations openable by xarray.

davidbrochart:
@joshmoore what other tool are you thinking about to generate multiscale representations?

joshmoore:
@davidbrochart : really anything producing (largish) images (or generally spatial arrays?) would benefit from a multiscale representation. For other datatypes, I'd defer to other domains whether it's useful or not.

rabernat:
The tests are hanging intermittently. But I think everything is working.

rabernat:
The tests have become so unreliable for this PR. I think it has to do with starting the HTTP server.

PR #29, opened 2021-01-18
Add GHA workflow for Sphinx previews

charlesbluca:
This adds a basic workflow to build Sphinx docs and push them to a dedicated preview branch on local PRs. This should make it easier to review proposed changes.

PR #30, opened 2021-01-19
Add Sphinx preview workflow

charlesbluca:
Same as #29, but now from a local branch.


charlesbluca:
Build CI is still failing, but I don't think that really matters as much as pre-commit here.

Issue #33, opened 2021-01-22
Tests hang on github actions


rabernat:
The tests are now consistently getting stuck, e.g.

https://github.com/pangeo-forge/pangeo-forge/runs/1749653870?check_suite_focus=true

It gets suck on tests like these
```
tests/test_fixtures.py::test_fixture_http_files[D]
tests/test_fixtures.py::test_fixture_http_files[2D] 
```
never finishes.

It's because of this fixture, which uses subprocess to spawn an HTTP server
https://github.com/pangeo-forge/pangeo-forge/blob/71c367cc42f0696da5d08c132f9940e9964fc6ce/tests/conftest.py#L60-L74

I'm not sure how we can fix this.

rabernat:
This is not actually resolve. I just make the test skip on github until we have a better solution.

rabernat:
Maybe this would help? https://pypi.org/project/pytest-httpserver/

PR #59, opened 2021-01-24
Add fsspec_open_kwargs param to NetCDFtoZarrSequentialRecipe

### LINKED ISSUE #53: Non-public source datasets requiring credentials
There are a lot of primary datasets out there that require user authentication before download. E.g. NASA Earthdata Login (https://urs.earthdata.nasa.gov ). I'm curious if pangeo-forge will be limited to public datasets, or if the roadmap includes a mechanism for securely using credentials in recipes? 

I wonder if it's possible for feedstocks to be owned by the creator so that they can add their own credentials as repository secrets for GitHub actions? https://github.com/conda-forge/pangeo-dask-feedstock. On conda-forge we don't have that ability (for example https://github.com/conda-forge/pangeo-dask-feedstock). Maybe @ocefpaf has some ideas on this?

davidbrochart:
This allows e.g. HTTP authentication by passing `client_kwargs={'auth': aiohttp.BasicAuth('username', 'password')}` to `HTTPFileSystem`:
```python
recipe = NetCDFtoZarrSequentialRecipe(
    input_urls=input_urls,
    sequence_dim="time",
    inputs_per_chunk=4,
    fsspec_open_kwargs={'client_kwargs': {'auth': aiohttp.BasicAuth('username', 'password')}}
)
```
See https://github.com/pangeo-forge/pangeo-forge/issues/53#issuecomment-766397594

fixes #53

### DIFF
--- a/pangeo_forge/recipe.py
+++ b/pangeo_forge/recipe.py
@@ -141,6 +141,7 @@ class NetCDFtoZarrSequentialRecipe(BaseRecipe):
       the inputs to form a chunk.
     :param delete_input_encoding: Whether to remove Xarray encoding from variables
       in the input dataset
+    :param fsspec_open_kwargs: Extra options for opening the inputs with fsspec.
     """
 
     input_urls: Iterable[str] = field(repr=False)
@@ -154,6 +155,7 @@ class NetCDFtoZarrSequentialRecipe(BaseRecipe):
     xarray_open_kwargs: dict = field(default_factory=dict)
     xarray_concat_kwargs: dict = field(default_factory=dict)
     delete_input_encoding: bool = True
+    fsspec_open_kwargs: dict = field(default_factory=dict)
 
     def __post_init__(self):
         self._chunks_inputs = {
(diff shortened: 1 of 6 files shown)


rabernat:
This is great, thanks @davidbrochart! 

Can you think of a good way to test this?

davidbrochart:
Maybe we could set up an HTTP server with authentication? That could also allow us to test something else than the local file system. What do you think?

rabernat:
I think the most important thing is to have test coverage for the option to pass parameters. It doesn't have to be http passwords. Any `fsspec_open_kwargs` will do, as long as we can test that they are passed through correctly.

rabernat:
> Maybe we could set up an HTTP server with authentication? That could also allow us to test something else than the local file system.

We actually already to have HTTP server tests!

https://github.com/pangeo-forge/pangeo-forge/blob/bba0be81a347954bdf898190a02672d137561e53/tests/conftest.py#L62-L78

But they are broken in github workflows (see #33).

davidbrochart:
I added an HTTP server with optional authentication, so that we can test passing credentials and the new `fsspec_open_kwargs` parameter at the same time.

rabernat:
Looks like one of these tests has hung, which was the original reason we disabled the http server tests inside github workflows (see #33).

rabernat:
LGTM!

Any comments @martindurant? Might be good to get a review from the fsspec guru.

martindurant:
Looks good to me. We often call this argument `storage_options` in fsspec/intake/dask - but since in this repo "storage" means many things, I don't think it needs to be changed.

davidbrochart:
> We often call this argument `storage_options` in fsspec/intake/dask - but since in this repo "storage" means many things, I don't think it needs to be changed.

I also wanted to be consistent with the already existing `xarray_open_kwargs` parameter, and be very explicit about what it's doing: forwarding the kwargs to fsspec's open function.

Issue #58, opened 2021-01-24
Intermittent blosc decompression errors


rabernat:
Executor tests occasionally fail like this

<details>

```
___________________________________ test_recipe_w_executor[D-DaskPipelineExecutor] ____________________________________

Executor = <class 'rechunker.executors.dask.DaskPipelineExecutor'>
netCDFtoZarr_sequential_recipe = (NetCDFtoZarrSequentialRecipe(sequence_dim='time', inputs_per_chunk=1, nitems_per_input=1, target=FSSpecTarget(fs=<fss...0x7fce414fe3a0>, root_path='/private/var/folders/n8/63q49ms55wxcj_gfbtykwp5r0000gn/T/pytest-of-rpa/pytest-51/target1'))

    @pytest.mark.parametrize(
        "Executor", [PythonPipelineExecutor, DaskPipelineExecutor, PrefectPipelineExecutor]
    )
    def test_recipe_w_executor(Executor, netCDFtoZarr_sequential_recipe):
        rec, ds_expected, target = netCDFtoZarr_sequential_recipe
        pipeline = rec.to_pipelines()
        ex = Executor()
        plan = ex.pipelines_to_plan(pipeline)
>       ex.execute_plan(plan)

tests/test_executors.py:19: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../rechunker/rechunker/executors/dask.py:28: in execute_plan
    return plan.compute(**kwargs)
/opt/miniconda3/envs/pangeo2020/lib/python3.8/site-packages/dask/base.py:167: in compute
    (result,) = compute(self, traverse=False, **kwargs)
/opt/miniconda3/envs/pangeo2020/lib/python3.8/site-packages/dask/base.py:452: in compute
    results = schedule(dsk, keys, **kwargs)
/opt/miniconda3/envs/pangeo2020/lib/python3.8/site-packages/dask/threaded.py:76: in get
    results = get_async(
/opt/miniconda3/envs/pangeo2020/lib/python3.8/site-packages/dask/local.py:486: in get_async
    raise_exception(exc, tb)
/opt/miniconda3/envs/pangeo2020/lib/python3.8/site-packages/dask/local.py:316: in reraise
    raise exc
[...]

PR #61, opened 2021-01-25
use codecov

rabernat:
I discovered something very weird on my local machine.

When I run the tests with coverage:
```
$ py.test -v --cov=pangeo_forge --cov-config .coveragerc --cov-report term-missing -s -k test_fixture_http_files
```

This test fails:
```
tests/test_fixtures.py::test_fixture_http_files[2D]
...
E               FileNotFoundError: http://127.0.0.1:8080/000.nc
```

However, it works without the coverage options!

Will the same happen on CI?


rabernat:
So actually those tests skip (see #33), so we don't have a chance to see the error here.

rabernat:
Blosc error again! #58

Issue #48, opened 2021-01-24
Input / chunk preprocessing


rabernat:
It will be very important to provide hooks for recipes to plug in custom processing functions. These can be useful for fixing weird metadata or deriving new variables. Raised by @naomi-henderson in #47:
 
> pre-processing is not available in open_dataset

We could easily add a preprocessing hook to the Recipe params here:

https://github.com/pangeo-forge/pangeo-forge/blob/f9b698a35a62ab169bce546b15fef9722c70cd90/pangeo_forge/recipe.py#L154-L155

There are actually several different points where such hooks could be called. In `open_input`.

https://github.com/pangeo-forge/pangeo-forge/blob/f9b698a35a62ab169bce546b15fef9722c70cd90/pangeo_forge/recipe.py#L244-L254

Or in `open_chunk`

https://github.com/pangeo-forge/pangeo-forge/blob/f9b698a35a62ab169bce546b15fef9722c70cd90/pangeo_forge/recipe.py#L259-L269

Maybe it would be useful to have hooks for both?

PR #46, opened 2021-01-23
Fix API doc

davidbrochart:
Fixes #38

See https://pangeo-forge--46.org.readthedocs.build/en/46/api.html

There are other errors:
```
WARNING: autodoc: failed to import class 'recipe.BaseRecipe' from module 'pangeo_forge'; the following exception was raised:
cannot import name 'MultiStagePipeline' from 'rechunker.types' (/home/david/mambaforge/envs/pangeo-forge-dev/lib/python3.9/site-packages/rechunker/types.py) 
WARNING: autodoc: failed to import class 'recipe.NetCDFtoZarrSequentialRecipe' from module 'pangeo_forge'; the following exception was raised:
cannot import name 'MultiStagePipeline' from 'rechunker.types' (/home/david/mambaforge/envs/pangeo-forge-dev/lib/python3.9/site-packages/rechunker/types.py) 
WARNING: autodoc: failed to import class 'executors.PythonPipelineExecutor' from module 'pangeo_forge'; the following exception was raised:
cannot import name 'DaskPipelineExecutor' from 'rechunker.executors' (/home/david/mambaforge/envs/pangeo-forge-dev/lib/python3.9/site-packages/rechunker/executors/__init__.py)                                                                                                                                           
WARNING: autodoc: failed to import class 'executors.DaskPipelineExecutor' from module 'pangeo_forge'; the following exception was raised:
cannot import name 'DaskPipelineExecutor' from 'rechunker.executors' (/home/david/mambaforge/envs/pangeo-forge-dev/lib/python3.9/site-packages/rechunker/executors/__init__.py)                                                                                                                                           
WARNING: autodoc: failed to import class 'executors.PrefectPipelineExecutor' from module 'pangeo_forge'; the following exception was raised:
cannot import name 'DaskPipelineExecutor' from 'rechunker.executors' (/home/david/mambaforge/envs/pangeo-forge-dev/lib/python3.9/site-packages/rechunker/executors/__init__.py)
```
Indeed, it looks like `MultiStagePipeline` and `DaskPipelineExecutor` don't exist in rechunker.

PR #65, opened 2021-01-28
Add process_input and process_chunk hooks

davidbrochart:
This PR adds a `process_input` parameter to `NetCDFtoZarrSequentialRecipe`, which is a function optionally called after each input has been opened. It takes in the name of the opened file and the dataset, and returns a dataset.
The motivation is that it is very likely that we need to tweak the input data for the rest of the flow. For instance, some datasets have their timestamp encoded in the file name, so if you want to concatenate on a time dimension, you need to extract the timestamp from the file name and put it in a time coordinate.
If you think this can go in, I'll write a test.

### DIFF
--- a/pangeo_forge/recipe.py
+++ b/pangeo_forge/recipe.py
@@ -160,8 +160,8 @@ class NetCDFtoZarrSequentialRecipe(BaseRecipe):
     xarray_concat_kwargs: dict = field(default_factory=dict)
     delete_input_encoding: bool = True
     fsspec_open_kwargs: dict = field(default_factory=dict)
-    process_input: Optional[Callable[[xr.Dataset, str], xr.Dataset] = None
-    process_chunk: Optional[callable] = None
+    process_input: Optional[Callable[[xr.Dataset, str], xr.Dataset]] = None
+    process_chunk: Optional[Callable[[xr.Dataset], xr.Dataset]] = None
 
     def __post_init__(self):
         self._chunks_inputs = {
--- a/pangeo_forge/recipe.py
+++ b/pangeo_forge/recipe.py
@@ -142,6 +142,8 @@ class NetCDFtoZarrSequentialRecipe(BaseRecipe):
     :param delete_input_encoding: Whether to remove Xarray encoding from variables
       in the input dataset
     :param fsspec_open_kwargs: Extra options for opening the inputs with fsspec.
+    :param process_input: Function to call on each opened input, with signature
+      `(fname: str, ds: xr.Dataset) -> ds: xr.Dataset`.
     """
 
     input_urls: Iterable[str] = field(repr=False)
@@ -156,6 +158,7 @@ class NetCDFtoZarrSequentialRecipe(BaseRecipe):
     xarray_concat_kwargs: dict = field(default_factory=dict)
     delete_input_encoding: bool = True
     fsspec_open_kwargs: dict = field(default_factory=dict)
+    process_input: Optional[callable] = None
 
     def __post_init__(self):
         self._chunks_inputs = {
@@ -255,7 +258,11 @@ class NetCDFtoZarrSequentialRecipe(BaseRecipe):
             for var in ds.variables:
                 ds[var].encoding = {}
 
+        if self.process_input is not None:
+            ds = self.process_input(fname, ds)
+
         logger.debug(f"{ds}")
+
         return ds
 
     def open_chunk(self, chunk_key):
--- a/pangeo_forge/recipe.py
+++ b/pangeo_forge/recipe.py
@@ -160,7 +160,7 @@ class NetCDFtoZarrSequentialRecipe(BaseRecipe):
     xarray_concat_kwargs: dict = field(default_factory=dict)
     delete_input_encoding: bool = True
     fsspec_open_kwargs: dict = field(default_factory=dict)
-    process_input: Optional[callable] = None
+    process_input: Optional[Callable[[xr.Dataset, str], xr.Dataset] = None
     process_chunk: Optional[callable] = None
 
     def __post_init__(self):
--- a/pangeo_forge/recipe.py
+++ b/pangeo_forge/recipe.py
@@ -143,7 +143,7 @@ class NetCDFtoZarrSequentialRecipe(BaseRecipe):
       in the input dataset
     :param fsspec_open_kwargs: Extra options for opening the inputs with fsspec.
     :param process_input: Function to call on each opened input, with signature
-      `(ds: xr.Dataset, fname: str) -> ds: xr.Dataset`.
+      `(ds: xr.Dataset, filename: str) -> ds: xr.Dataset`.
     :param process_chunk: Function to call on each concatenated chunk, with signature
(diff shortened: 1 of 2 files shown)


martindurant:
This kind of functionality would be nice to have in Intake catalogues in general, and then you could express your inputs as usual Data Sources. We have that, e.g., for the CSV case (intake.source.base.PatternMixin), so the template syntax used there might be a good model to adhere to.

rabernat:
Thanks David! In general this is great and is and partially addresses #48. If you can also add a `process_chunk` option, then it will close #48 entirely.

We will need tests. I would prefer to have these things parametrized such that we are testing both with and without the option. I was puzzled why this didn't trigger codecov (see #61), but I realized I had not activated the webhook. So if you push again, we should get a coverage report.

rabernat:
> so the template syntax used there might be a good model to adhere to.

@martindurant could you link us directly to the code you're talking about?

martindurant:
The docstring for the CSV case: https://intake.readthedocs.io/en/latest/api_user.html#intake.source.csv.CSVSource.__init__
(path_as_pattern kwargs)
A blog about the functionality: https://www.anaconda.com/blog/intake-parsing-data-from-filenames-and-paths
Implementation: https://github.com/intake/intake/blob/master/intake/source/base.py#L430

davidbrochart:
Awesome, thanks @martindurant and @rabernat.

rabernat:
See https://github.com/pangeo-forge/pangeo-forge/issues/63#issuecomment-767947346 for a discussion of how we imagine refactoring this monster class.

davidbrochart:
I think the pattern matching in the file name mentioned by Martin could be implemented in another PR. We could provide decorators that could be applied to the hooks, like this:
```python
@match(fname="data_{year}_{month}_{day}.nc")
def process_input(ds):
    # do some processing on ds using year, month and day
    return ds
```
What do you think?

tests/test_recipe.py
```diff
@@ -8,6 +8,13 @@
 dummy_fnames = ["a.nc", "b.nc", "c.nc"]
 
 
+def incr_date(ds, fname=""):
```
rabernat, line 11: Can you explain the point of the `fname` kwarg?

tests/test_recipe.py
```diff
@@ -97,6 +110,18 @@ def test_NetCDFtoZarrSequentialRecipe(
...
+    if process_input is not None:
+        # check that the process_input hook made some changes
+        assert not ds_target.identical(ds_expected)
+        # apply these changes to the expected dataset
+        ds_expected = process_input(ds_expected)
+    if process_chunk is not None:
+        # check that the process_chunk hook made some changes
+        assert not ds_target.identical(ds_expected)
+        # apply these changes to the expected dataset
+        ds_expected = process_chunk(ds_expected)
```
rabernat, line 123: Great test.

rabernat (commented):
🎉 This is fantastic @davidbrochart!

tests/test_recipe.py
davidbrochart: We use `incr_date` for both `process_input` and `process_chunk` in the test, but they have a different signature:
- `process_input(ds, fname) -> ds`
- `process_chunk(ds) -> ds`

Would you like to have different hooks?

rabernat: Ok, yes I see now. Adding the fname argument in the `process_input` makes sense I guess. If we were opening files directly with xarray (rather than first with fsspec) this information would also be available in the xarray `.encoding` attributes. But that's not the case here.

rabernat:
It would be nice to get #6 in and then rebase so we could see the documentation built from this PR.

davidbrochart:
I think you meant #46, I'm looking at it.

rabernat:
Docs look great! Thanks so much for your work on this David!

I'd like to get one more set of eyes on this. I've tagged a few potential reviewers.

pangeo_forge/recipe.py
```diff
@@ -156,6 +160,8 @@ class NetCDFtoZarrSequentialRecipe(BaseRecipe):
     xarray_concat_kwargs: dict = field(default_factory=dict)
     delete_input_encoding: bool = True
     fsspec_open_kwargs: dict = field(default_factory=dict)
+    process_input: Optional[callable] = None
```
TomAugspurger, line 163:
```suggestion
    process_input: Optional[Callable[[xr.Dataset, str], xr.Dataset] = None
```

(just guessing on the type. But I think that's the syntax for typing the inputs and outputs of the callable.)

tests/test_recipe.py
TomAugspurger: I was also confused on this point. Can we rename `fname` to `filename`? I read `fname` as function name.

TomAugspurger (approved):

pangeo_forge/recipe.py
```diff
@@ -156,6 +160,8 @@ class NetCDFtoZarrSequentialRecipe(BaseRecipe):
     xarray_concat_kwargs: dict = field(default_factory=dict)
     delete_input_encoding: bool = True
     fsspec_open_kwargs: dict = field(default_factory=dict)
+    process_input: Optional[Callable[[xr.Dataset, str], xr.Dataset] = None
```
TomAugspurger, line 163: Whoops, sorry my suggestion was missing a `]`.

Might want to make sure `typing.Callable` is imported as well.

davidbrochart: Thanks @TomAugspurger!

pangeo_forge/recipe.py
```diff
@@ -142,6 +142,10 @@ class NetCDFtoZarrSequentialRecipe(BaseRecipe):
     :param delete_input_encoding: Whether to remove Xarray encoding from variables
       in the input dataset
     :param fsspec_open_kwargs: Extra options for opening the inputs with fsspec.
+    :param process_input: Function to call on each opened input, with signature
+      `(ds: xr.Dataset, fname: str) -> ds: xr.Dataset`.
```
rabernat, line 146:
```suggestion
      `(ds: xr.Dataset, filename: str) -> ds: xr.Dataset`.
```

davidbrochart:
I see we're still having an issue with the HTTP server in the CI?

rabernat:
> I see we're still having an issue with the HTTP server in the CI?

Yes. The short term fix is maybe to just bump the `sleep` call again. The long term one is maybe to use this: https://pypi.org/project/pytest-httpserver/

davidbrochart:
Not sure pytest-httpserver will help since the issue doesn't seem to be with pytest but with the CI (I've never seen the tests hang locally).

martindurant:
I commonly have something like this to wait for servers:
```
    timeout = 5
    while True:
        try:
            r = requests.get(endpoint_uri)
            if r.ok:
                break
        except:
            pass
        timeout -= 0.1
        assert timeout > 0, "Timed out waiting for server"
        time.sleep(0.1)
```

davidbrochart:
Thanks Martin, that's much better than waiting for an arbitrary amount of time! I'll open a PR.

rabernat:
> (I've never seen the tests hang locally).

I have! On my os-x laptop.

PR #66, opened 2021-01-29
Try and fix HTTP server in tests

davidbrochart:
See https://github.com/pangeo-forge/pangeo-forge/pull/65#issuecomment-769814126

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,7 +3,7 @@ max-line-length = 100
 
 [isort]
 known_first_party=pangeo_forge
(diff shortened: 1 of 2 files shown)


martindurant (commented):
Yes, I approve, but note that the tests for HTTPFileSystem (in fsspec) show a way to do this in-process with a thread, which appears to start essentially instantly. I haven't looked in http_auth_server.py, it's probably not worth the hassle here.

davidbrochart:
Actually the tests still hang, could you point to the HTTPFileSystem test in fsspec?

martindurant:
The class: https://github.com/intake/filesystem_spec/blob/master/fsspec/implementations/tests/test_http.py#L18

and the context and fixture that use it follow. This serves custom payloads, not files.

PR #67, opened 2021-01-31
Use pytest-httpserver

davidbrochart:
I tested locally and didn't manage to get the hanging issue we were having (I could reproduce it before this PR).
I had to change the way we test `fsspec_open_kwargs` because basic authentication doesn't seem to be possible with pytest-httpserver (it seems to only provide digest authentication). But anyway the point of this test was rather to see if the kwargs were propagated to the fsspec opener, so it doesn't really matter what we propagate. The simplest thing I could think about is to force an HTTP request to time out.
Closes #65


tests/conftest.py
```diff
@@ -65,33 +64,19 @@ def netcdf_local_paths(daily_xarray_dataset, tmpdir_factory, request):
...
-
-        def teardown():
-            p.kill()
-
-        request.addfinalizer(teardown)
+        url = f"http://127.0.0.1:{httpserver.port}"
+        for fname in fnames:
+            with open(os.path.join(basedir, fname), "rb") as f:
+                data = f.read()
+            httpserver.expect_request(f"/{fname}").respond_with_data(data)
```
davidbrochart, line 78: This loads the file contents in memory, but I couldn't think of another way to do it. Any suggestions?

csernazs: pytest-httpserver allows to register a callable which can return a Response object, so you don't have to load all the files to the memory.

https://pytest-httpserver.readthedocs.io/en/latest/tutorial.html#specifying-responses

```python
def my_handler(request):
    # here, examine the request object
    return Response("Hello world!")

respond_with_handler(my_handler)
```

It uses the request/response objects from werkzeug, for the API docs, see here: https://werkzeug.palletsprojects.com/en/1.0.x/wrappers/

For the authentication: it supports to handle headers, such as the `Authorization` header, the documentation includes example for digest authentication but based on the code, it should work with basic auth as well, you would only need to include the base64 string to the header matcher.

davidbrochart: Thanks for the help @csernazs! I'll update this PR with your suggestions.

davidbrochart:
I added back the authentication test and used the `respond_with_handler` to avoid loading all files in memory, thanks again @csernazs.
It doesn't seem to solve our issues with the hanging test though. But even though, I think it is better to rely on pytest-httpserver than setting up our own server.

davidbrochart:
Actually maybe pytest-httpserver can help debugging this issue, through the use of `check_assertions()`. I'm going to add one in the handler and see if we can catch an error. I'll trigger the CI until the issue appears.

davidbrochart:
I had the issue locally, and `check_assertions()` didn't help. Here is the trace after pressing CTRL-C (from where it enters fsspec):

<details>
  <summary>Click to expand</summary>

>   ???

h5py/h5fd.pyx:158: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <File-like object HTTPFileSystem, http://127.0.0.1:46711/001.nc>, b = <MemoryView of 'array' at 0x7f3aba4f51e0>

    def readinto(self, b):
        """mirrors builtin file's readinto method
    
        https://docs.python.org/3/library/io.html#io.RawIOBase.readinto
        """
        out = memoryview(b).cast("B")
>       data = self.read(out.nbytes)

../../../mambaforge/envs/pangeo-forge-dev/lib/python3.9/site-packages/fsspec/spec.py:1442: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <File-like object HTTPFileSystem, http://127.0.0.1:46711/001.nc>, length = 512

    def read(self, length=-1):
        """Read bytes from file
    
        Parameters
        ----------
        length: int
            Read up to this many bytes. If negative, read all content to end of
            file. If the server has not supplied the filesize, attempting to
            read only part of the data will raise a ValueError.
        """
        if (
            (length < 0 and self.loc == 0)
            or (length > (self.size or length))  # explicit read all
            or (  # read more than there is
                self.size and self.size < self.blocksize
            )  # all fits in one block anyway
        ):
>           self._fetch_all()

../../../mambaforge/envs/pangeo-forge-dev/lib/python3.9/site-packages/fsspec/implementations/http.py:436: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

args = (<File-like object HTTPFileSystem, http://127.0.0.1:46711/001.nc>,), kwargs = {}
self = <File-like object HTTPFileSystem, http://127.0.0.1:46711/001.nc>

    @functools.wraps(func)
    def wrapper(*args, **kwargs):
        self = obj or args[0]
>       return maybe_sync(func, self, *args, **kwargs)

../../../mambaforge/envs/pangeo-forge-dev/lib/python3.9/site-packages/fsspec/asyn.py:121: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

func = <function HTTPFile.async_fetch_all at 0x7f3aba51b310>, self = <File-like object HTTPFileSystem, http://127.0.0.1:46711/001.nc>
args = (<File-like object HTTPFileSystem, http://127.0.0.1:46711/001.nc>,), kwargs = {}, loop = <_UnixSelectorEventLoop running=True closed=False debug=False>

    def maybe_sync(func, self, *args, **kwargs):
        """Make function call into coroutine or maybe run
    
        If we are running async, run coroutine on current loop until done;
        otherwise runs it on the loop (if is a coroutine already) or directly. Will guess
        we are running async if either "self" has an attribute asynchronous which is True,
        or thread_state does (this gets set in ``sync()`` itself, to avoid nesting loops).
        """
        loop = self.loop
        # second condition below triggers if this is running in the thread of the
        # event loop *during* the call to sync(), i.e., while running
        # asynchronously
        if getattr(self, "asynchronous", False) or getattr(
            thread_state, "asynchronous", False
        ):
            if inspect.iscoroutinefunction(func):
                # run coroutine while pausing this one (because we are within async)
                return _run_until_done(func(*args, **kwargs))
            else:
                # make awaitable which then calls the blocking function
                return _run_as_coroutine(func, *args, **kwargs)
        else:
            if inspect.iscoroutinefunction(func):
                # run the awaitable on the loop
>               return sync(loop, func, *args, **kwargs)

../../../mambaforge/envs/pangeo-forge-dev/lib/python3.9/site-packages/fsspec/asyn.py:100: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

loop = <_UnixSelectorEventLoop running=True closed=False debug=False>, func = <function HTTPFile.async_fetch_all at 0x7f3aba51b310>, callback_timeout = None
args = (<File-like object HTTPFileSystem, http://127.0.0.1:46711/001.nc>,), kwargs = {}, f = <function sync.<locals>.f at 0x7f3aba43c9d0>

    def sync(loop, func, *args, callback_timeout=None, **kwargs):
        """
        Run coroutine in loop running in separate thread.
        """
    
        e = threading.Event()
        main_tid = threading.get_ident()
        result = [None]
        error = [False]
    
        async def f():
            try:
                if main_tid == threading.get_ident():
                    raise RuntimeError("sync() called from thread of running loop")
                await asyncio.sleep(0)
                thread_state.asynchronous = True
                future = func(*args, **kwargs)
                if callback_timeout is not None:
                    future = asyncio.wait_for(future, callback_timeout)
                result[0] = await future
            except Exception:
                error[0] = sys.exc_info()
            finally:
                thread_state.asynchronous = False
                e.set()
    
        asyncio.run_coroutine_threadsafe(f(), loop=loop)
        if callback_timeout is not None:
            if not e.wait(callback_timeout):
                raise TimeoutError("timed out after %s s." % (callback_timeout,))
        else:
            while not e.is_set():
>               e.wait(10)

../../../mambaforge/envs/pangeo-forge-dev/lib/python3.9/site-packages/fsspec/asyn.py:68: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <threading.Event object at 0x7f3aba42b280>, timeout = 10

    def wait(self, timeout=None):
        """Block until the internal flag is true.
    
        If the internal flag is true on entry, return immediately. Otherwise,
        block until another thread calls set() to set the flag to true, or until
        the optional timeout occurs.
    
        When the timeout argument is present and not None, it should be a
        floating point number specifying a timeout for the operation in seconds
        (or fractions thereof).
    
        This method returns the internal flag on exit, so it will always return
        True except if a timeout is given and the operation times out.
    
        """
        with self._cond:
            signaled = self._flag
            if not signaled:
>               signaled = self._cond.wait(timeout)

../../../mambaforge/envs/pangeo-forge-dev/lib/python3.9/threading.py:574: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <Condition(<unlocked _thread.lock object at 0x7f3aba42bba0>, 0)>, timeout = 10

    def wait(self, timeout=None):
        """Wait until notified or until a timeout occurs.
    
        If the calling thread has not acquired the lock when this method is
        called, a RuntimeError is raised.
    
        This method releases the underlying lock, and then blocks until it is
        awakened by a notify() or notify_all() call for the same condition
        variable in another thread, or until the optional timeout occurs. Once
        awakened or timed out, it re-acquires the lock and returns.
    
        When the timeout argument is present and not None, it should be a
        floating point number specifying a timeout for the operation in seconds
        (or fractions thereof).
    
        When the underlying lock is an RLock, it is not released using its
        release() method, since this may not actually unlock the lock when it
        was acquired multiple times recursively. Instead, an internal interface
        of the RLock class is used, which really unlocks it even when it has
        been recursively acquired several times. Another internal interface is
        then used to restore the recursion level when the lock is reacquired.
    
        """
        if not self._is_owned():
            raise RuntimeError("cannot wait on un-acquired lock")
        waiter = _allocate_lock()
        waiter.acquire()
        self._waiters.append(waiter)
        saved_state = self._release_save()
        gotit = False
        try:    # restore state no matter what (e.g., KeyboardInterrupt)
            if timeout is None:
                waiter.acquire()
                gotit = True
            else:
                if timeout > 0:
>                   gotit = waiter.acquire(True, timeout)
E                   KeyboardInterrupt

../../../mambaforge/envs/pangeo-forge-dev/lib/python3.9/threading.py:316: KeyboardInterrupt
</details>

@martindurant any chance you can see something wrong by looking at where we are in fsspec?

martindurant:
We are waiting on an event loop in another thread. Perhaps the server is waiting for the same event loop? I don't know how this server is implemented.

davidbrochart:
Interesting, we had the same issue before using pytest-httpserver, when the server was just `SimpleHTTPServer` running in another process. From what I saw, the issue always happens the second time we run the tests involving HTTP requests, never the first time.

davidbrochart:
If I understand correctly, fsspec is natively async, and when it is not running in an async environment, runs the async functions on a loop in another thread (at least for HTTP file systems)?

martindurant:
Yes, that's correct

davidbrochart:
@csernazs I can see [in the documentation](https://pytest-httpserver.readthedocs.io/en/latest/howto.html#multi-threading-support):
```
Multi-threading support
When your client runs in a thread, everything completes without waiting for the first response. To overcome this problem, you can wait until all the handlers have been served or there’s some error happened.
```
But I guess this is not the case here, since our tests pass (when they don't hang).
More generally, do you see an issue with our client running in another thread?

martindurant:
Nothing obvious, sorry

davidbrochart:
Thanks @martindurant, this was more a question for @csernazs (sorry I was not clear).
I guess we can't use the `with httpserver.wait` context manager, since our handlers are permanent.

martindurant:
You *might* just want to play with the scope of the fixture, though. Perhaps ensuring that it outlives the test function.

davidbrochart:
From the documentation on the [server lifetime](https://pytest-httpserver.readthedocs.io/en/latest/tutorial.html#server-lifetime):
> Http server is started when the first test uses the httpserver fixture, and it will be running for the rest of the session. The server is not stopped and started between the tests as it is an expensive operation, it takes up to 1 second to properly stop the server.
To avoid crosstalk (eg one test leaving its state behind), the server’s state is cleaned up between test runnings.

martindurant:
I wonder why it would take so long!

csernazs:
I'm not sure I fully understood your problem (I ran into this issue by accident and I wanted to help), but here's what I can add:

pytest-httpserver is based on werkzeug which uses the socketserver implementation of the standard library.
This is the code which causes this delay, see poll_interval's default value of 0.5 seconds:
https://github.com/python/cpython/blob/fc40b3020cf3c869833fd5d3720cf9768fe3bb46/Lib/socketserver.py#L232

The server is started in a thread by the pytest-httpserver's code, so you can run your client's code in the main thread if that's possible. As you said, using the wait() context manager is only possible with "expiring" handlers such as oneshot and ordered ones.

In pytest-httpserver, the plugin's code is just a glue code for the real server so you can implement your own version (eg if you want to start/stop the server for each test) if you want. All you need is to instantiate the HTTPServer object and call it's start and stop methods accordingly.

Replacing the socketserver with a better implementation is on my todo list and I think it could be possible without breaking the API but there's no ETA for this.

davidbrochart:
> I'm not sure I fully understood your problem (I ran into this issue by accident and I wanted to help)

No worries, I was just wondering if there was something obviously wrong in our use case that you could spot .

> The server is started in a thread by the pytest-httpserver's code, so you can run your client's code in the main thread if that's possible.

Here we don't have a choice, our client (fsspec) makes requests from another thread (not the main thread).

> In pytest-httpserver, the plugin's code is just a glue code for the real server so you can implement your own version (eg if you want to start/stop the server for each test)

I don't think we want to start/stop the server for each test, as long as its state is cleared between tests.

csernazs:
Well, I think I cannot help more from the httpserver's point, but maybe it helps:
There's a `log` attribute of the httpserver which could be polled periodically to see which requests had been served and consider the test completed on some condition (eg amount of requests, or checking the last request served). Also you can set a threading event from the request handler on some condition which finish the test. There would be some condition which means the end of the test running I think.

rabernat:
Let's discuss this PR at our meeting on Monday.

PR #74, opened 2021-02-07
Support variables in different files

### LINKED ISSUE #39: Support multiple variables / multiple timesteps
Currently we only support sequential inputs, along a single dimension.

```
foo.0.nc
foo.1.nc
...
foo.N.nc
```

However, many potential recipes instead have multiple different variables to be combined, e.g.
```
foo.nc
bar.nc
baz.nc
```

And in combination
```
foo.0.nc
bar.0.nc
baz.0.nc
foo.1.nc
...
baz.N.nc
```

For example the now broken [terraclimate recipe](https://github.com/pangeo-forge/terraclimate-feedstock) follows this pattern.

A good next step for the recipe complexity is to figure out how to handle this. Some ideas come to mind.

How will the API look? Currently we just use a flat list of `input_urls`. That won't work for multiple variables.
https://github.com/pangeo-forge/pangeo-forge/blob/276755393859840ee184e1b4f3bdb3c60ad6fe7f/pangeo_forge/recipe.py#L146 
Is a nested list enough to resolve this? Or a dictionary? What sort of validation is required on such inputs?

Currently we store the relationship between chunks and inputs in a private dictionary:
https://github.com/pangeo-forge/pangeo-forge/blob/276755393859840ee184e1b4f3bdb3c60ad6fe7f/pangeo_forge/recipe.py#L158-L161
The chunk keys are just sequential integers: `0, 1, ..., N`.

For multiple variables (but no timesteps), we might use variable names instead: `'foo', 'bar', 'baz'`.

For multiple variables AND timesteps, a tuple would do the trick: `('foo', 0), ('foo', 1), ..., ('baz', N)`.

We will also may to modify how we do `initialize_target`:
https://github.com/pangeo-forge/pangeo-forge/blob/276755393859840ee184e1b4f3bdb3c60ad6fe7f/pangeo_forge/recipe.py#L175

Right now we just read the first chunk and use that. But will that still work for multiple variables? Or do we have to read one of each variable?

rabernat:
closes #39 

This PR refactors `NetCDFtoZarrSequentialRecipe`. It now inherits from `NetCDFtoZarrRecipe`. Another child class `NetCDFtoZarrMultiVarSequentialRecipe`, implements the functionality needed for #39: allowing data for each timestep to be spread across multiple different files.

In the process, I introduce a new module called `patterns` which contains some classes to represent how filenames are organized.

Todo:
- [x] Add a tutorial for this recipe
- [x] Check all docstrings
- [ ] Refactor testing to use more parametrization over different recipe types


rabernat:
hung http server again (#33)

pangeo_forge/patterns.py
```diff
@@ -0,0 +1,54 @@
...
+
+import itertools
+import string
+from abc import ABC
+from dataclasses import dataclass
+from typing import Dict, Generator, Iterable, List, Tuple
+
+
+class BasePattern(ABC):
+    def __iter__(self) -> Generator[Tuple[Tuple, str], None, None]:
```
rabernat, line 13:
```suggestion
    def __iter__(self) -> Generator[Tuple[Hashable, str], None, None]:
```

rabernat:
The blosc decompression issue (#58) is becoming very reliable. 🙄

pangeo_forge/recipe.py
```diff
@@ -146,10 +153,10 @@ class NetCDFtoZarrSequentialRecipe(BaseRecipe):
       `(ds: xr.Dataset, filename: str) -> ds: xr.Dataset`.
     :param process_chunk: Function to call on each concatenated chunk, with signature
       `(ds: xr.Dataset) -> ds: xr.Dataset`.
+    :param target_chunks: Desired chunk structure for the targret dataset.
     """
 
-    input_urls: Iterable[str] = field(repr=False)
-    sequence_dim: str
+    sequence_dim: str = ""
```
TomAugspurger, line 159: What does an empty sting for sequence_dim mean? Does that mean the dimension name should literally be `""`? Or should we use an Optional[str] here?

pangeo_forge/recipe.py
```diff
@@ -321,18 +372,75 @@ def region_for_chunk(self, chunk_key):
...
         # chunking
         return {self.sequence_dim: self.inputs_per_chunk * self.nitems_per_input}
 
     def iter_chunks(self):
         for k in self._chunks_inputs:
             yield k
+
+
+@dataclass
+class NetCDFtoZarrSequentialRecipe(NetCDFtoZarrRecipe):
```
TomAugspurger, line 389: Thoughts on how to document this? I guess sphinx will pick up the parent doctoring. But the most important thing here is `input_urls`, which I think would be undocumented.

So I think that document the `input_urls` parameter is most important, and we can include a ref to `NetCDFtoZarrRecipe`.

Same comment on NetCDFtoZarrMultiVarSequentialRecipe.

TomAugspurger (commented):
Looks good on a quick skim.

Ok to add the review-nb GitHub app to this repo? It'll make reviewing notebooks easier: https://www.reviewnb.com. Here's  an example from dask-examples: https://github.com/dask/dask-examples/pull/171#issuecomment-719039237.

pangeo_forge/recipe.py
rabernat: All parameters will need defaults if we want to use multiple inheritance / mixins without going crazy. I agree with Optional.

pangeo_forge/recipe.py
rabernat: Fixed in latest commit.

PR #76, opened 2021-02-17
Use uvloop event loop

davidbrochart:
It seems to solve the issue with the hanging HTTP tests, or maybe it makes them much more unlikely...

### DIFF
--- a/ci/py3.7.yml
+++ b/ci/py3.7.yml
@@ -23,6 +23,7 @@ dependencies:
   - pynio
   - pytest
   - pytest-cov
+  - uvloop
   - rasterio
   - scipy
   - setuptools
--- a/ci/py3.8.yml
+++ b/ci/py3.8.yml
@@ -23,6 +23,7 @@ dependencies:
   - pynio
   - pytest
   - pytest-cov
+  - uvloop
   - rasterio
   - scipy
   - setuptools
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,6 +1,7 @@
 pytest
 pytest-sugar
 pytest-cov
+uvloop
 scipy
 netcdf4
 h5netcdf
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,7 +3,7 @@ max-line-length = 100
 
 [isort]
 known_first_party=pangeo_forge
-known_third_party=aiohttp,fsspec,numpy,pandas,pkg_resources,prefect,pytest,rechunker,setuptools,sphinx_book_theme,xarray,zarr
+known_third_party=aiohttp,fsspec,numpy,pandas,pkg_resources,prefect,pytest,rechunker,setuptools,sphinx_book_theme,uvloop,xarray,zarr
 multi_line_output=3
 include_trailing_comma=True
 force_grid_wrap=0
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,10 +7,13 @@ import fsspec
 import numpy as np
 import pandas as pd
 import pytest
+import uvloop
 import xarray as xr
 
 from pangeo_forge import recipe
 from pangeo_forge.storage import CacheFSSpecTarget, FSSpecTarget, UninitializedTarget
+
+uvloop.install()
 
 
 def get_open_port():


martindurant:
Is uvloop being used by fsspec, the server, or both?
It ought to be perfectly compatible with fsspec, but I'm not aware of it being used elsewhere. fsspec does do some low-level event loop things in certain conditions (nested calls).

davidbrochart:
> Is uvloop being used by fsspec, the server, or both?

Only by fsspec.

> It ought to be perfectly compatible with fsspec, but I'm not aware of it being used elsewhere. fsspec does do some low-level event loop things in certain conditions (nested calls).

As long as fsspec doesn't patch asyncio's code, uvloop should be a drop-in replacement.
FYI, the issue seems to boil down to `asyncio.run_coroutine_threadsafe(f(), loop=loop)` not actually calling `f` [here](https://github.com/intake/filesystem_spec/blob/4d03e1e07ada08f91784e7826c355da71a13aa1e/fsspec/asyn.py#L62).

martindurant:
Can you explain further? Why would f not run or get stuck?

On February 17, 2021 6:16:03 PM EST, David Brochart <notifications@github.com> wrote:
>> Is uvloop being used by fsspec, the server, or both?
>
>Only by fsspec.
>
>> It ought to be perfectly compatible with fsspec, but I'm not aware of
>it being used elsewhere. fsspec does do some low-level event loop
>things in certain conditions (nested calls).
>
>As long as fsspec doesn't patch asyncio's code, uvloop should be a
>drop-in replacement.
>FYI, the issue seems to boil down to
>`asyncio.run_coroutine_threadsafe(f(), loop=loop)` not actually calling
>`f`
>[here](https://github.com/intake/filesystem_spec/blob/4d03e1e07ada08f91784e7826c355da71a13aa1e/fsspec/asyn.py#L62).
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/pangeo-forge/pangeo-forge/pull/76#issuecomment-780916916

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

davidbrochart:
I have no explanation, I stopped investigating after checking that when it hangs, it is because `asyncio.run_coroutine_threadsafe(f(), loop=loop)` doesn't run `f`. Next step is looking at asyncio's code, but this is becoming very strange.

davidbrochart:
BTW, I was wondering why there is this kind of polling [here](https://github.com/intake/filesystem_spec/blob/4d03e1e07ada08f91784e7826c355da71a13aa1e/fsspec/asyn.py#L67-L68). I think you could just `e.wait()`.

martindurant:
That was copied from `distributed` - happy to change it. Note, that I don't think any of the code calls with a timeout, so the whole function could be a bit simpler. On the other hand, perhaps exactly a timeout would have prevented the hanging.

davidbrochart:
There are things that look suspicious in the code, although I've not been able to prove they're responsible for the hanging. For instance, don't you have to use a lock when you set/get variables from different threads, [here](https://github.com/intake/filesystem_spec/blob/4d03e1e07ada08f91784e7826c355da71a13aa1e/fsspec/asyn.py#L51), [here](https://github.com/intake/filesystem_spec/blob/4d03e1e07ada08f91784e7826c355da71a13aa1e/fsspec/asyn.py#L59) and [there](https://github.com/intake/filesystem_spec/blob/4d03e1e07ada08f91784e7826c355da71a13aa1e/fsspec/asyn.py#L88-L90)?

martindurant:
`thread_state` is a thread-local variable, so those are fine.

rabernat:
I've been neglecting this, sorry. I don't really understand this stuff, so if you two agree this approach is the right one, I'm happy to merge. Are we at a consensus?

davidbrochart:
I don't understand either what is the cause of the issue, I just saw that using uvloop seems to solve it. Not sure we should merge this PR.

martindurant:
Do you want to try with https://github.com/intake/filesystem_spec/pull/572 ?

rabernat:
@martindurant - do you mean run the test suite unchanged with an unreleased version of fsspec? Yes, I'd be happy to try it. Could you make a new PR to update the test env?

martindurant:
Not even merged at the moment, so would have to install from the branch, i.e., good to see if the CI works, but you wouldn't want to merge the PR.

PR #77, opened 2021-02-22
use incremental downloading

rabernat:
This avoids loading the whole input file into memory. As discussed with @martindurant here: https://gitter.im/ContinuumIO/intake?at=6033b5f16825b964d6268de9


martindurant:
+1

PR #78, opened 2021-02-23
Cache input metadata

### LINKED ISSUE #50: Handle inputs with variable items per input
This problem was originally noted by @naomi-henderson in #47.

Our current recipe assumes that the number of "items" per input is constant:

https://github.com/pangeo-forge/pangeo-forge/blob/f9b698a35a62ab169bce546b15fef9722c70cd90/pangeo_forge/recipe.py#L149

where "items" refers to the length along the `sequence_dim` dimension (e.g. time). This assumptions buys us something big: we don't have to examine each input file, and we can determine which region of the target to write each chunk to using simple math:

https://github.com/pangeo-forge/pangeo-forge/blob/f9b698a35a62ab169bce546b15fef9722c70cd90/pangeo_forge/recipe.py#L304-L310

This makes sense for 1-file-per-day netCDFs, but many datasets have variable sizes, e.g. one file per month (31 days, 30, days, 28 days) or just whatever random stuff CMIP models do. For those cases, we will inevitably have to peek into each input file to determine its dimensions / shape and pass this information on to `prepare_target` and `store_chunk` somehow.

Here are some of the issues raised by this scenario

### How to determine the total size of the target dataset?

In the current recipe class, we calculate the length along `sequence_dim` deterministically as 

https://github.com/pangeo-forge/pangeo-forge/blob/f9b698a35a62ab169bce546b15fef9722c70cd90/pangeo_forge/recipe.py#L312-L314

and then use it to resize the target zarr.

https://github.com/pangeo-forge/pangeo-forge/blob/f9b698a35a62ab169bce546b15fef9722c70cd90/pangeo_forge/recipe.py#L183

A key point is that the size of the dataset is known from the beginning, without reading any files.

For these uneven datasets, we can't do that. We need to look at each file. Fortunately, we already almost do that, when we cache the inputs.

**I propose we write a .json "sidecar" file for each input that contains the dataset shape, variables, etc.**. A very easy way to do this would using xarray's [to_dict](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.to_dict.html) function (with `data=False`) from `cache_input`. Reading these files should be cheap enough that we can read them all in `prepare_target`. Then we can figure out how big the target dataset will be.

### How to handle chunking and chunk overlap?

Currently we assume a fixed number of `inputs_per_chunk`, which implies the pangeo forge "chunks" will vary in size and not line up with the target chunks (zarr chunks have to be constant shape). Our whole lexicon breaks down a bit, since pangeo-forge's "chunks" no longer map directly to target zarr chunks. We possibly need a new parameter, `target_chunks`, to specify the target chunking explicitly.

I see two different directions we could go to handle the problem of overlap:
- Keep `inputs_per_chunk` fixed, accept that pangeo-forge chunks will not line up with zarr chunks, and either
  - switch to append-style writes (as in the original recipe). We lose the ability for concurrency in the `store_chunk` stage, as the appends have to be sequential.
  - manually calculate the overlap between the pangeo-forge chunks (irregular) and the zarr chunks (regular). Then use distributed locks (also raised in #45) to prevent concurrent writes to a the same zarr chunk by different processes in `store_chunk`. 
- Keep the 1:1 mapping between between pangeo-forge chunks and zarr chunks, but break the requirement that `inputs_per_chunk` be fixed. Instead, allow multiple different chunks to draw from the same input. `open_chunk` possibly needs to figure out how to subset inputs. This is maybe easier because we don't need locks for concurrent reads. However, _it makes it impossible to precalculate the number of `store_chunk` tasks_, since we don't know the total size of the dataset until we start running the recipe! 🤯 This would violate our current Pipeline execution model and require a deeper refactoring of execution.

In any case, the `open_chunk` and `store_chunk` functions will need to have more context about the dataset they are writing to and their position within it. This information could possibly be calculated and stored with the sidecar file mentioned above.

---

In general, I think we should not hesitate to use the storage layer to communicate. The alternative is having tasks pass information to each other directly within the pipeline. Currently all of the pipeline function calls and arguments are determined at the outset, when the recipe is initialized, and none of the functions return anything. Changing that would be a pretty deep change at this point and would likely force us to drop this whole Pipeline structure I've cooked up and just start writing prefect task / flows directly again.

rabernat:
This mega PR fixes #50 by introducing a new concept: a metadata cache that is essentially used as a database across different stages of the recipe. The metadata cache is used to remember the dimensions of each input, which can be used later by `prepare_target` to figure out the expected size of the sequence dim.

I also implemented locking to prevent concurrent writes to overlapping regions.

All file-opening methods (e.g. `open_input` and `open_chunk` are now contextmanagers).

- [x] More tests
- [x] Update docstrings
- [x] Update narrative docs
- [x] Update tutorials


rabernat:
Hmm, new tests seems to be hanging.

davidbrochart:
Do you want to give #76 a try, see if that helps?

rabernat:
This actually does not use the HTTP tests (temporarily disabled). I think the hanging is related to dask.

davidbrochart:
Oh, if that's the case I was really not searching in the right direction.

rabernat:
No I think you were! They are distinct issues.

rabernat:
I am hoping to finish this up in the next few days. I realize it is a bit un-reviewable. This is my fault for making such a massive PR. 

If either @TomAugspurger or @davidbrochart has 15 minutes free, I'd love to just talk through some of these changes with someone. I think I'm heading in the right direction, but, for the sake of maintainability, it's important that people besides me understand this code. Let me know if you think you can find some time for this.

TomAugspurger:
I'll take a look through it. I can chat today between 10:30 - 2:00 Eastern.

rabernat:
Awesome, do you want to just hop on https://whereby.com/pangeo 12-12:15? Thanks so much Tom!

pangeo_forge/recipe.py
```diff
@@ -117,10 +123,24 @@ def __post_init__(self):
...
+
+def _chunk_metadata_fname(chunk_key) -> str:
+    return "chunk-meta-" + _encode_key(chunk_key) + ".json"
+
+
 # Notes about dataclasses:
 # - https://www.python.org/dev/peps/pep-0557/#inheritance
 # - https://stackoverflow.com/questions/51575931/class-inheritance-in-python-3-7-dataclasses
 
+_GLOBAL_METADATA_KEY = "pangeo-forge-recipe-metadata.json"
```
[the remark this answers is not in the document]
TomAugspurger: On further reading, it seems like two things get written here: a bit of metadata that applies to the entire dataset, and a piece of metadata that applies to each chunk. So this seems fine.

TomAugspurger (commented):
Left a few inline comments, but overall this seems to make sense.

pangeo_forge/recipe.py
```diff
@@ -197,37 +215,44 @@ def _prepare_target():
...
-                # actually not necessary if we use decode_times=False
-
-                if self.target_chunks:
-                    # target_chunks has undergone some validation in __post_init__
-                    ds = ds.chunk(self.target_chunks)
-
-                self.initialize_target(ds)
+                    with self.open_chunk(chunk_key) as ds:
+                        # need to have the data in memory to avoid weird chunk problems
+                        ds.load()
```
rabernat, line 223: This is really not ideal. Can we avoid it somehow and keep the data loading lazy?

pangeo_forge/recipe.py
```diff
@@ -197,37 +215,44 @@ def _prepare_target():
...
+
+                                chunks = tuple(
+                                    chunks.get(n, s) for n, s in enumerate(this_var.shape)
+                                )
+                                ds[v].encoding["chunks"] = chunks
+                            else:
+                                ds[v].encoding["chunks"] = ds[v].shape
+
+                        target_mapper = self.target.get_mapper()
+                        ds.to_zarr(target_mapper, mode="a", compute=False)
```
rabernat, line 243: `compute=False` does not have the desired effect if the data are not dask arrays.

TomAugspurger: Here (and or maybe up above before `load`) are you able to slice all the dimensions to size 0? Does that accomplish essentially the same thing?

```
ds = ds.isel(**{k: slice(0) for k in ds.dims})
```

pangeo_forge/recipe.py
```diff
@@ -197,37 +215,44 @@ def _prepare_target():
...
-            self.expand_target_dim(self.sequence_dim, self.sequence_len())
+            input_sequence_lens = self.sequence_lens()
+            n_sequence = sum(input_sequence_lens)
+            self.expand_target_dim(self.sequence_dim, n_sequence)
+
+            if not self.nitems_per_input:
+                # if nitems_per_input is not constant, we need to cache this info
+                recipe_meta = {"input_sequence_lens": input_sequence_lens}
+                meta_mapper = self.metadata_cache.get_mapper()
+                meta_mapper[_GLOBAL_METADATA_KEY] = json.dumps(recipe_meta).encode("utf-8")
```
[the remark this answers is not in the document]
rabernat: Correct. It reads per-input metadata, which is different from global metadata.

pangeo_forge/recipe.py
rabernat: But then the target variables wouldn't get the right dimensions.

I think https://github.com/pydata/xarray/issues/5056 could help. It would allow us to call `chunk` on the inputs but not worry about potential conflicts with the target chunks.

rabernat:
It's almost always the 3.8 build that hangs. Does that tell us anything?

PR #81, opened 2021-03-03
Support incremental appending

davidbrochart:
See https://github.com/pangeo-forge/pangeo-forge/issues/37#issuecomment-782291347

### DIFF
--- a/pangeo_forge/recipe.py
+++ b/pangeo_forge/recipe.py
@@ -132,6 +132,7 @@ class NetCDFtoZarrRecipe(BaseRecipe):
     :param sequence_dim: The dimension name along which the inputs will be concatenated.
     :param inputs_per_chunk: The number of inputs to use in each chunk.
     :param nitems_per_input: The length of each input along the `sequence_dim` dimension.
+    :param processed_input_nb: The number of inputs already used to generate the existing dataset.
     :param target: A location in which to put the dataset. Can also be assigned at run time.
     :param input_cache: A location in which to cache temporary data.
     :param require_cache: Whether to allow opening inputs directly which have not
@@ -154,6 +155,7 @@ class NetCDFtoZarrRecipe(BaseRecipe):
     sequence_dim: Optional[str] = None
     inputs_per_chunk: int = 1
     nitems_per_input: int = 1
+    processed_input_nb: int = 0
     target: Optional[AbstractTarget] = field(default_factory=UninitializedTarget)
     input_cache: Optional[AbstractTarget] = field(default_factory=UninitializedTarget)
     require_cache: bool = True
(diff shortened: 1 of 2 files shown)


pangeo_forge/recipe.py
```diff
@@ -392,15 +405,20 @@ class NetCDFtoZarrSequentialRecipe(NetCDFtoZarrRecipe):
...
     input_urls: Iterable[str] = field(repr=False, default_factory=list)
 
     def __post_init__(self):
         super().__post_init__()
-        input_pattern = ExplicitURLSequence(self.input_urls)
-        self._inputs = {k: v for k, v in input_pattern}
+        self.processed_input_nb = len(self.processed_input_urls)
+        input_pattern = ExplicitURLSequence(self.processed_input_urls + self.input_urls)
+        self._inputs = {
+            k: {"url": v, "processed": k < self.processed_input_nb} for k, v in input_pattern
```
davidbrochart, line 420: The implementation of `ExplicitURLSequence` leaks here, as `k` would not necessarily be an incrementing index. We could instead have:
```python
"processed": v in self.processed_input_urls
```

davidbrochart: Actually this is not true, since `self.input_urls` are the new URLs (they don't contain the already processed URLs), which is actually different than for the `NetCDFtoZarrMultiVarSequentialRecipe`.

pangeo_forge/recipe.py
```diff
@@ -420,15 +438,21 @@ class NetCDFtoZarrMultiVarSequentialRecipe(NetCDFtoZarrRecipe):
...
+    processed_input_urls: Iterable[str] = field(repr=False, default_factory=list)
     input_pattern: VariableSequencePattern = field(default_factory=VariableSequencePattern)
 
     def __post_init__(self):
         super().__post_init__()
         self._variables = self.input_pattern.keys["variable"]
-        self._inputs = {k: v for k, v in self.input_pattern}
+        self.processed_input_nb = len(self.processed_input_urls) // len(self._variables)
+        self._inputs = {
+            k: {"url": v, "processed": v in self.processed_input_urls}
```
davidbrochart, line 453: Here the `input_pattern` generates the input file URLs including the already processed ones (which is different than for the `NetCDFtoZarrSequentialRecipe`). This allows to tag the already processed URLs.
If the `input_pattern` didn't generate the already processed URLs, the pattern would have to know about the (number of) already processed URLs, which is not great.

pangeo_forge/recipe.py
davidbrochart: We could change `self.input_urls` so that it also includes the already processed URLs, this would make `NetCDFtoZarrSequentialRecipe` similar to `NetCDFtoZarrMultiVarSequentialRecipe`.

rabernat:
Hi @davidbrochart - thanks so much for this PR! We really appreciate it! Sorry it has taken me so long to reply--I had a crazy week last week and wasn't able to spend much time on Pangeo Forge.

I'm going to try to review this in detail in the next few days. I'm a bit concerned about the intersection with #78, but we will see.

rabernat:
David thanks for working on this. We really appreciate your efforts.

Closed this because the code base has evolved quite a bit since this PR started. I plan to work on this feature in January and will use this PR as a basis for that work.

Appreciate your patience with this high-churn code base. It hasn't been trivial to get the base abstractions right.

PR #82, opened 2021-03-04
Example of a Pangeo-Forge-type workflow using Apache Beam

shoyer:
This doesn't use the Pangeo-Forge library (yet), but may be a useful for
design discussions.

When I run this on Google's internal infrastructure, it processes a 1 TB
dataset from hundreds of netCDF files into Zarr (all already in
distributed storage) in 20 minutes or so. Presumably one could achieve
similar performance with Cloud Dataflow.

Compared to the existing `NetCDFtoZarrMultiVarSequentialRecipe`:

- My equivalent of `prepare_target` (`write_zarr_metadata`) is performed
  in parallel, on different workers. This is quite helpful for datasets
  with moderately large numbers of variables and/or large (multi-GB)
  input files. Otherwise waiting on downloading all of the source files
  to the same machine to figure out the metadata becomes the slowest
  step. This would probably require some modifications to the
  "PipelineExecutor" model, to allow for collecting distributed inputs
  for a downstream task.
- I expand the time-dimension using `xarray.concat` rather than writing
  and then modifying a Zarr array. This lets me write the metadata with
  `consolidated=True`, which removes my need for anything like
  `finalize_target`.
- Because I create the time-variable in the destination Zarr file before
  writing chunks and use xarray's indexing to lookup the right `region`,
  I can handle variable length inputs (e.g., data stored in one file
  per real-world month or year, which can have variable length).
- By the looks of it, the existing recipes were designed to handle the
  case of many input files -> one zarr chunk. My recipe was designed for
  the opposite (one file -> many chunks). I think the general template
  you have works for both, but the names `iter_chunks` and `store_chunk`
  may be a little confusing because a "pangeo forge chunk" may
  correspond to one or more "zarr chunks".

### DIFF
--- a/examples/beam_netcdf_to_zarr.py
+++ b/examples/beam_netcdf_to_zarr.py
@@ -1,0 +1,183 @@
+"""Example of a "NetCDF Zarr Multi-Variable Sequential Recipe" using Beam.
+"""
+import contextlib
+import functools
+import logging
+from typing import Callable, List, Sequence, TypeVar
+
+from absl import app
+from absl import flags
+import apache_beam as beam
+import pandas as pd
+import xarray
+
+# Google-internal libaries
+from XXX import gfile  # a library that can access either local or remote files
+from XXX import zarr_gfile_store  # a zarr store built on top of gfile
+from XXX import file_util
+
+
+# pylint: disable=logging-format-interpolation
+
+EXAMPLE_PATTERN = flags.DEFINE_multi_string(
+    'example_pattern', '',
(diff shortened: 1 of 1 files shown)


rabernat:
Thanks so much for this Stephan! Super interesting and useful.

I have a big deadline this week, but I plan to dig into this next week and see how we can incorporate some of your insights.

rabernat:
Now that we have the beam executor, I'm going to close this for the sake of cleaning up the list of pending PRs. Thanks a lot Stephan for getting the discussion started.

PR #83, opened 2021-03-16
Try tests with per-thread loop for http

martindurant:
ref: #76

### DIFF
--- a/ci/py3.7.yml
+++ b/ci/py3.7.yml
@@ -9,7 +9,7 @@ dependencies:
   - codecov
   - dask
   - distributed
-  - fsspec
+  # - fsspec
   - h5netcdf
   - h5py
   - hdf5
@@ -31,3 +31,4 @@ dependencies:
   - zarr>=2.6.0
   - pip:
     - git+https://github.com/rabernat/rechunker.git@refactor-executors
+    - git+https://github.com/martindurant/filesystem_spec.git@ioloop_massage2
--- a/ci/py3.8.yml
+++ b/ci/py3.8.yml
@@ -9,7 +9,7 @@ dependencies:
   - codecov
   - dask
   - distributed
-  - fsspec
+  # - fsspec
   - h5netcdf
   - h5py
   - hdf5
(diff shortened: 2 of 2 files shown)


martindurant:
Well, everything passes except for something with blosc

davidbrochart:
Let's restart the jobs, failures were kind of random, but that's great news. So there was a bug in fsspec?

martindurant:
> So there was a bug in fsspec?

Not exactly. The event loop was being run in a particular event loop, and all IO channelled through it, while the thread launching the IO work would wait. Event loops are not thread-safe, but this mostly worked so long as there were no other event loops around. I think actually it was a change in aiohttp that made the difference, that they would require to run their requests in an `sync with` block, whereas before `await` would do. In any case, there is now one event loop per thread, which starts and stops as IO stuff is needed.

davidbrochart:
Blosc error again, restarting.

davidbrochart:
> In any case, there is now one event loop per thread, which starts and stops as IO stuff is needed.

I'm evaluating this technique in [jupyter-client](https://github.com/jupyter/jupyter_client), but it's not an ideal solution either. In the end I've found that having the sync API run async funtions until complete is the best, like we did in [nbclient](https://github.com/jupyter/nbclient), but that requires using [nest-asyncio](https://github.com/erdewit/nest_asyncio) which patches the event loop, so not an ideal solution either. I wish reentrant event loops were possible in standard asyncio, but [core devs are not in favor of it](https://bugs.python.org/issue22239).

martindurant:
> using nest-asyncio which patches the event loop

I don't think that would be acceptable to fsspec users, unfortunately, because they may be using those async-related classes for other things

> I wish reentrant event loops were possible in standard asyncio

Me too!

`fsspec.async._run_until_done` is my internal nested async runner; but I intend to rewrite a bit to remove its need.

rabernat:
So is our conclusion that the fsspec fix will fix our problem? If so, great!

davidbrochart:
It looks like it, let's relaunch the tests a few times to make sure.

davidbrochart:
I ran the tests a dozen times and never saw them hanging.

rabernat:
I'm curious why this works but #85 does not.

Could someone rebase this PR on the latest master to see if the issue is with the new tests introduced in #78?

rabernat:
This seems to be resolved in latest master so closing.

PR #85, opened 2021-03-21
update CI envs to use fsspec master + python 3.9

rabernat:
If this works, it supersedes #83.

### DIFF
--- a/ci/py3.9.yml
+++ b/ci/py3.9.yml
@@ -20,7 +20,8 @@ dependencies:
   - pip
   - prefect
   - pydap
-  - pynio
+# bring back eventually once pynio conda-forge package supports py3.9
+#  - pynio
   - pytest
   - pytest-cov
   - pytest-lazy-fixture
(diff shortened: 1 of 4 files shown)


rabernat:
Well that did not go well. I canceled after 50m...

rabernat:
This works fine on my macbook. Frustrated by CI again.

ci/py3.9.yml
```diff
@@ -0,0 +1,35 @@
...
+  - h5py
+  - hdf5
+  - lxml    # Optional dep of pydap
+  - netcdf4
+  - numpy
+  - pandas
+  - pip
+  - prefect
+  - pydap
+  - pynio
```
andersy005, line 23: As far I can tell, the latest build of pynio on conda-forge supports `Python <=3.8`. Remove it from the environment fixes the issue (tested on a linux system). 

```suggestion
```

rabernat:
Yay! One test successfully finished. The other two are hung I think.

Issue #41, opened 2021-01-22
GRIB inputs


rabernat:
There are lots of potential recipes that use GRIB as inputs (e.g. https://github.com/pangeo-forge/staged-recipes/issues/15).

It should be easy to support them, since xarray can read grib. However, grib can't be opened over the network; it has to be downloaded. Furthermore, @rsignell noted that 

> It turns out that instead of reading the grib2 files with engine=cfgrib, it's faster to download the grib2, convert them to netcdf using wgrib2 and then load the resulting netcdf file into xarray.

This raises two design questions about pangeo forge:
- How do we deal with situations that require local file storage? Can we assume workers have a local disk and tmp storage?
- What's the best way to call command line tools like wgrib2 from pangeo forge?
- How do we make sure these dependencies are on remote workers?

PR #87, opened 2021-03-23
Add copy_input_to_local_file option

rabernat:
This PR makes two changes:

- Rename `require_cache` option to `cache_inputs`. If this option is False, then the pipeline generation now skips the `cache_input` loop entirely.
- Add an `copy_input_to_local_file`. This causes the input to be copied to a **local** temporary file before opening. This will enable us to open files that have to be opened via a local path (e.g. GRIB, #41) rather than a file-like object.

I swear the tests pass on my local machine, but I assume they will time out in CI.

### DIFF
--- a/pangeo_forge/recipe.py
+++ b/pangeo_forge/recipe.py
@@ -4,6 +4,8 @@ A Pangeo Forge Recipe
 
 import json
 import logging
+import os
+import tempfile
 from abc import ABC, abstractmethod
 from contextlib import ExitStack, contextmanager
 from dataclasses import dataclass, field
@@ -107,7 +109,8 @@ class BaseRecipe(ABC):
         """
 
         pipeline = []  # type: MultiStagePipeline
-        pipeline.append(Stage(self.cache_input, list(self.iter_inputs())))
+        if getattr(self, "cache_inputs", False):  # TODO: formalize this contract
+            pipeline.append(Stage(self.cache_input, list(self.iter_inputs())))
         pipeline.append(Stage(self.prepare_target))
         pipeline.append(Stage(self.store_chunk, list(self.iter_chunks())))
         pipeline.append(Stage(self.finalize_target))
@@ -144,6 +147,22 @@ def _copy_btw_filesystems(input_opener, output_opener, BLOCK_SIZE=10_000_000):
                 target.write(data)
 
 
+@contextmanager
+def _maybe_open_or_copy_to_local(opener, copy_to_local, orig_name):
+    _, suffix = os.path.splitext(orig_name)
+    if copy_to_local:
+        ntf = tempfile.NamedTemporaryFile(suffix=suffix)
+        tmp_name = ntf.name
+        logger.info(f"Copying {orig_name} to local file {tmp_name}")
+        target_opener = open(tmp_name, mode="wb")
+        _copy_btw_filesystems(opener, target_opener)
+        yield tmp_name
+        ntf.close()  # cleans up the temporary file
+    else:
+        with opener as fp:
+            yield fp
+
+
 # Notes about dataclasses:
 # - https://www.python.org/dev/peps/pep-0557/#inheritance
 # - https://stackoverflow.com/questions/51575931/class-inheritance-in-python-3-7-dataclasses
@@ -171,7 +190,7 @@ class NetCDFtoZarrRecipe(BaseRecipe):
     :param input_cache: A location in which to cache temporary data.
     :param metadata_cache: A location in which to cache metadata for inputs and chunks.
       Required if ``nitems_per_input=None``.
-    :param require_cache: Whether to allow opening inputs directly which have not
+    :param cache_inputs: Whether to allow opening inputs directly which have not
       yet been cached. This could lead to very unstanble behavior if the inputs
       live behind a slow network connection.
     :param copy_input_to_local_file: Whether to copy the inputs to a temporary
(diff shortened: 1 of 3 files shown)


pangeo_forge/recipe.py
```diff
@@ -139,6 +135,45 @@ def _chunk_metadata_fname(chunk_key) -> str:
...
+@contextmanager
+def _fsspec_safe_open(fname, **kwargs):
+    # workaround for inconsistent behavior of fsspec.open
+    # https://github.com/intake/filesystem_spec/issues/579
+    with fsspec.open(fname, **kwargs) as fp:
+        if isinstance(fp, fsspec.implementations.local.LocalFileOpener):
+            with fp as fp2:
+                yield fp2
+        else:
+            yield fp
```
rabernat, line 174: The need for this helper function is due to https://github.com/intake/filesystem_spec/issues/579.

martindurant: What error did you get without this? The `LocalFileOpener` supports all of the methods that an `_io.BufferedReader` would

rabernat:
> Note that you may want to consider how to upstream some of this kind of thing to fsspec - which already has a cache mechanism, of course.

You've suggested this several times. I'd like to know more about how it might work. I'm open to the critique that I'm reinventing caching here. But I'm not sure I understand fsspec's caching well enough to know whether it meets our needs.

There are two distinct levels of caching used in Pangeo Forge

1. The input files are [optionally] systematically cached to _globally accessible_ storage, e.g. each files is saved in a GCS bucket. This is necessary to support robust extraction of data from slow / unreliable FTP servers. Before building the Zarr target dataset, we need to know that we have all the input files accessible close to our compute. This needs to be globally accessible because we will have reads coming from many different workers at the later stages of the pipeline.
2. When opening the input files with xarray (whether cached via 1 or not), the input file is [optionally] copied to a local tmp directory, such that we can pass a local file path (str) to `xarray.open_dataset`. This is required for some backends.

My impression is that fsspec can probably handle 2 but not 1. Is that correct?

This PR implements 2.

If you would like to implement 2 via fsspec-based caching, I would gladly accept that PR. It should have the same API (`copy_input_to_local_file` option) and functionality as this one.

pangeo_forge/recipe.py
rabernat:
```
ValueError                                Traceback (most recent call last)
<ipython-input-77-951e7a504e4c> in <module>
      1 with of as fp:
----> 2     xr.open_dataset(fp)

~/Code/xarray/xarray/backends/api.py in open_dataset(filename_or_obj, engine, chunks, cache, decode_cf, mask_and_scale, decode_times, decode_timedelta, use_cftime, concat_characters, decode_coords, drop_variables, backend_kwargs, *args, **kwargs)
    491 
    492     if engine is None:
--> 493         engine = plugins.guess_engine(filename_or_obj)
    494 
    495     backend = plugins.get_backend(engine)

~/Code/xarray/xarray/backends/plugins.py in guess_engine(store_spec)
    108             logging.exception(f"{engine!r} fails while guessing")
    109 
--> 110     raise ValueError("cannot guess the engine, try passing one explicitly")
    111 
    112 

ValueError: cannot guess the engine, try passing one explicitly
```

martindurant:
Number 1 is this issue https://github.com/intake/filesystem_spec/issues/494 - would not be too hard to implement, but it isn't done. Indeed, fsspec's caching is currently local, Number 2. Careful thought around https://github.com/intake/filesystem_spec/issues/555 would be good too.

pangeo_forge/recipe.py
martindurant: I would say this is an xarray problem: you should not require a file-like object to be a subclass of io.IOBase. In other locations, it just does `hasattr(obj, "read")` instead.

rabernat: Ok good to know. Could you raise an xarray issue for this?

rabernat: But just to clarify the source of my confusion, if I call `fsspe.open` the file with `mode='r'` (rather than `mode='rb'`), the object I get _is_ a subclass of ` io.IOBase`!

rabernat:
For 2, we don't care about the file name at this stage, so perhaps https://github.com/intake/filesystem_spec/issues/555 is not relevant?

pangeo_forge/recipe.py
martindurant: Yes, this is true. If you wanted LocalFileOpener to be a subclass, you would need a load of methods implemented like
```python
    def read(self, size=-1):
        return self.f.read(size)
```
**or** lose the serialisable trait by making a `BufferedReader` directly.

martindurant:
> we don't care about the file name at this stage

Maybe it *does* make sense to keep hold of the names/directory structure, though, if this "long-term" cache is to be accessed directly at all.

rabernat:
> if this "long-term" cache is to be accessed directly at all.

If we are talking about 2 (the subject of this PR), this is NOT long-term cache, nor is it globally accessible. It's just a one-time copy for formats that can't be read from file-like objects, like GRIB.

I see 1 as much more complex, probably requiring a significant refactor of lots of existing code. So let's keep those things separate.

If you can show me how to do what this PR does (item 2 from my list) using fsspec's cache instead, and if it is simpler / more robust / fewer lines of code, then I would gladly use that instead of the implementation here.

martindurant:
OK, I'll do that this afternoon.

martindurant:
(but, yes agree that the naming convention is only material for the Number 1 cache ; should fsspec implement this, the "layout" option would probably come in the same PR)

rabernat:
FYI, our current solution to the naming problem is here

https://github.com/pangeo-forge/pangeo-forge/blob/3314162568f7d8fe3d6f06627b936cee55d99c9c/pangeo_forge/storage.py#L82-L87

https://github.com/pangeo-forge/pangeo-forge/blob/3314162568f7d8fe3d6f06627b936cee55d99c9c/pangeo_forge/storage.py#L96-L103

PR #96, opened 2021-04-06
Allow uninitialized input_cache when cache_inputs=False

rabernat:
Catches a little edge case that was missed in #87.


rabernat:
This had a hung test: https://github.com/pangeo-forge/pangeo-forge/pull/96/checks?check_run_id=2274781050#step:8:27

Restarting

PR #102, opened 2021-04-14
Don't overwrite cache

rabernat:
This adds a check to prevent caching if it looks like the file is already there.


rabernat:
Bit by the same [cfgrib issue](https://github.com/pangeo-forge/pangeo-forge/pull/101#issuecomment-819603046).

pangeo_forge/recipe.py
```diff
@@ -172,6 +172,11 @@ def _fsspec_safe_open(fname, **kwargs):
         with fp as fp2:
             yield fp2
 
+            
+def _get_url_size(fname):
+    with fsspec.open(fname, mode='rb') as of:
+        size = of.fs.size(fname)
```
martindurant, line 178: I think `of.size` should also exist as an attribute and potentially avoid a lookup

rabernat:
I apparently made a mess of the commit history by rebasing this on master. Whoops.

PR #108, opened 2021-04-26
Update required xarray to v0.17.1 (currently unreleased)

rabernat:
I'm not sure this will even pass CI, let's see.


rabernat:
I will merge this PR, do a release, and then revert the change on master.

PR #119, opened 2021-05-03
Renaming

### LINKED ISSUE #75: Rename this package?
Today's discussion / whiteboarding was very useful. It surfaced another point of confusion about naming.

We are currently using the term "Pangeo Forge" for two distinct things
- The overall project
- The `pangeo_forge` python library

Is this confusing? We have not actually released the package yet, so it's really not too late to rename it. Is `pangeo_forge` actually `pangeo_smithy`? Or should we break with them and call it `pangeo_recipe`.

cisaacstern:
Closes #75

Here is a second attempt at making this PR. I've now updated the package name in `.github/workflows/main.yaml`, which is a guess at making the tests run correctly.

In addition to underscored name instances, I’ve also changed dashed name instances (i.e. ‘pangeo-forge’ became ‘pangeo-forge-recipes’). These dashed names are primarily in the README, but also appear in setup.py and perhaps a few other places.

Local testing indicates that, following these changes, the only failed test is test_fixture_local_files_by_variable (line 17 in tests/test_fixtures.py). I note, however, that this test is marked for replacement with a TODO comment, so perhaps this is not of great concern.

A final note on methodology, for the sake of record keeping: all of these name changes have been made with various versions of the command `find . -type f -print0 | xargs -0 sed -i '' -e 's/pangeo_forge/pangeo_forge_recipes/g'`, referenced from [this superuser.com thread](https://superuser.com/questions/428493/how-can-i-do-a-recursive-find-and-replace-from-the-command-line). Notably, I did not make all changes in one command from the top-level directory, as this appeared to break something in the git index; instead commands were executed in each relevant sub-directory one at a time. In the case of changes to individual files in the top-level directory, the -name flag was used, as demonstrated in the linked thread.


cisaacstern:
`black` is failing the pre-commit due to line lengths, which are extended by the renaming.

Closing this PR to correct and resubmit.

PR #122, opened 2021-05-04
Renaming

### LINKED ISSUE #75: Rename this package?

cisaacstern:
Closes #75

A fourth (and hopefully final) attempt at this PR. In the second attempt, updating the package name in `.github/workflows/main.yaml` allowed tests to pass. In the third attept, wrapping a few lengthened import statements in `tests/` got pre-commit linting closer to passing. Trailing whitespace trimmed in this attempt should ideally get pre-commit to pass.

In addition to underscored name instances, I’ve also changed dashed name instances (i.e. ‘pangeo-forge’ became ‘pangeo-forge-recipes’). These dashed names are primarily in the README, but also appear in setup.py and perhaps a few other places.

Local testing indicates that, following these changes, the only failed test is test_fixture_local_files_by_variable (line 17 in tests/test_fixtures.py). I note, however, that this test is marked for replacement with a TODO comment, so perhaps this is not of great concern.

A final note on methodology, for the sake of record keeping: all of these name changes have been made with various versions of the command `find . -type f -print0 | xargs -0 sed -i '' -e 's/pangeo_forge/pangeo_forge_recipes/g'`, referenced from [this superuser thread](https://superuser.com/questions/428493/how-can-i-do-a-recursive-find-and-replace-from-the-command-line). Notably, I did not make all changes in one command from the top-level directory, as this appeared to break something in the git index; instead commands were executed in each relevant sub-directory one at a time. In the case of changes to individual files in the top-level directory, the `-name` flag was used, as demonstrated in the linked thread.


docs/bakeries.md
```diff
@@ -8,9 +8,9 @@ Cloud based execution is provided by Bakeries.
 
 There are currently two types of Bakeries implemented:
 
-- <https://github.com/pangeo-forge/pangeo-forge-aws-bakery> -
+- <https://github.com/pangeo-forge-recipes/pangeo-forge-recipes-aws-bakery> -
```
rabernat, line 11: So here it looks like you are also changing the name of the github org (`github.com/pangeo-forge`)  and another associated repo (`pangeo-forge-aws-bakery`). We _only_ want to rename this repo, not the org, nor the other project. So I would revert this.

docs/bakeries.md
```diff
@@ -8,9 +8,9 @@ Cloud based execution is provided by Bakeries.
...
   Bakery for Amazon Web Services cloud
-- <https://github.com/pangeo-forge/pangeo-forge-azure-bakery> -
+- <https://github.com/pangeo-forge-recipes/pangeo-forge-recipes-azure-bakery> -
```
rabernat, line 13: Same.

docs/conf.py
```diff
@@ -31,14 +31,14 @@
 # https://sphinx-book-theme.readthedocs.io/en/latest/configure.html
 html_theme = "sphinx_book_theme"
 html_theme_options = {
-    "repository_url": "https://github.com/pangeo-forge/pangeo-forge",
+    "repository_url": "https://github.com/pangeo-forge-recipes/pangeo-forge-recipes",
```
rabernat, line 34: Same. Don't rename the org, just the repo.

docs/conf.py
```diff
@@ -31,14 +31,14 @@
...
     "repository_branch": "master",
     "path_to_docs": "docs",
     "use_repository_button": True,
     "use_issues_button": True,
     "use_edit_page_button": True,
 }
-html_logo = "_static/pangeo-forge-logo-blue.png"
+html_logo = "_static/pangeo-forge-recipes-logo-blue.png"
```
rabernat, line 41: Unless you renamed the file, this will break the docs.

docs/recipe_box.md
```diff
@@ -19,13 +19,13 @@ You will also need a thorough understanding of the source dataset you wish to in
 and a vision for the desired target dataset (desired format, chunk structure, etc.)
 and any cleaning / processing steps needed along the way.
 
-Once you are ready to begin, head over to <https://github.com/pangeo-forge/staged-recipes>.
+Once you are ready to begin, head over to <https://github.com/pangeo-forge-recipes/staged-recipes>.
 This repo is used to triage all new recipes for including in the Recipe Box.
 The basic steps for adding you recipe are
 
-1. [fork](https://docs.github.com/en/free-pro-team@latest/github/getting-started-with-github/fork-a-repo) the <https://github.com/pangeo-forge/staged-recipes> GitHub repository
+1. [fork](https://docs.github.com/en/free-pro-team@latest/github/getting-started-with-github/fork-a-repo) the <https://github.com/pangeo-forge-recipes/staged-recipes> GitHub repository
```
rabernat, line 26: Don't rename the org.

docs/tutorials/cmip6-recipe.ipynb
```diff
@@ -875,7 +875,7 @@
    "metadata": {},
    "source": [
     "## Postscript\n",
-    "- If you find a CMIP6 dataset for which this recipe does not work, Please report it at [issue#105](https://github.com/pangeo-forge/pangeo-forge/issues/105) so we can refine the recipe, if possible.\n"
+    "- If you find a CMIP6 dataset for which this recipe does not work, Please report it at [issue#105](https://github.com/pangeo-forge-recipes/pangeo-forge-recipes/issues/105) so we can refine the recipe, if possible.\n"
```
rabernat, line 878: Don't rename the org.

setup.py
```diff
@@ -24,21 +24,21 @@
...
+    name="pangeo-forge-recipes",
     description="Pipeline tools for building and publishing analysis ready datasets",
     long_description=long_description,
     long_description_content_type="text/markdown",
     python_requires=">=3.7",
     maintainer="Joe Hamman",
     maintainer_email="jhamman@ucar.edu",
     classifiers=CLASSIFIERS,
-    url="https://github.com/pangeo-forge/pangeo-forge",
+    url="https://github.com/pangeo-forge-recipes/pangeo-forge-recipes",
```
rabernat, line 35: Don't rename the org.

rabernat (changes requested):
Excellent! Thanks @cisaacstern! Really impressed by your quick work here.

The only problem I see is probably due to my unclear instructions: we don't want to rename the whole `pangeo-forge` github organization (or any of the other sub-projects). We just want to rename _this repo_. In fact, if we rename everything, we will be back where we started, with no way to differentiate this python package from the broader project. 🙃 

Once you fix those instances, we will be good to go.

rabernat:
FYI, it looks like we had a hung test, as happens from time to time. I restarted them.

cisaacstern:
@rabernat, the accidental org-renaming was the result of changing all `'pangeo-forge'` strings to `'pangeo-forge-recipes'`. While reversing this blanket change was a first step, I subsequently realized that there are certain edge cases where we do want to make this change: e.g., in the topline header of the README and, more frequently, in various places that link to (what will now be) https://github.com/pangeo-forge/pangeo-forge-recipes.

Beyond these obvious cases, there were a handful of edge cases where I was not entirely certain. I'll list the choices I made here, and you can decide if any of them should be corrected before merging:

1. In `/pangeo_forge_recipes/utils.py` the function `lock_for_conflicts` defined on line 87 uses the default kwarg `base-name="pangeo-forge"`. I left this as-is, in part because I wasn't clear on the purpose of this kwarg, and in part because it's marked for change with an in-line TODO comment.
2. In `Dockerfile` the final two lines now use `pangeo-forge-recipes`, which I believe is correct because the repo name itself will be changed to that, once this PR is merged.
3. In all three `.yml` files within the `/ci` directory, I've renamed the specified environments to `"pangeo-forge-recipes"`
4. In `setup.py`, the `name=` and `entry_points=` kwargs now use `pangeo-forge-recipes`, which was sort of just a best guess.
5. In `.github/workflows/main.yaml`, this PR now has `activate-environment: pangeo-forge-recipes`, which I believe makes it consistent with the changed env names in the `/ci` directory. In the same file, `- name: install pangeo-forge-recipes` was another best guess.
6. The string `"pangeo-forge"` pops up in a few of the tutorial notebooks as part of RuntimeWarnings. These are presumably part of notebook outputs, so I've left them unchanged.
7. The string `"pangeo-forge3.8"` appears twice as part of metadata in `/docs/tutorials/cmip6-recipe.ipynb`. This is presumably a kernel and/or environment name, but since it's the only place where I encountered this specific string, and I wasn't sure of its precise role, it sort of threw me off and I just left it as-is.

On a meta level, after writing this list, I'm left wondering if in the future it would be better to provide file- and line-specific comments like this in-line in the PR diff. Let me know if that would be stylistically preferable and I can look into how to adjust for future PRs.

rabernat:
> 1. In `/pangeo_forge_recipes/utils.py` the function `lock_for_conflicts` defined on line 87 uses the default kwarg `base-name="pangeo-forge"`. I left this as-is, in part because I wasn't clear on the purpose of this kwarg, and in part because it's marked for change with an in-line TODO comment.

Correct, this name doesn't matter. The name here is used to synchronize overlapping writes among different processes within a dask cluster.

> 2\. In `Dockerfile` the final two lines now use `pangeo-forge-recipes`, which I believe is correct because the repo name itself will be changed to that, once this PR is merged.

These are actually not used any more and should be removed from the repo. I'll do that soon.

PR #133, opened 2021-05-12
Refactor storage

rabernat:
The goal of this refactor is to _remove code_ from `recipes/xarray_zarr.py` and _add code_ to `storage.py`. The aim of this is to improve our abstractions around file handling, such that most of the generic logic related to downloading, caching, and copying files lives in the `storage` module.

The main new additions are:
- A context manager function called `storage.file_opener` which takes care of downloading, caching, and copying to local tmpdir
- A new `MetadataTarget` class which serializes / deserializes data from / to python dictionaries. (We were already doing this in `xarray_zarr.py` but now it lives in `storage`
- Better type hints, which caught some edge cases
- Removal of the `UninitializedTarget` class. I'm sure there was some reason for this at some point, but it seems unnecessary now.

### DIFF
--- a/pangeo_forge_recipes/recipes/xarray_zarr.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr.py
@@ -202,7 +202,7 @@ class XarrayZarrRecipe(BaseRecipe):
     @closure
     def prepare_target(self) -> None:
         if self.target is None:
-            raise ValueError("target has not been set.")
+            raise ValueError("target is not set.")
         try:
             ds = self.open_target()
             logger.info("Found an existing dataset in target")
@@ -264,7 +264,7 @@ class XarrayZarrRecipe(BaseRecipe):
 
         if self._cache_metadata:
             if self.metadata_cache is None:
-                raise ValueError("metadata_cache has not been set")
+                raise ValueError("metadata_cache is not set")
             # if nitems_per_input is not constant, we need to cache this info
             recipe_meta = {"input_sequence_lens": input_sequence_lens}
             self.metadata_cache[_GLOBAL_METADATA_KEY] = recipe_meta
@@ -275,7 +275,7 @@ class XarrayZarrRecipe(BaseRecipe):
     def cache_input(self, input_key: InputKey) -> None:  # type: ignore
         if self.cache_inputs:
             if self.input_cache is None:
-                raise ValueError("input_cache has not been set.")
+                raise ValueError("input_cache is not set.")
             logger.info(f"Caching input '{input_key}'")
             fname = self.file_pattern[input_key]
             self.input_cache.cache_file(fname, **self.fsspec_open_kwargs)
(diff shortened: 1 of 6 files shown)


pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -140,9 +89,9 @@ class XarrayZarrRecipe(BaseRecipe):
     file_pattern: FilePattern
     inputs_per_chunk: Optional[int] = 1
     target_chunks: Dict[str, int] = field(default_factory=dict)
-    target: AbstractTarget = field(default_factory=UninitializedTarget)
-    input_cache: AbstractTarget = field(default_factory=UninitializedTarget)
-    metadata_cache: AbstractTarget = field(default_factory=UninitializedTarget)
+    target: Optional[AbstractTarget] = None
```
cisaacstern, line 92: Is the decision to define a default with `field(default_factory=...)` vs. as `None` a consequence of whether or not we want the user to be able to access certain methods of a default type vs. just overwrite the default value?

pangeo_forge_recipes/storage.py
```diff
@@ -98,7 +131,74 @@ def _full_path(self, path: str) -> str:
...
+        return json.loads(self.get_mapper()[key])
+
+    def getitems(self, keys: Sequence[str]) -> dict:
+        mapper = self.get_mapper()
+        all_meta_raw = mapper.getitems(keys)
+        return {k: json.loads(raw_bytes) for k, raw_bytes in all_meta_raw.items()}
+
+
+@contextmanager
+def file_opener(
```
cisaacstern, line 168: Once merged, will this new method mean re-writing the **Step 5: Create storage targets** cell in the CMIP6 tutorial in some way? If so, I'll add this as a TODO for #135.

pangeo_forge_recipes/recipes/xarray_zarr.py
rabernat: No. TBH I can't remember the original reason why I thought I had to use `UninitializedTarget`. It may have had to do with type hints. Before, all these objects _had_ to be `AbstractTarget`. Now they are `Optional[AbstractTarget]` or `Optional[MetadataTarget]`, etc. The `Optional` type hint means that `None` is a valid value. This is simpler and more in line with standard python best practices.

One downstream effect, however, is that we need statements like `if self.target is not None` in various places in order to keep mypy happy.

pangeo_forge_recipes/storage.py
rabernat: This function (`file_opener`) is not something seen by the users. It's purely internal.

However, your comment makes me realize that there _is_ one user-facing change needed as a result of this. In step 5, instead of 
```python
meta_store = FSSpecTarget(fs_local, meta_dir.name)
```
we would need
```python
meta_store = MetadataTarget(fs_local, meta_dir.name)
```

cisaacstern: Great. Added this to item to #135.

PR #117, opened 2021-05-03
Serialize file patterns

rabernat:
This PR potentially closes #116 by making the FilePattern object smaller in serialized form.

PR #139, opened 2021-05-19
Add copy_pruned() method to XarrayZarrRecipe

### LINKED ISSUE #97: Testing / profiling recipes
We need a way to validate whether recipes "work" well enough without running the whole thing.

Ideally this would work as follows:
1. Figure out the minimum number of inputs needed to initialize the target
2. Prune the recipe inputs to include just these
3. Run the recipe as normal
4. Return a dataset / temporary directory for inspection

Ideally we would also measure disk usage and memory usage along the way.

rabernat:
### What I changed

This perhaps closes #97 by adding a new method called `copy_pruned` to the `XarrayZarrRecipe` class. The way you use this is

```python
rec = XarrayZarrRecipe(**kwargs)
rec_pruned = rec.copy_pruned()  # much smaller version of the same recipe
```

The way this works is by keeping the first `nkeep` files (default `nkeep=2`) along the the ConcatDim of the recipe. MergeDims, in contrast are not pruned. The thinking here is that we don't want to skip any variables when testing the recipe.

This does not guarantee that the recipe will be smaller than a specific size in MB. It simply naively reduces the number of input files.

### Next Step

If we like this feature, then we can incorporate it into a hypothetical recipe testing mechanism that can be used in our CI (see discussion in https://github.com/pangeo-forge/staged-recipes/pull/28, [ADR5](https://github.com/pangeo-forge/roadmap/blob/master/doc/adr/0005-administrative-commands-for-pr-testing.md))

(Note that this builds on #117, which ideally should be merged first.)

### DIFF
--- a/pangeo_forge_recipes/patterns.py
+++ b/pangeo_forge_recipes/patterns.py
@@ -62,11 +62,8 @@ class FilePattern:
       product of the keys is used to generate the full list of file paths.
     """
 
-    def __init__(self, format_function: Callable, *combine_dims: Union[MergeDim, ConcatDim]):
-
-        self.format_function = format_function
-        self.combine_dims = combine_dims
-
+    @staticmethod
+    def _make_da(format_function, combine_dims) -> xr.DataArray:
         dim_names = [cdim.name for cdim in combine_dims]
         fnames = []
         for keys in product(*[cdim.keys for cdim in combine_dims]):
@@ -78,7 +75,17 @@ class FilePattern:
         # This way of defining coords is incompatible with xarray type annotations.
         # I don't understand why.
         coords = {cdim.name: (cdim.name, cdim.keys) for cdim in combine_dims}
-        self._da = xr.DataArray(fnames_np, dims=list(coords), coords=coords)  # type: ignore
+        return xr.DataArray(fnames_np, dims=list(coords), coords=coords)  # type: ignore
+
+    def __init__(self, format_function: Callable, *combine_dims: Union[MergeDim, ConcatDim]):
+        self.__setstate__((format_function, combine_dims))
+
+    def __getstate__(self):
+        return self.format_function, self.combine_dims
+
+    def __setstate__(self, state):
+        self.format_function, self.combine_dims = state
+        self._da = self._make_da(self.format_function, self.combine_dims)
 
     def __repr__(self):
         return f"<FilePattern {self.dims}>"
--- a/pangeo_forge_recipes/patterns.py
+++ b/pangeo_forge_recipes/patterns.py
@@ -2,7 +2,7 @@
 Filename / URL patterns.
 """
 
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace
 from itertools import product
 from typing import Any, Callable, Dict, Iterator, List, Optional, Sequence, Tuple, Union
 
@@ -175,10 +175,7 @@ def prune_pattern(fp: FilePattern, nkeep: int = 2) -> FilePattern:
             new_combine_dims.append(cdim)
         elif isinstance(cdim, ConcatDim):
             new_keys = cdim.keys[:nkeep]
-            # this feels like a fragile way to copy a DataClass but it works for now
-            new_cdim = ConcatDim(
-                name=cdim.name, keys=new_keys, nitems_per_file=cdim.nitems_per_file
-            )
+            new_cdim = replace(cdim, keys=new_keys)
             new_combine_dims.append(new_cdim)
         else:  # pragma: no cover
             assert "Should never happen"
--- a/pangeo_forge_recipes/patterns.py
+++ b/pangeo_forge_recipes/patterns.py
@@ -47,6 +47,7 @@ class MergeDim:
 
 
 Index = Tuple[int, ...]
+CombineDim = Union[MergeDim, ConcatDim]
 
 
 class FilePattern:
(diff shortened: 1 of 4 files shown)


pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -181,6 +181,17 @@ def _validate_input_and_chunk_keys(self):
...
             raise ValueError("_inputs_chunks and _chunks_inputs don't use the same chunk keys.")
 
+    def copy_pruned(self, nkeep: int = 2) -> BaseRecipe:
+        """Make a copy of this recipe with a pruned file pattern.
+
+        :param nkeep: The number of items to keep from each ConcatDim sequence.
+        """
+
+        new_pattern = prune_pattern(self.file_pattern, nkeep=nkeep)
+        return replace(self, file_pattern=new_pattern)
```
cisaacstern, line 191: That `dataclasses.replace` method is elegant. Hadn't seen it before.

tests/test_patterns.py
```diff
@@ -41,14 +44,22 @@ def test_pattern_from_file_sequence():
...
     def format_function(time, variable):
         return f"T_{time}_V_{variable}"
 
     fp = FilePattern(format_function, merge, concat)
+
+    if pickle:
+        # regular pickle doesn't work here because it can't pickle format_function
+        from cloudpickle import dumps, loads
+
+        fp = loads(dumps(fp))
```
cisaacstern, line 61: Generally speaking, what is pickling (or, cloudpickling, in this case?) achieving here? Pickling is one of those concepts I've seen around the Python world a lot, but haven't understood yet. (Google exists, of course, but thought it might be worth asking within our specific context.)

rabernat: Why don't you move your comment to #117, which is specifically about this.

cisaacstern (commented):
Just installed this PR branch locally and can confirm that the `copy_pruned()` method works for clipping the time (i.e. concat) dimension for a dict of "real world recipes". (See [this gist](https://gist.github.com/cisaacstern/baef5fb0d4305601692605cf494254a0) for a record, in which I am importing the `recipe-dict` from [this draft recipe]( https://github.com/cisaacstern/staged-recipes/blob/nasa-smap-sss/recipes/nasa-smap-sss/recipe.py).)

This works exactly as I'd hoped for in my comment at https://github.com/pangeo-forge/staged-recipes/pull/31#issuecomment-841522873. It's great to be able to great a pruned recipe for testing purposes without changing the intended production version of `recipe.py`. 

I will be manually executing these pruned recipes shortly and will follow-up with any relevant insights from that experience.

cisaacstern:
@sharkinsspatial, thinking now about our action points recorded in https://github.com/pangeo-forge/staged-recipes/issues/32#issuecomment-843715983. (Specifically, your suggestion to submit a pruned PR of NASA SMAP SSS to start.)

Now that the `copy_pruned()` method exists, is there a way that we could included something like
```yaml
recipes:
  - id: ...
     object: ...
     pruned: "True"
```
in the `meta.yaml` spec?

If pruning is something that we'll want to do frequently, perhaps this implementation will be less error-prone than changing `recipe.py`? I suppose there would need to be a provision on the bakery side for iteratively pruning if the object is a dictionary of recipes.

rabernat:
I don't think we should be putting this in `meta.yaml`. 

Instead, I think that the CI--whatever script / workflow we decide on for "checking" a recipe PR--should be constituting the full recipe and then make a pruned copy of it to test.

cisaacstern:
Makes sense. I'd forgotten that testing a pruned version of recipe will become part of the standard workflow.

For the test case I'll submit later today, I'll just do the pruning manually in `recipe.py`.

rabernat:
What I don't quite understand is _where should this testing workflow live_? That's question for @sharkinsspatial / @ciaranevans.

sharkinsspatial (commented):
@cisaacstern As @rabernat mentioned the use of `copy_pruned` can happen at the CI workflow level.  

1. I'll extend [pangeo-forge-prefect](https://github.com/pangeo-forge/pangeo-forge-prefect/blob/master/pangeo_forge_prefect/flow_manager.py#L214) to support a "testing" argument that will use the `copy_pruned` method of the recipe as well as enable debug logging for all `pangeo-forge-recipes` modules.
2. I'll include this testing argument in [recipe-prefect-action](https://github.com/pangeo-forge/recipe-prefect-action) so that the same action can be utilized to register both pruned and normal recipes.
3. I'll update the test workflow in [staged-recipes](https://github.com/pangeo-forge/staged-recipes/blob/master/.github/workflows/run-recipe-test.yaml#L40) to use the "testing" argument.  (I also should update this to ignore the bakery specified in the meta.yaml and use the designated development bakery"

pangeo_forge_recipes/patterns.py
```diff
@@ -150,3 +158,29 @@ def format_function(**kwargs):
...
+    """
+
+    new_combine_dims = []  # type: List[CombineDim]
+    for cdim in fp.combine_dims:
+        if isinstance(cdim, MergeDim):
+            new_combine_dims.append(cdim)
+        elif isinstance(cdim, ConcatDim):
+            new_keys = cdim.keys[:nkeep]
+            # this feels like a fragile way to copy a DataClass but it works for now
+            new_cdim = ConcatDim(
```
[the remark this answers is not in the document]
rabernat: Good idea!

rabernat:
The plan is to hopefully merge #117 and then this later today, then release.

Issue #140, opened 2021-05-20
Support for more than one ConcatDim


nbarlowATI:
Hi all, I'm trying to convert a dataset from NetCDF to Zarr, and would really like to concatenate over more than one dimension (in my case "time" and "ensemble_id").
I see that XarrayZarrRecipe/FilePattern doesn't currently support this.
I'd be happy to work on implementing this if others think it's worthwhile, and if there isn't already someone working on it?

martindurant:
I'd like to take the chance to once more gently remind that RefernceFileSystem really wants this logic: of figuring out for every chunk in the input, what chunk index it should have in a theoretical output zarr dataset. This process doesn't need to live inside a recipe class, but I suppose that's a good place to start.

rabernat:
Hi @nbarlowATI and welcome!

You are correct that this is not yet supported. You are welcome to try to implement it, and we would love to have your contribution. However, I am concerned that the current code related to this is not very clear or accessible to outside contributors. Consequently, you may become frustrated if you try to work on this. As Martin suggested, abstracting this logic into a standalone module would probably be wise; however that is also a difficult task.

I will try to give you some hints about how to proceed:

The key method to look at is `region_and_conflicts_for_chunk`:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/32e920185934e42baf32650c5b3d1e76ea078ab3/pangeo_forge_recipes/recipes/xarray_zarr.py#L412-L415

This method translates a `chunk_key` to a specific slice within a multidimensional dataset (e.g. `{'time': slice(10, 20)}`). `chunk_key` is a tuple of ints. It is a key internal data structure which encodes the chunk's place within the broader dataset. It is set here:
https://github.com/pangeo-forge/pangeo-forge-recipes/blob/32e920185934e42baf32650c5b3d1e76ea078ab3/pangeo_forge_recipes/recipes/xarray_zarr.py#L151
(This is the part of the code that is hardest to understand.) Currently `chunk_key` is either a one or two-item tuple (e.g. `(2,)` [only a ConcatDim] or `(2, 3)` [ConcatDim and MergeDim]). I To implement multiple ConcatDims, we would probably need to generalize `chunk_key` to an arbitrary length.

Another key method is `input_position`:
[...]

Issue #98, opened 2021-04-06
Allowing multiple dimensions in `sequence_dim`


chiaral:
Right now my understanding is that only one dimension can be used/created to concatenate files.

Example:
```
recipe = NetCDFtoZarrSequentialRecipe(
    input_urls=input_urls,
    sequence_dim="time",
)
```

Forecast data - however - they have multiple "concatenated" dimensions, or at least 2: Start time, ensemble members

As of now in [this issue](https://github.com/pangeo-forge/staged-recipes/issues/17) I am concatenating the files along Start time.  Something like the following:

```
recipe = NetCDFtoZarrSequentialRecipe(
    input_urls=input_urls,
    sequence_dim=["time", "ensemble"]
)
```

would be required, with some type of nested processing of the data.

To be honest, I don't think this is a super urgent issue. We can easily create a file for each ensemble member - and then the user can easily `xr.merge` or `xr.combine_by_coords` the various ensemble members.  It might actually be a good compromise to handle separately the [ensemble members that have different time resolution](https://github.com/pangeo-forge/staged-recipes/issues/17#issuecomment-765680086). We simply generate one file for each ensemble member. Still, I wanted to report this need.

PR #141, opened 2021-05-20
Add chunk grid logic

rabernat:
This is a first start on closing #140 and #98.

Let's create a new abstract class called ChunkGrid that understands about the mapping between chunk indexes and array slices and implement methods to find the overlap / intersection between different grids. Then we can refactor XarrayZarrRecipe to use this rather than its current impenetrable logic.

This is definitely WIP; just wanted to push something to get the conversation started.

### DIFF
--- a/pangeo_forge_recipes/chunk_grid.py
+++ b/pangeo_forge_recipes/chunk_grid.py
@@ -1,0 +1,67 @@
+"""
+Abstract representation of ND chunked arrays
+"""
+
+from dataclasses import dataclass
+from typing import Tuple
+
+import numpy as np
+
+# Most of this is probably already in Dask and Zarr!
+# However, it's useful to write up our own little model that does just what we need.
+
+ChunkIndex = Tuple[int, ...]
+
+
+@dataclass
+class ChunkGrid:
+    shape: Tuple[int, ...]
+
+    # we will always be dealing with named dims
+    dims: Tuple[str, ...]
+
+    # chunks are stored explicitly (don't assume uniform length)
+    chunks: Tuple[Tuple[int, ...], ...]
+
+    @property
+    def ndim(self):
+        return len(self.shape)
+
+    def __post_init__(self):
+        if len(self.dims) != self.ndim:  # pragma: no cover
+            raise ValueError("dims must have the same len as shape")
+        if len(self.chunks) != self.ndim:  # pragma: no cover
+            raise ValueError("chunks must have the same len as shape")
+
+        self._chunk_axes = [
+            ChunkAxis(dim_shape, dim_chunks)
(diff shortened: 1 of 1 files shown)


alxmrs:
Hey Ryan! I noticed that this approach seems a lot like XArray-Beams's `ChunkKey`s ([source]). I've used them on an internal parallel version of the XarrayToZarr recipe that supports using multiple `ConcatDim`s (it uses XArrayBeam under the hood, but supports the FilePattern interface). 

I'd love to talk more about the approach you're taking here, and how we can generalize this pattern. 


[source]: https://github.com/google/xarray-beam/blob/main/xarray_beam/_src/core.py#L39

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -191,67 +192,45 @@ def region_and_conflicts_for_chunk(
...
         ), "Doesn't make sense to have multiple inputs per chunk plus subsetting"
-        subset_factor = subset_inputs[concat_dim]
-        input_sequence_lens = tuple(
-            chain(*(calc_subsets(input_len, subset_factor) for input_len in input_sequence_lens))
-        )
-        subset_idx = [
-            dim_idx.index
-            for dim_idx in chunk_key
-            if dim_idx.operation == CombineOp.SUBSET and dim_idx.name == concat_dim
-        ][0]
```
rabernat, line 222: Note how we no longer need this opaque logic in this function. It now lives in ChunkGrid.

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -191,67 +192,45 @@ def region_and_conflicts_for_chunk(
...
-    region_slice = slice(start, stop)
-
-    this_chunk_conflicts = set()
-    for idx in range(start_idx, stop_idx):
-        conflict = all_chunk_conflicts[idx]
-        if conflict:
-            for conflict_index in conflict:
-                this_chunk_conflicts.add(conflict_index)
-
-    return {concat_dim: region_slice}, this_chunk_conflicts
```
rabernat, line 254: Same. All this code goes away.

rabernat (commented):
Note that the overall effect of this is to shift the reasoning about chunks, overlaps, subsets, etc. into an abstract, standalone class where the logic can be clearly implemented and thoroughly tested.

pangeo_forge_recipes/recipes/xarray_zarr.py
rabernat: (and the new `chunk_position` function, which translates a ChunkKey to a chunk index)

PR #150, opened 2021-05-28
Remove some contexts and simplify storage code

### LINKED ISSUE #134: 3.7-build Actions hang on Run Tests
On #132, all checks aside from 3.7-build passed within ~10 minutes. The 3.7-build checks hung on the Run Tests step until they eventually timed out. I just restarted them but they appear to be hanging again.

### LINKED ISSUE #144: Intermittent hanging when opening cached files without failure.
In recent attempts at running the [SMAP SSS Recipe](https://github.com/pangeo-forge/staged-recipes/pull/31) we have noted intermittent hanging when opening from cache via [_maybe_open_or_copy_to_local](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/0.3.3/pangeo_forge_recipes/recipes/xarray_zarr.py#L61).  These are dask worker log entries from an recently attempted run where workers were hung

```
INFO:pangeo_forge_recipes.recipes.xarray_zarr:Opening 'https://podaac-opendap.jpl.nasa.gov/opendap/allData/smap/L3/RSS/V4/monthly/SCI/2015//RSS_smap_SSS_L3_monthly_2015_08_FNL_v04.0.nc' from cache
```
```
INFO:pangeo_forge_recipes.recipes.xarray_zarr:Opening 'https://podaac-opendap.jpl.nasa.gov/opendap/allData/smap/L3/RSS/V4/monthly/SCI/2015//RSS_smap_SSS_L3_monthly_2015_11_FNL_v04.0.nc' from cache
```
Interestingly, though the files which hang for cache retrieval are different between tests but consistently 3 workers seem to hang.  Not sure if this indicates some type of concurrency issue with cache retrieval.

Besides blocking the recipe, this is problematic due to the asynchronous nature of these operations which can potentially leave workers in an idle healthy state and prevent cleanup from the scheduler.

@rabernat I had look over the storage refactoring [PR](https://github.com/pangeo-forge/pangeo-forge-recipes/commit/32e920185934e42baf32650c5b3d1e76ea078ab3#diff-e12c886cc124886c5cfa5313d760a36c39649af9da845077c663e6feab8487b5L388) but I'm unsure if this might potentially address this issue?  Do you have recommendations on any additional logging or telemetry you would like me to collect that might be helpful in diagnosing the root cause for this?

If you feel the storage refactoring PR might address this issue, can we craft a new release which includes this so it can be deployed in our `staged-recipes` CI workflows.

martindurant:
Fixes #144 
Fixes #134

### DIFF
--- a/ci/py3.8.yml
+++ b/ci/py3.8.yml
@@ -10,7 +10,7 @@ dependencies:
   - codecov
   - dask
   - distributed
-  # - fsspec>=2021.04.0
+  - fsspec>=2021.04.0
   - h5netcdf
   - h5py
   - hdf5
@@ -33,6 +33,7 @@ dependencies:
   - toolz
   - xarray>=0.18.0
   - zarr>=2.6.0
+  - pytest-repeat
+  - pip
   - pip:
     - pytest-timeout
-    - git+https://github.com/martindurant/filesystem_spec.git@log_and_cache
--- a/pangeo_forge_recipes/recipes/xarray_zarr.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr.py
@@ -353,7 +353,10 @@ class XarrayZarrRecipe(BaseRecipe):
         with file_opener(fname, cache=cache, copy_to_local=self.copy_input_to_local_file) as f:
             with dask.config.set(scheduler="single-threaded"):  # make sure we don't use a scheduler
                 logger.debug(f"about to call xr.open_dataset on {f}")
-                ds = xr.open_dataset(f, **self.xarray_open_kwargs)
+                kw = self.xarray_open_kwargs.copy()
+                if "engine" not in kw:
+                    kw['engine'] = "h5netcdf"
+                ds = xr.open_dataset(f, **kw)
                 logger.debug("successfully opened dataset")
                 ds = fix_scalar_attr_encoding(ds)
 
--- a/pangeo_forge_recipes/recipes/xarray_zarr.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr.py
@@ -355,7 +355,7 @@ class XarrayZarrRecipe(BaseRecipe):
                 logger.debug(f"about to call xr.open_dataset on {f}")
                 kw = self.xarray_open_kwargs.copy()
                 if "engine" not in kw:
-                    kw['engine'] = "h5netcdf"
+                    kw["engine"] = "h5netcdf"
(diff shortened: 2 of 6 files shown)


martindurant:
I explicitly tell the various instances of `xr.open_dataset` to use driver "h5netcdf", since sometimes it guesses netcdf4, which fails for some file-like sometimes.

rabernat:
Interesting and useful suggestion. Curious to see if it works.

If it works, we should think carefully about the intersection with `xarray_open_kwargs`. For some recipes, the user will specify an engine (e.g. `engine='cfgrib'`) in order to read special formats.

ci/py3.8.yml
```diff
@@ -10,7 +10,7 @@ dependencies:
   - codecov
   - dask
   - distributed
-  # - fsspec>=2021.04.0
+  - fsspec>=2021.04.0
```
rabernat, line 13: Didn't I already do this in #148?

martindurant: I must not have pulled/merged.

martindurant:
> For some recipes, the user will specify an engine (e.g. engine='cfgrib')

Those would presumably *require* copy-to-local.

rabernat:
Did this really just fix it?

rabernat:
> Those would presumably _require_ copy-to-local.

YES

rabernat:
I'll restart the tests just to get another datapoint.

martindurant:
> Did this really just fix it?

I'll wait for more data points.

martindurant:
Is that a green?

rabernat:
Btw, didn't I make an fsspec issue about this "double context manager" thing? I can't seem to find it anymore.

It's really an xarray bug IMO.

martindurant:
The  LocalFileOpener became a derivative of io.IOBase in 65880ef8dbeadfd655a5aefb5436f2390042d837 , so it shouldn't be necessary any more; but for local files that netcdf4 can normally cope with, it still fails. I guess there is an exact instance check somewhere, but I can't see where. Perhaps it's tied up with how xarray guesses file types by looking at magic bytes. I haven't checked whether netcdf4 itself can cope with LocalFileOpener or if it's only via xarray.

martindurant:
https://github.com/intake/filesystem_spec/pull/589

rabernat:
I'm gonna merge this, ok?

martindurant:
Yep

PR #155, opened 2021-06-03
Reset default copy_input_to_local_file=False for XarrayZarrRecipe

cisaacstern:
@rabernat set a default of `copy_input_to_local_file=True` in `XarrayZarrRecipe` at commit https://github.com/pangeo-forge/pangeo-forge-recipes/pull/146/commits/e92c5f2344329be7ff38e2ca5c244e488cebf4cf of https://github.com/pangeo-forge/pangeo-forge-recipes/pull/146. This was part of an extensive debugging journey which is (hopefully) complete now, and can therefore be reset.

Resetting this to `copy_input_to_local_file=False` came up as part of our conversation about the questions I outlined in https://github.com/pangeo-forge/pangeo-forge-recipes/issues/154. As it turns out, my (mis)perception of the cost of metadata caching resulted partially from misreading the logs: what I thought was a lot of logging for metadata caching was actually the logging for `copy_input_to_local_file`.

PR #157, opened 2021-06-04
Remove copy_input_to_local hang warning

cisaacstern:
@rabernat and @sharkinsspatial, following #155 `pangeo-forge:master` now raises the warning I've removed in this PR by default whenever importing recipes from an `XarrayZarrRecipe` instance for which `copy_input_to_local` has not been changed from the (newly reset) default of `False`

Submitting this as a draft because I'm not sure if this desired or not 🤷‍♂️


rabernat:
Definitely desirable. We should have done this in #155.

PR #158, opened 2021-06-10
Configurable lock timeouts

TomAugspurger:
This allows for setting a timeout on a recipe to control how long it waits to acquire a lock.

xref https://github.com/pangeo-forge/pangeo-forge-azure-bakery/issues/10, where I'm apparently seeing some locking issues.

### DIFF
--- a/tests/test_recipes.py
+++ b/tests/test_recipes.py
@@ -1,10 +1,13 @@
 from contextlib import nullcontext as does_not_raise
+from unittest.mock import patch
 
 import pytest
 import xarray as xr
 
 # need to import this way (rather than use pytest.lazy_fixture) to make it work with dask
 from pytest_lazyfixture import lazy_fixture
+
+from tests.conftest import execute_recipe
 
 all_recipes = [
     lazy_fixture("netCDFtoZarr_sequential_recipe"),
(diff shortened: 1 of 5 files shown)


rabernat:
I totally understand why something like this is needed.

My only question is whether it should live in the recipe itself. This seems like a property of the execution environment, not the recipe. How would a recipe creator--ostensibly an expert in the data but knowing little about dask, etc.--be able to choose this parameter in an informed way?

Could we consider using an environment variable or other bakery-level configuration to set this instead of a new recipe attribute?

rabernat:
On the other hand, you could say the same thing about storage, so perhaps this critique is inconsistent.

TomAugspurger:
That was my thought. It's usage at https://github.com/pangeo-forge/pangeo-forge-azure-bakery/commit/c1d6f22f33d35a555af2913ee14516f36954b8ed#diff-b4a38f17564695fb839c037798280e157bc1bc5f19d4ded9ec5ec7189552eb17R54 is pretty similar to `recipe.target`.

A global config would work as well, but would need to be synchronized to all the workers prior to executing. This setup keeps the usage of the value closer to where it's assigned. My real preference would be for it to be an argument to `store_chunk`, but that might be more difficult to plumb through.

tests/test_recipes.py
```diff
@@ -154,3 +155,18 @@ def test_chunks(
     ds_actual.load()
     print(ds_actual)
     xr.testing.assert_identical(ds_actual, ds_expected)
+
+
+def test_lock_timeout(netCDFtoZarr_sequential_recipe, execute_recipe):
+    RecipeClass, file_pattern, kwargs, ds_expected, target = netCDFtoZarr_sequential_recipe
+    recipe = RecipeClass(file_pattern=file_pattern, lock_timeout=1, **kwargs)
+
+    with patch("pangeo_forge_recipes.recipes.xarray_zarr.lock_for_conflicts") as p:
```
rabernat, line 164: Could you just add a quick comment stating what this does? I have never used mocking before.

rabernat: nm you did right below

rabernat:
LGTM! Let's let @cisaacstern take a look and then we can merge.

cisaacstern:
👀 ✅

Issue #116, opened 2021-04-30
Improve serializability of Recipe and related classes


rabernat:
After the refactor in #101, our dask graphs have gotten a lot bigger.

I did the following, based on the NOAA OISST recipe from the docs

```python
import pandas as pd
import s3fs
import gcsfs
import json

from pangeo_forge.patterns import pattern_from_file_sequence
from pangeo_forge.recipes import XarrayZarrRecipe
from pangeo_forge.storage import FSSpecTarget, CacheFSSpecTarget

# set up recipe

input_url_pattern = (
    "https://www.ncei.noaa.gov/data/sea-surface-temperature-optimum-interpolation"
    "/v2.1/access/avhrr/{yyyymm}/oisst-avhrr-v02r01.{yyyymmdd}.nc"
)

dates = pd.date_range("1981-09-01", "2021-01-05", freq="D")
input_urls = [
    input_url_pattern.format(
        yyyymm=day.strftime("%Y%m"), yyyymmdd=day.strftime("%Y%m%d")
    )
    for day in dates
]

pattern = pattern_from_file_sequence(input_urls, "time", nitems_per_file=1)
recipe = XarrayZarrRecipe(pattern, inputs_per_chunk=5)

# set up storage

fmt = 'zarr'
recipe_key = 'oisst_avhrr_v02r01'
recipe_name = f'{feedstock_name}/{recipe_key}'

feedstock_name = 'noaa_oisst'

cache_base = f'gs://pangeo-forge-us-central1/pangeo-forge-cache'
metadata_base = f'gs://pangeo-forge-us-central1/pangeo-forge-metadata'

with open('pangeo-181919-cc01a4fbb1ef.json') as fp:
    token = json.load(fp)
fs_gcs = gcsfs.GCSFileSystem(token=token, cache_timeout=-1, cache_type="none")

endpoint_url = 'https://ncsa.osn.xsede.org'
fs_osn = s3fs.S3FileSystem(
    key=osn_key,
    secret=osn_secret,
    client_kwargs={'endpoint_url': endpoint_url},
    default_cache_type='none',
    default_fill_cache=False,
    listings_expiry_time=0.01,
)

target_base = f's3://Pangeo/pangeo-forge'
recipe.input_cache = CacheFSSpecTarget(fs_gcs, f"{cache_base}/{recipe_name}")
recipe.metadata_cache = FSSpecTarget(fs_gcs, f"{metadata_base}/{recipe_name}")
[...]

PR #153, opened 2021-06-02
New implementation of Dask & Prefect executors

rabernat:
possibly fixes #116
maybe eventually fixes #151

Issue #151, opened 2021-05-30
Dask distributed client memory overruns when with large time dimensions.


sharkinsspatial:
I performed some successful initial tests processing the [noaa-oisst](https://github.com/pangeo-forge/staged-recipes/pull/36) recipe using a smaller time dimension. When extending the recipe to use the full time range `pd.date_range("1981-09-01", "2021-01-05", freq="D")` I am seeing memory overruns on the container where the Prefect Flow is being executed (the container is killed for exceeding its 30GB hard limit).  

To obtain some more details I attempted to run the Flow in a local container directly using `flow.run` rather than the Prefect Agent's subprocess.  I added some `tracemalloc` instrumentation and found rapidly growing memory utilization from `msgpack`.
```
INFO:root:/srv/conda/envs/notebook/lib/python3.8/site-packages/msgpack/__init__.py:0: size=66.9 MiB, count=6, average=11.2 MiB
INFO:root:/srv/conda/envs/notebook/lib/python3.8/site-packages/msgpack/__init__.py:0: size=1156 MiB, count=120, average=9868 KiB
INFO:root:/srv/conda/envs/notebook/lib/python3.8/site-packages/msgpack/__init__.py:0: size=2408 MiB, count=251, average=9826 KiB
INFO:root:/srv/conda/envs/notebook/lib/python3.8/site-packages/msgpack/__init__.py:0: size=3603 MiB, count=377, average=9787 KiB
```
Before the process was killed after exceeding the container's 5GB hard limit.

A traceback of the largest block prior to the container being killed points to message serialization in distributed
```
INFO:root:34 memory blocks: 1448426.9 KiB
INFO:root:  File "/srv/conda/envs/notebook/lib/python3.8/threading.py", line 890
INFO:root:    self._bootstrap_inner()
INFO:root:  File "/srv/conda/envs/notebook/lib/python3.8/threading.py", line 932
INFO:root:    self.run()
INFO:root:  File "/srv/conda/envs/notebook/lib/python3.8/threading.py", line 870
INFO:root:    self._target(*self._args, **self._kwargs)
INFO:root:  File "/srv/conda/envs/notebook/lib/python3.8/concurrent/futures/thread.py", line 80
INFO:root:    work_item.run()
INFO:root:  File "/srv/conda/envs/notebook/lib/python3.8/concurrent/futures/thread.py", line 57
INFO:root:    result = self.fn(*self.args, **self.kwargs)
[...]

PR #160, opened 2021-06-16
Refactor XarrayZarrRecipe to be serialization-friendly

TomAugspurger:
This is (another) refactor to the XarrayZarrRecipe to resolve some of the memory issues we've seen when executing large jobs (https://github.com/pangeo-forge/pangeo-forge-azure-bakery/issues/10, https://github.com/pangeo-forge/pangeo-forge-recipes/issues/151)

It builds on https://github.com/pangeo-forge/pangeo-forge-recipes/pull/153, which adds `to_dask` and `to_prefect` methods (probably should do that at the base recipe level). 

It looks like a large diff, but it's primarily just moving code from methods on `XarrayZarrRecipe` to top-level functions, and forwarding arguments appropriately. This eliminates `self` from the functions sent to workers. More explanation at https://github.com/pangeo-forge/pangeo-forge-recipes/pull/160/files#diff-e12c886cc124886c5cfa5313d760a36c39649af9da845077c663e6feab8487b5R685-R693. 

The main outstanding task right now is ensuring that the metadata_cache is being handled properly. I need to better understand where that object is supposed to live (on the client or workers?) and who is supposed to be able to write to it (do workers write to it? Is it expected to be global so that writes from a worker process are seen by the client?)

- [x] Move the new compilation methods to `BaseRecipe`. This would also help clarify the API questions (`finalize_target` vs `_finalize_target`, which are required to be implemented where, etc.)
- [x] Remove the `to_pipelines` method
- [x] Implement `to_function` method, which is equivalent to the existing PythonExecutor
- [x] Update executor docs (started in 9c39bf77c332474260d0109ae083f4ffec694df7)
- [x] Rerun all notebooks in `docs/tutorials` to verify no end-user API changes are needed.
- [x] Verify metadata caching isn't broken

Closes https://github.com/pangeo-forge/pangeo-forge-recipes/issues/116.


rabernat:
Thanks for working on this Tom!

> It builds on #153,

Does it directly build on #153? Or is it a new implementation of similar ideas.

This is a pretty big refactoring, so I'd like to understand the motivation more clearly, specifically, why this is needed on top of #153.

Is the basic issue that we cannot use _any_ methods (that contain `self` as an argument) whatsoever without embedding the large Recipe objects in every task? So therefore you need to essentially rewrite everything in functional form?

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -46,6 +48,480 @@ def _chunk_metadata_fname(chunk_key) -> str:
...
+    cache_inputs: bool,
+    input_cache: Optional[CacheFSSpecTarget],
+    file_pattern: FilePattern,
+    fsspec_open_kwargs: dict,
+    cache_metadata: bool,
+    copy_input_to_local_file: bool,
+    xarray_open_kwargs: dict,
+    delete_input_encoding: bool,
+    process_input: Optional[Callable[[xr.Dataset, str], xr.Dataset]],
+    metadata_cache: Optional[MetadataTarget],
```
rabernat, line 124: I feel like these long argument blocks may hurt maintainability. There is so much room for programmer error when passing long lists of arguments through the stack. If we go this route, perhaps we want to enclose some of the arguments in a `Config` object?

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -206,303 +682,178 @@ def _set_target_chunks(self):
...
-        selector = [slice(0, 1)] * len(self.file_pattern.dims)
-        selector[concat_dim_axis] = slice(None)  # this should broadcast correctly agains all_lens
-        sequence_lens = all_lens[tuple(selector)]
-        if not (all_lens == sequence_lens).all():
-            raise ValueError(f"Inconsistent sequence lengths found: f{all_lens}")
-        return sequence_lens.squeeze().tolist()
+        cache_input_task = task(self._cache_input, name="cache_input")
+        prepare_target_task = task(self._prepare_target, name="prepare_target")
+        store_chunk_task = task(self._store_chunk, name="store_chunk")
+        finalize_target_task = task(self._finalize_target, name="finalize_target")
```
rabernat, line 849: Here you not using `_finalize_target` rather than `finalize_target`, etc.

Is there any point in maintaining the `finalize_target` methods if we are going to remove the existing `to_pipelines` method in `BaseRecipe`?

TomAugspurger: I think having the top-level property `.finalize_target` is still really nice for users who are developing / debugging recipes. We want to keep that, while still allowing Dask and Prefect to access the (partially applied) function itself, so they can wrap it in a Task.

pangeo_forge_recipes/recipes/xarray_zarr.py
TomAugspurger: Agreed! Though I *think* the biggest risk of programmer error is right now / in future major refactors. So it comes down to simplicity of the code (not having to look up what is this "FooConfig" thing?) vs. typing all these out. I'm 51% in favor of writing things out explicitly like this, but am happy to go with a Config / Options style object.

I will say that mypy has been helpful here. It caught a couple issues before running the tests.

rabernat:
Ok, so I think I am on board with this as hopefully the right solution to the serializability problems.

Here is a partial checklist of some things that IMO would need to be done before we merge this:

- [ ] Move the new compilation methods to `BaseRecipe`. This would also help clarify the API questions (`finalize_target` vs `_finalize_target`, which are required to be implemented where, etc.)
- [ ] Remove the `to_pipelines` method
- [ ] Implement `to_function` method, which is equivalent to the existing PythonExecutor
- [ ] Update executor docs (started in 9c39bf77c332474260d0109ae083f4ffec694df7)
- [ ] Rerun all notebooks in `docs/tutorials` to verify no end-user API changes are needed.

Happy to help with some of these via pushing directly to this brach.

Thanks again Tom for taking the time to sort out the core issues.

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -206,303 +682,178 @@ def _set_target_chunks(self):
...
-            logger.debug(f"{ds}")
+        return self._prepare_target()
+
+    @property
+    def _cache_input(self):
+        return functools.partial(
+            cache_input,
+            cache_inputs=self.cache_inputs,
+            input_cache=self.input_cache,
+            file_pattern=self.file_pattern,
```
rabernat, line 730: I am having a hard time understanding why these partial functions are "small" in a serialization sense, since each function in the graph essentially contains a `functools.partial`-wrapped version of all of the same attributes that are part of the recipe class? In particular, `file_pattern` and `target`s are all fairly complex / large objects themselves, which are now being curried into a single-argument function. How is that any better than having a method attached to a class?

TomAugspurger: Just noting that my initial attempt to answer this question have failed. My hunch is that it's just easier to serialize these functions than it is to serialize the dataclass, but I don't I'd like to have a stronger justification than that. Pointing to the workflow in https://github.com/pangeo-forge/pangeo-forge-azure-bakery/issues/10 is some evidence, but it's pretty indirect.

I'll keep trying to come up with a clear answer.

rabernat: To be clear, having a forensic understanding of this is interesting academically but far less important than actually shipping code that works. 😁  So given limited time, I would focus on the checklist above (rather than digging deeper). 

Let me know if you want help on any aspects here. I have stopped working on #153 in the meantime.

TomAugspurger: I'll take care of the first few items around `to_pipelines()` and will look into the outstanding metadata caching.

TomAugspurger: I updated the docs and notebooks. I had some trouble running the notebook, but I think it was an issue with my local internet connection.

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -206,303 +682,178 @@ def _set_target_chunks(self):
...
-        region_slice = slice(start, stop)
-        return {self._concat_dim: region_slice}, this_chunk_conflicts
+        return Delayed(key, dsk)
 
-    def iter_chunks(self):
-        for k in self._chunks_inputs:
-            yield k
+    def to_prefect(self):
+        """Compile the recipe to a Prefect.Flow object."""
+        from prefect import Flow, task, unmapped
```
rabernat, line 844: Isn't this now implemented twice? You also have this in BaseRecipe, no?

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -206,303 +682,178 @@ def _set_target_chunks(self):
...
+        # Cache Input --------------------------------------------------------
+        dsk = {}
+        token = dask.base.tokenize(self)
 
-    def open_target(self):
-        target_mapper = self.target.get_mapper()
-        return xr.open_zarr(target_mapper)
+        for i, input_key in enumerate(self.iter_inputs()):
+            dsk[(f"cache_input-{token}", i)] = (self._cache_input, input_key)
+        dsk[f"checkpoint_0-{token}"] = (lambda *args: None, list(dsk))
```
rabernat, line 813: Same, this is also implemented in BaseRecipe, no?

rabernat (commented):
Is it intentional to override the `to_dask` and `to_prefect` methods from `BaseRecipe` in `XarrayZarrRecipe`?

TomAugspurger:
> Is it intentional to override the to_dask and to_prefect methods from BaseRecipe in XarrayZarrRecipe?

Fixed in 6d190ff5cc87abe296b82498f474688de4f898f9, to remove the `to_*` methods from XarrayZarrRecipe and move them to the base.

That also removes the `@closure` and underscore-versions of the properties (`_prepare_target`, etc). since they aren't needed anymore.

TomAugspurger:
Well, I feel a bit silly. The difficulty in answering https://github.com/pangeo-forge/pangeo-forge-recipes/pull/160/files/e70d52662875f8835cc180cb289fc4e6d4445e4a#diff-e12c886cc124886c5cfa5313d760a36c39649af9da845077c663e6feab8487b5 spurred some more investigation into what was actually taking a lot of size to serialize.

It turns out  that most of the size was in the FilePattern class. Maybe this isn't too surprising, but it did surprise me that the size was actually in the function `filepatterns_from_sequence` at https://github.com/pangeo-forge/pangeo-forge-recipes/blob/562e76ccb77831540d5424e948a6fe4ec440263a/pangeo_forge_recipes/patterns.py#L151-L160, rather than the class itself. And then it was obvious, in https://github.com/pangeo-forge/pangeo-forge-azure-bakery/blob/8c7c40183d3a0eedbaab7dac483287b8abf54c46/flow_test/oisst_recipe.py#L100-L106 we create a list with 14,000 (largeish) strings. I'm still not sure why, but apparently multiple instances of that list were being created when the function was deserialized (maybe since it's referenced via a closure, rather than a top-level function?).

The alternative is to create the `FIlePattern` class "manually": https://github.com/pangeo-forge/pangeo-forge-azure-bakery/blob/c70246d209a529ba921b73ed779bfbb57e9178a6/flow_test/oisst_recipe.py#L96-L110. This just requires a function, and a `range` object. Much smaller.

For some reason in my debugging, I had turned `cache_input` off. Which led me to conclude that this branch was helping. I'll now test the alternative FilePattern against pangeo-forge-recipes master, to see if this is even necessary.

rabernat:
> It turns out that most of the size was in the FilePattern class. Maybe this isn't too surprising, but it did surprise me that the size was actually in the function `filepatterns_from_sequence` at

:+1: This is definitely expected from my POV. Perhaps we should deprecate that function.

TomAugspurger:
> Perhaps we should deprecate that function.

I think that's worthwhile. It is convenient, but I worry that this will bite us again as we scale users's recipes to large jobs. And I hope that manually constructing the FilePattern isn't too much more difficult.

---

FWIW, when I ran the flow using the fixed FilePattern construction on pangeo-forge-recipes master, my scheduler was OOM Killed. It does seem like this branch is doing something.

rabernat:
The big question I have is whether this PR is still significantly better than #153 in terms of serialization. (Assuming the more efficient FilePattern approach is used in both cases.)

TomAugspurger:
> The big question I have is whether this PR is still significantly better than #153 in terms of serialization.

It seems like it. I ran a flow that just sleeps for the `cache_input` step with pangeo-forge-recipes master, #153, and this PR. #153 still had memory issues on the scheduler, while this PR didn't.

https://github.com/pangeo-forge/pangeo-forge-azure-bakery/issues/10#issuecomment-867165719 has the results.

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -46,6 +47,480 @@ def _chunk_metadata_fname(chunk_key) -> str:
...
+    )
+    n_sequence = sum(input_sequence_lens)
+    logger.info(f"Expanding target concat dim '{concat_dim}' to size {n_sequence}")
+    expand_target_dim(target, concat_dim, n_sequence)
+
+    # TODO(Tom): Handle state on the object
+    # if cache_metadata:
+    #     # if nitems_per_input is not constant, we need to cache this info
+    #     recipe_meta = {"input_sequence_lens": input_sequence_lens}
+    #     return recipe_meta
```
rabernat, line 424: What's going on here? I don't follow your comment.

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -46,6 +47,480 @@ def _chunk_metadata_fname(chunk_key) -> str:
...
+        input_cache=input_cache,
+        cache_inputs=cache_inputs,
+        copy_input_to_local_file=copy_input_to_local_file,
+        xarray_open_kwargs=xarray_open_kwargs,
+        delete_input_encoding=delete_input_encoding,
+        process_input=process_input,
+    ) as ds:
+        input_metadata = ds.to_dict(data=False)
+        # TODO(METADATA): set
+        metadata_cache[_input_metadata_fname(input_key)] = input_metadata
```
rabernat, line 109: Is there some reason metadata requires different treatment here?

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -46,6 +47,480 @@ def _chunk_metadata_fname(chunk_key) -> str:
...
+    # also return the conflicts with other chunks
+
+    input_keys = chunks_inputs[chunk_key]
+
+    if nitems_per_input:
+        input_sequence_lens = (nitems_per_input,) * file_pattern.dims[concat_dim]  # type: ignore
+    # TODO(Tom): Handle metadata caching here
+    # else:
+    # global_metadata = metadata_cache[_GLOBAL_METADATA_KEY]
+    # input_sequence_lens = global_metadata["input_sequence_lens"]
```
rabernat, line 166: Just flagging that metadata stuff has been commented out.

rabernat (commented):
I went through this again and noticed that all of the global metadata references have been commented out. Can you explain why global metadata is more difficult with this new functional syntax?

Is there something I can do to help here?

TomAugspurger:
Fixing the metadata stuff is my (hopefully) last TODO. I'll dig into it today, but maybe you can answer this easily: Where do the writes and reads to the global metadata store happen, client process or the worker processes?

rabernat:
Worker processes.

TomAugspurger:
Sounds good. So it's assumed that the metadata cache is globally read / writeable, like a blob storage file system. In that case, I think my last commit fixes things, but the tests were passing without it. It's probably worth adding some kind of check to the tests to verify that.


---

I'm running through the notebooks now, and at least some are failing with `TypeError: 'FSSpecTarget' object does not support item assignment.` I'm guessing they're failing on master too. The targets should be a `MetadataTarget`, rather than an `FsspecTarget`. Maybe related to https://github.com/pangeo-forge/pangeo-forge-recipes/pull/133.

I'll update the notebooks to use MetadataTarget.

TomAugspurger:
The notebooks caught an issue: I had removed `XarrayZarrRecipe.open_input` and `XarrayZarrRecipe.open_chunk`. These are useful helpers for debugging a recipe so I restored them and added a basic test to ensure they're present.

The notebooks are now updated to use a `MetadataCache` and are all passing. I think this should be good to go.

rabernat (approved):
This is fantastic. Thanks so much Tom for all your hard work on this!

Could you just update the [release notes](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/master/docs/development/release_notes.md)? I think at this point our next release is 0.4, so probably need to add a new section.

TomAugspurger:
Done in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/160/commits/f57e36adc661f55f7b1e9703857228b79674a651.

rabernat:
@TomAugspurger, can you think of a test we could add to guard against regressions related to serialization? Relevant for big PRs like #166 which touch a lot of different pieces...

TomAugspurger:
Nothing really comes to mind :/ The typical way to check is to serialize the object and then check the size of the bytestring. But IIRC we found that these objects weren't all that large.

PR #165, opened 2021-07-13
File subsetting

### LINKED ISSUE #93: Handling of large files
Our base use cases for Pangeo Forge (e.g. https://github.com/pangeo-forge/staged-recipes/pull/20) are _many small netCDF files ➡️ one big Zarr_. But what if we have just one or a few big netCDF files (e.g. https://github.com/pangeo-forge/staged-recipes/issues/2)? As an extreme example, we might have a single file that is 100 GB or larger. We need the recipe to be able to subset the file lazily to avoid loading the whole things into memory. There are at least two ways to do this.

## Subset by variable

This happens now in a limited way. In `store_chunk`, we iterate through each variable and store each individually. We are not using dask or async at this level of the code, so these writes are guaranteed to happen in serial for each chunk.

https://github.com/pangeo-forge/pangeo-forge/blob/e5f7e7cc463ea31513088e14047b962a90cb69f9/pangeo_forge/recipe.py#L357-L359

There is an opportunity for parallelism here, especially if each chunk dataset has many different variables. We could add an option to split chunks by variable. In this case, the user would probably have to provide an explicit list of variables they want to store when the recipe is being created. Implementing this will unfortunately require a deepish refactor because it will affects the `_chunks_inputs` data structure.

Also, subsetting by variable can't help us if a single file contains very large arrays. Note that the the whole array from each chunk is read into memory:

https://github.com/pangeo-forge/pangeo-forge/blob/e5f7e7cc463ea31513088e14047b962a90cb69f9/pangeo_forge/recipe.py#L377-L382

To avoid this, we need...

## Chunked writing of individual variables

Imagine that an individual file contains a single variable of 10 GB. Unless our workers have significantly more than 10 GB of RAM, the execution will fail. We can overcome this by using _chunked writing of individual variables_, i.e. by writing the 10 GB array in 500 MB pieces.

The easiest way to do this is probably to leverage Xarray's lazily-index-array wrappers (that we get for free when we open the data with Xarray) together with Dask. In this case, Dask would not be providing parallelism per-se but rather a quick way to implement blocked writes. For example, we could do something like
```python
chunk_size = '500MB'  # how to choose?
data = ds_chunk[varname].chunk(sequence_dim=chunk_size).data
with dask.config.set(scheduler="single-threaded"): 
    dask.array.store(data, zarr_array, lock=false, regions=zarr_region)
 ```

Again, we don't want any parallelism _within_ this write operation. So far I am always assuming that all of the key methods in the recipe will execute in a single thread. The parallelism always occurs at a higher level, by mapping out many such tasks simultaneously.

## Summary

- [ ] Implement `split_chunks_by_variable=['foo', 'bar']` option in `NetCDFtoZarrRecipe` classes.
- [ ] #164

### PR DESCRIPTION
**Pull request by rabernat**
Closes #93 by allowing us to have many virtual "inputs" per file. This is one possible implementation, and I'm looking for feedback on whether this approach makes sense.

### New Data Structures: `OpenSpec`, `SubsetSpec`

Before this PR, FilePatterns simply mapped input keys to strings, understood to be filenames.

With this PR, we introduce a new data structure called `OpenSpec`

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/e47d49ffbc1d345be1d90846b8484ac4b6426aa4/pangeo_forge_recipes/patterns.py#L26-L32

The `OpenSpec` contains the filename and also potentially many `SubsetSpecs`

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/e47d49ffbc1d345be1d90846b8484ac4b6426aa4/pangeo_forge_recipes/patterns.py#L11-L18

Each `SubsetSpec` tells us how to decimate a file into an arbitrary number of "segments" or pieces (open to changing this terminology) along a specified dimension. Now, when iterating through a FilePattern, we get not just the filename, but also, optionally, these SubsetSpecs.

<details>
  <summary>Example</summary>

```python
from pangeo_forge_recipes import patterns

def format_function(time):
    return f'file_{time}.nc'

concat_dim = patterns.ConcatDim(name="time", keys=[10, 20, 30])
subset_dim = patterns.SubsetDim(dim="time", subset_factor=2)
fp = patterns.FilePattern(format_function, concat_dim, subset_dim)

for key, value in fp.items():
    print(key, value)
```

```
(0, 0) OpenSpec(fname='file_10.nc', subsets=[SubsetSpec(dim='time', this_segment=0, total_segments=2)])
(0, 1) OpenSpec(fname='file_10.nc', subsets=[SubsetSpec(dim='time', this_segment=1, total_segments=2)])
(1, 0) OpenSpec(fname='file_20.nc', subsets=[SubsetSpec(dim='time', this_segment=0, total_segments=2)])
(1, 1) OpenSpec(fname='file_20.nc', subsets=[SubsetSpec(dim='time', this_segment=1, total_segments=2)])
(2, 0) OpenSpec(fname='file_30.nc', subsets=[SubsetSpec(dim='time', this_segment=0, total_segments=2)])
(2, 1) OpenSpec(fname='file_30.nc', subsets=[SubsetSpec(dim='time', this_segment=1, total_segments=2)])
```

</details>

This effectively allows us to break a big file into many small pieces, lazily slicing it to a range determined by the SubsetSpec, and treating these as distinct individual inputs. Here is how the `XarrayZarrRecipe.open_input` handles these SubsetSpecs:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/e47d49ffbc1d345be1d90846b8484ac4b6426aa4/pangeo_forge_recipes/recipes/xarray_zarr.py#L227-L230

### The Problem: Caching given One-to-Many relationship between actual Files and Inputs

Before, each "input" was an actual physical file. So caching inputs could be performed like

```python
for input in recipe.iter_inputs():
    recipe.cache_input(input)
```

Caching of subsetted files does not make sense. We want to either cache a whole file or not at all. If we run this code now, and simply discard the subset information, we will end up caching each file multiple times. If we just cache on the first iteration of each subset, the cache_metadata step will break.

### Alternative: implement Subsetting at the Chunk Level

Writing this up made me realize I make have made a bad choice in how I implemented this. We should probably be implementing subsetting at the chunk stage, not the input stage. 🤦 

I'm going to make a second try in a new PR.

### DIFF
--- a/pangeo_forge_recipes/patterns.py
+++ b/pangeo_forge_recipes/patterns.py
@@ -9,6 +9,30 @@ from typing import Any, Callable, Dict, Iterator, List, Optional, Sequence, Tupl
 
 
 @dataclass
+class SubsetSpec:
+    """A data structure that explains how to subset a file in one dimension.
+    """
+
+    dim_name: str
+    this_segment: int
+    total_segments: int
+
+    def __post_init__(self):
+        assert self.total_segments > 0
+        assert self.this_segment >= 0
+        assert self.this_segment < self.total_segments
+
+
+@dataclass
+class OpenSpec:
+    """A data structure that explains how to open a file.
+    """
+
+    fname: str
+    subsets: List[SubsetSpec] = field(default_factory=list)
+
+
+@dataclass
 class ConcatDim:
     """Represents a concatenation operation across a dimension of a FilePattern.
 
@@ -68,21 +92,16 @@ class SubsetDim:
         return list(range(self.subset_factor))
 
 
-# the type which should be returned by format_function
-OpenSpec = Tuple[str, dict]
-
-
 class FilePattern:
     """Represents an n-dimensional matrix of individual files to be combined
     through a combination of merge and concat operations. Each operation generates
     a new dimension to the matrix.
 
     :param format_function: A function that takes one argument for each
-      combine_op and returns an open_spec (tuple of fname and open_kwargs)
+      non-subset combine_op and returns a filename.
       Each argument name should correspond to a ``name`` in the ``combine_dims``
       list.
-    :param combine_dims: A sequence of either concat or merge dimensions. The outer
-      product of the keys is used to generate the full list of file paths.
+    :param combine_dims: A sequence of concat, merge, or subset dimensions.
     """
 
     def __init__(self, format_function: Callable, *combine_dims: CombineDim):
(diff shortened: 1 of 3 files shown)


rabernat:
Closed in favor of #166.

PR #166, opened 2021-07-15
Subset chunks

### LINKED ISSUE #93: Handling of large files

rabernat:
Closes #93 by allowing chunks to be a subset of inputs.

Lots of work left to do here.

### DIFF
--- a/pangeo_forge_recipes/recipes/xarray_zarr.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr.py
@@ -557,6 +557,10 @@ class XarrayZarrRecipe(BaseRecipe):
     :param process_chunk: Function to call on each concatenated chunk, with signature
       `(ds: xr.Dataset) -> ds: xr.Dataset`.
     :param lock_timeout: The default timeout for acquiring a chunk lock.
+    :param subset_inputs: If set, break each input file up into multiple chunks
+      along dimension according to the specified mapping. For example,
+      ``{'time': 5}`` would split each input file into 5 chunks along the
+      time dimension. Multiple dimensions are allowed.
     """
 
     file_pattern: FilePattern
@@ -575,6 +579,7 @@ class XarrayZarrRecipe(BaseRecipe):
     process_input: Optional[Callable[[xr.Dataset, str], xr.Dataset]] = None
     process_chunk: Optional[Callable[[xr.Dataset], xr.Dataset]] = None
     lock_timeout: Optional[int] = None
+    subset_inputs: Dict[str, int] = field(default_factory=dict)
 
     # internal attributes not meant to be seen or accessed by user
     _concat_dim: Optional[str] = None
--- a/pangeo_forge_recipes/recipes/xarray_zarr.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr.py
@@ -7,15 +7,16 @@ import logging
 import warnings
 from contextlib import ExitStack, contextmanager
 from dataclasses import dataclass, field, replace
-from itertools import product
-from typing import Callable, Dict, Hashable, List, Optional, Tuple
+from itertools import chain, product
+from math import ceil
+from typing import Callable, Dict, Hashable, Iterator, List, Optional, Sequence, Tuple
 
 import dask
 import numpy as np
 import xarray as xr
 import zarr
 
-from ..patterns import FilePattern, prune_pattern
+from ..patterns import CombineOp, DimIndex, FilePattern, FilePatternIndex, prune_pattern
 from ..storage import AbstractTarget, CacheFSSpecTarget, MetadataTarget, file_opener
 from ..utils import (
     chunk_bounds_and_conflicts,
@@ -31,17 +32,62 @@ _GLOBAL_METADATA_KEY = "pangeo-forge-recipe-metadata.json"
 
 logger = logging.getLogger(__name__)
 
+InputKey = FilePatternIndex
 
-def _encode_key(key) -> str:
-    return "-".join([str(k) for k in key])
+# InputKey is a tuple that indexes the recipe's FilePattern.
+# e.g. file_pattern[input_key] returns a file name
+# For now, it has at most two elements, e.g. (3, 4) or may have just one e.g. (3, ).
+# One of these elements must be a ConcatDim and the other might be a MergeDim.
+
+ChunkKey = Tuple[DimIndex, ...]
+# ChunkKey is a tuple that indexes the chunks of the dataset.
+# The first one / two elements map to the first one / two elements the recipe's InputKey.
+# The MergeDim indexes have the same meaning.
+# For the ConcatDim, if nitems_per_input > 1, there is a one-to-many relationship
+# btw the chunk indexes and input indexes for that dimension.
+# If there is no subsetting, len(chunk_key) == len(input_key).
+# With subsetting, ChunkKey gets an extra index for each subset_dim, such tgat
+# len(chunk_key) == (len(input_key) + len(subset_inputs))
+
+SubsetSpec = Dict[str, int]
+# SubsetSpec is a dictionary mapping dimension names to the number of subsets along that dimension
+# (e.g. {'time': 5, 'depth': 2})
+
+# now let's encode that twisted logic in functions! 😈
 
 
 def _input_metadata_fname(input_key):
-    return "input-meta-" + _encode_key(input_key) + ".json"
+    key_str = "-".join([f"{k.name}_{k.index}" for k in input_key])
+    return "input-meta-" + key_str + ".json"
 
 
-ChunkKey = Tuple[int]
-InputKey = Tuple[int]
+def inputs_for_chunk(
+    chunk_key: ChunkKey, inputs_per_chunk: int, ninputs: int
+) -> Sequence[InputKey]:
+    merge_dim = [dim_idx for dim_idx in chunk_key if dim_idx.operation == CombineOp.MERGE]
+    concat_dim = [dim_idx for dim_idx in chunk_key if dim_idx.operation == CombineOp.CONCAT]
+    subset_dims = [dim_idx for dim_idx in chunk_key if dim_idx.operation == CombineOp.SUBSET]
+    assert len(merge_dim) <= 1
+    if len(merge_dim) == 1:
+        merge_dim = merge_dim[0]
+    else:
+        merge_dim = None
+    assert len(concat_dim) == 1
+    concat_dim = concat_dim[0]
+
+    input_keys = []
+
+    for n in range(inputs_per_chunk):
+        input_index = (inputs_per_chunk * concat_dim.index) + n
+        if input_index >= ninputs:
+            break
+        input_concat_dim = DimIndex(concat_dim.name, input_index, ninputs, CombineOp.CONCAT)
+        input_key = [input_concat_dim]
+        if merge_dim is not None:
+            input_key.append(merge_dim)
+        input_keys.append(tuple(input_key))
+
+    return input_keys
 
 
 def expand_target_dim(target: CacheFSSpecTarget, concat_dim: Optional[str], dimsize: int) -> None:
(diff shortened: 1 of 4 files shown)


pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -218,10 +277,23 @@ def open_input(
...
+def subset_dataset(ds: xr.Dataset, subset_spec: DimIndex) -> xr.Dataset:
+    assert subset_spec.operation == CombineOp.SUBSET
+    dim = subset_spec.name
+    dim_len = ds.dims[dim]
+    subset_lens = calc_subsets(dim_len, subset_spec.sequence_len)
+    start = sum(subset_lens[: subset_spec.index])
+    stop = sum(subset_lens[: (subset_spec.index + 1)])
+    subset_slice = slice(start, stop)
+    indexer = {dim: subset_slice}
+    logger.debug(f"Subseting dataset with indexer {indexer}")
```
cisaacstern, line 289:
```suggestion
    logger.debug(f"Subsetting dataset with indexer {indexer}")
```

cisaacstern (commented):
Flagged one logger typo in-line. Will follow-up on the conversation thread with results from my tests against FESOM and eNATL60 datasets.

cisaacstern:
This is really cool, Ryan. And even better: it appears to work. 🎉 🏆 🎉 

Following your instructions above, I pushed https://github.com/pangeo-forge/staged-recipes/pull/52/commits/4bf58e8846b825d955cdf6bf76b70e55f8396660 to the FESOM recipe (which, as noted in https://github.com/pangeo-forge/staged-recipes/pull/52#issuecomment-877332968 was previously blocked by #93). With `pangeo-forge-recipes` installed from https://github.com/pangeo-forge/pangeo-forge-recipes/pull/166/commits/e379f4f625744cafd81c8474a4669723a7a38232 (my install is one commit behind this PR!), I've now stored about one third of the chunks for FESOM to OSN with no issues.

Noting for future reference to myself or others: for situations where a pre-#166 install of `pangeo-forge-recipes` was used to cache inputs, the metadata cache (if there is one) will need to be rebuilt before storing chunks with a post-#166 install:

```python
from pangeo_forge_recipes.recipes.xarray_zarr import cache_input_metadata

for input_name in rec.iter_inputs():
    cache_input_metadata(
        input_name,
        file_pattern=rec.file_pattern,
        input_cache=rec.input_cache,
        cache_inputs=rec.cache_inputs,
        copy_input_to_local_file=rec.copy_input_to_local_file,
        xarray_open_kwargs=rec.xarray_open_kwargs,
        delete_input_encoding=rec.delete_input_encoding,
        process_input=rec.process_input,
        metadata_cache=rec.metadata_cache,
    )
```
I'll try the eNATL60 recipe now and report back shortly.

cisaacstern:
Update: FESOM build using https://github.com/pangeo-forge/pangeo-forge-recipes/pull/166/commits/e379f4f625744cafd81c8474a4669723a7a38232 now complete and all chunks appear to be initialized as expected:

<details> 

```python
import s3fs
import zarr

endpoint_url = 'https://ncsa.osn.xsede.org'
fs_osn = s3fs.S3FileSystem(anon=True, client_kwargs={'endpoint_url': endpoint_url},)

fesom = "s3://Pangeo/pangeo-forge/swot_adac/FESOM/surf/fma.zarr"
group = zarr.open_consolidated(fs_osn.get_mapper(fesom))

for a in group.arrays():
    print(
        str(group[a[0]].info).split("Type")[0][:-1],
        str(group[a[0]].info).split("FSMap")[1], 
    )
```
```
Name               : /dflux 
No. bytes          : 17280000000 (16.1G)
Chunks initialized : 216/216

Name               : /lat 
No. bytes          : 8000 (7.8K)
Chunks initialized : 1/1

Name               : /lon 
No. bytes          : 8000 (7.8K)
Chunks initialized : 1/1

Name               : /ssh 
No. bytes          : 17280000000 (16.1G)
Chunks initialized : 216/216

Name               : /sss 
No. bytes          : 17280000000 (16.1G)
Chunks initialized : 216/216

Name               : /sst 
No. bytes          : 17280000000 (16.1G)
Chunks initialized : 216/216

Name               : /time 
No. bytes          : 17280 (16.9K)
Chunks initialized : 216/216

Name               : /tx_sur 
No. bytes          : 17280000000 (16.1G)
Chunks initialized : 216/216

Name               : /ty_sur 
No. bytes          : 17280000000 (16.1G)
Chunks initialized : 216/216

Name               : /u_surf 
No. bytes          : 17280000000 (16.1G)
Chunks initialized : 216/216

Name               : /v_surf 
No. bytes          : 17280000000 (16.1G)
Chunks initialized : 216/216

```
</details>

Will now re-install from https://github.com/pangeo-forge/pangeo-forge-recipes/pull/166/commits/d3e8b2cfd4c8fd4ed78136bfc63cd4484613a8b0 and try eNATL60.

cisaacstern:
As noted in today's coordination meeting, I was not able to successfully build the eNATL60 recipe from this PR. The `ValueError: I/O operation on closed file.` issue I encountered appears to be an upstream issue with `h5py` and is reproducible for the specific combination of this PR + eNATL60 recipe using the binder link provided here: https://github.com/cisaacstern/subset-enatl60

For the most minimal and generalized example of this error, see the example I've given in:

cisaacstern:
I've proposed some additional logging in https://github.com/rabernat/pangeo-forge/pull/2. Is this the best way to suggest edits to a big refactor like this (pull request to PR branch)? Seemed better than committing directly tp `rabernat:subset_chunks` so as not interfere with the main flow of the work

rabernat:
I think this is ready. Any sort of review would be appreciated.

In particular, it's important for @pangeo-forge/dev-team to grok the changes in indexing, summarized in the [docs](https://pangeo-forge--166.org.readthedocs.build/en/166/file_patterns.html#inspect-a-filepattern). The basic API has not changed, but the indexing objects returned by `Recipe.iter_inputs()` and `Recipe.iter_chunks()` are now a more complex data type. If you are used to constructing these indexes manually as tuples, then those workflows will break.

cisaacstern (changes requested):
> the changes in indexing

These make sense, and aesthetically seem to bring us into closer alignment with xarray's named dimensions. As such, I imagine they actually reduce the likelihood of human error. Very cool.

Only change I'd recommend (beyond inline type fix, below) is that we merge some version of https://github.com/rabernat/pangeo-forge/pull/2 to this PR. I've just updated it to reflect your comments, which led to a much nicer implementation. Here's the logging it now returns for the [eNATL60 example](https://github.com/cisaacstern/subset-enatl60) using the default threshold of 500 MB:
```
pangeo_forge_recipes.recipes.xarray_zarr - DEBUG - Converting variable votemper of 5346601680 bytes to `numpy.ndarray`
pangeo_forge_recipes.recipes.xarray_zarr - WARNING - Variable votemper of 5346601680 bytes is 10.69 times larger than specified maximum variable array size of 500000000 bytes. Consider re-instantiating recipe with `subset_inputs = {"time_counter": 11}`. If `len(ds["time_counter"])` < 11, substitute "time_counter" for any name in ds["votemper"].dims with length >= 11 or consider subsetting along multiple dimensions. Setting PANGEO_FORGE_MAX_MEMORY env variable changes the variable array size which will trigger this warning.
```

and with a non-default setting supplied via an env variable:
```python
import os
os.environ["PANGEO_FORGE_MAX_MEMORY"] = "100_000_000"
```
```
pangeo_forge_recipes.recipes.xarray_zarr - DEBUG - Converting variable votemper of 5346601680 bytes to `numpy.ndarray`
pangeo_forge_recipes.recipes.xarray_zarr - WARNING - Variable votemper of 5346601680 bytes is 53.47 times larger than specified maximum variable array size of 100000000 bytes. Consider re-instantiating recipe with `subset_inputs = {"time_counter": 54}`. If `len(ds["time_counter"])` < 54, substitute "time_counter" for any name in ds["votemper"].dims with length >= 54 or consider subsetting along multiple dimensions. Setting PANGEO_FORGE_MAX_MEMORY env variable changes the variable array size which will trigger this warning.
```

Without something like this users will encounter silent kernel crashes without being informed that `subset_inputs` can resolve their problems for them.

cisaacstern:
> the changes in indexing

These make sense, and aesthetically seem to bring us into closer alignment with xarray's named dimensions. As such, I imagine they actually reduce the likelihood of human error. Very cool.

Only change I'd recommend (beyond inline type fix, below) is that we merge some version of https://github.com/rabernat/pangeo-forge/pull/2 to this PR. I've just updated it to reflect your comments, which led to a much nicer implementation. Here's the logging it now returns for the [eNATL60 example](https://github.com/cisaacstern/subset-enatl60) using the default threshold of 500 MB:
```
pangeo_forge_recipes.recipes.xarray_zarr - DEBUG - Converting variable votemper of 5346601680 bytes to `numpy.ndarray`
pangeo_forge_recipes.recipes.xarray_zarr - WARNING - Variable votemper of 5346601680 bytes is 10.69 times larger than specified maximum variable array size of 500000000 bytes. Consider re-instantiating recipe with `subset_inputs = {"time_counter": 11}`. If `len(ds["time_counter"])` < 11, substitute "time_counter" for any name in ds["votemper"].dims with length >= 11 or consider subsetting along multiple dimensions. Setting PANGEO_FORGE_MAX_MEMORY env variable changes the variable array size which will trigger this warning.
```

and with a non-default setting supplied via an env variable:
```python
import os
os.environ["PANGEO_FORGE_MAX_MEMORY"] = "100_000_000"
```
```
pangeo_forge_recipes.recipes.xarray_zarr - DEBUG - Converting variable votemper of 5346601680 bytes to `numpy.ndarray`
pangeo_forge_recipes.recipes.xarray_zarr - WARNING - Variable votemper of 5346601680 bytes is 53.47 times larger than specified maximum variable array size of 100000000 bytes. Consider re-instantiating recipe with `subset_inputs = {"time_counter": 54}`. If `len(ds["time_counter"])` < 54, substitute "time_counter" for any name in ds["votemper"].dims with length >= 54 or consider subsetting along multiple dimensions. Setting PANGEO_FORGE_MAX_MEMORY env variable changes the variable array size which will trigger this warning.
```

Without something like this users will encounter silent kernel crashes without being informed that `subset_inputs` can resolve their problems for them.

Issue #79, opened 2021-03-01
Handling Secrets


TomAugspurger:
Just capturing a thought on how to handle secrets.

There are two main kinds:

1. Bakery secrets: Things like "credentials to write to the {azure / aws / google} bucket"
2. Recipe secrets: Things like "The password to access this password-protected HTTP server for this data" (https://github.com/pangeo-forge/pangeo-forge/issues/53)

The bakery secrets should never leave that bakery. In particular, things shouldn't be stored in GitHub and provided to the GitHub action kicking things off. That ensures that the maintainer of an AWS bakery doesn't have access to the Azure bakery's secrets.

Recipe-specific secrets will need to be handled separately. Since a recipe may run on multiple bakeries, these should probably be stored in GitHub and provided to the flow through the GitHub action (possibly set as an environment variable that's passed along to the worker executing the Prefect flow, and then looked up in the recipe). Since they're recipe-specific, they would ideally be stored in the repository's secrets (i.e. *not* the organization). The `staged-recipes` stuff slightly complicates this, since we need to ensure that the secret is copied over when the repository is created. I've never used it, but the GitHub API apparently supports working with GitHub secrets: https://docs.github.com/en/rest/reference/actions#secrets

rabernat:
Another option could be to store the secrets directly in the recipe repo but encrypted with a pangeo-forge public key, with the private key stored as an org-level secret. This still requires the recipe maintainer to trust the org with their secrets, but it might make things easier to manage.

scottyhq:
Sharing some notes on a recent foray into managing both types of secrets in github. 1) cloud-account keys as github secrets and 2) configuration secrets in encrypted text files with [mozilla SOPS](https://github.com/mozilla/sops)

I'm sure there are a million different ways to configure this stuff, but i've been impressed with using mozilla SOPS to keep encrypted data on github. what's nice about it is that you can still see the config, rather than just a binary file:
https://github.com/uwhackweek/jupyterhub-deploy/blob/main/hub/secrets.yaml 

[...]

PR #167, opened 2021-07-16
Make `fsspec_open_kwargs`, `query_string_secrets`, & `is_opendap` attributes of `FilePattern`

cisaacstern:
## Proposed feature description

Users may source data from APIs which authenticate via parameters (e.g., token, username, etc.) appended to a url path. This PR provides [`DropAPIParamsCache`](https://github.com/cisaacstern/pangeo-forge-recipes/blob/5caa883f47160c977540b63d39542cb0f3d40785/pangeo_forge_recipes/storage.py#L141-L159), a subclass of `CacheFSSpecTarget` that removes any of these (likely secret) API parameters from the cache filenames and logs generated at execution time.

> An assumption here is that there are situations where someone may have license to build a zarr store from password-protected data, but they don't want their credentials being reproduced on the execution worker system(s). See below for motivating recipe.

If we want to move forward with this feature, the following TODO items remain (assuming we preserve all existing logging):

- [ ] [`storage.file_opener`](https://github.com/cisaacstern/pangeo-forge-recipes/blob/5caa883f47160c977540b63d39542cb0f3d40785/pangeo_forge_recipes/storage.py#L178-L216) would need a conditional `public_fname` assignment; something like [this](https://github.com/cisaacstern/pangeo-forge-recipes/blob/scrub-tokens/pangeo_forge_recipes/storage.py#L247)
- [ ] `recipes.xarray_zarr.open_input` also needs a `public_fname` assignment before logging [here](https://github.com/cisaacstern/pangeo-forge-recipes/blob/5caa883f47160c977540b63d39542cb0f3d40785/pangeo_forge_recipes/recipes/xarray_zarr.py#L198) 

## Motivating recipe

This feature draft is motivated by @paigem's recipe https://github.com/pangeo-forge/staged-recipes/pull/56, which as described in the linked issue here https://github.com/pangeo-forge/staged-recipes/issues/46, sources files from NCAR's Climate Data Gateway via an HTTP API.

Once signing into the web GUI with valid credentials, the user is given an API token to append to the data url. For example, the first source url for the above-linked recipe is:

`https://tds.ucar.edu/thredds/fileServer/datazone/campaign/cesm/collections/ASD/v5_rel04_BC5_ne30_g16/ocn/proc/tseries/daily/v5_rel04_BC5_ne30_g16.pop.h.nday1.HMXL_2.00010101-01661231.nc?api-token=<TOKEN>`

where `<TOKEN>` is a secret.


cisaacstern:
This [`recipe.py`](https://github.com/pangeo-forge/staged-recipes/pull/56/files#diff-b8333b94b4a511704dda4ae72f4779a417751fd9bcb7e5f84375be1fff23bf96) for https://github.com/pangeo-forge/staged-recipes/pull/56 shows one way these secret API parameters could be appended to the file paths at execution time. Note the usage pattern for this recipe would be something like:

```python
from recipe import instantiate_recipe

api_token = "my-secret-token"
authenticated_recipe = instantiate_recipe(api_params=api_token)

# etc.
```
Automating this would presumably require an update to [the `meta.yaml` spec `recipes` section](https://github.com/pangeo-forge/roadmap/blob/master/doc/adr/0002-use-meta-yaml-to-track-feedstock-metadata.md#recipes-section) to permit something like:
```yaml
recipes:
  - callable: "recipe:instantiate_recipe"
    args: "api_params:{{ pointer to GitHub Secrets }}" 
```

AFAIK, the "recipe secrets" discussed in https://github.com/pangeo-forge/pangeo-forge-recipes/issues/79 have mostly been thought of as usernames and passwords which can be passed as `fsspec_open_kwargs` at execution time. This scenario is a little different, because the credentials need to be appended to the url paths.

cc @sharkinsspatial

rabernat:
> This scenario is a little different, because the credentials need to be appended to the url paths.

Just curious if it would also work if you put the token in the [HTTP Authorization header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization), rather than in a query string. That is a very common way to do token-based API access.

cisaacstern:
> Just curious if it would also work if you put the token in the HTTP Authorization header, rather than in a query string. 

As far as I can tell, this doesn't work for this particular source (NCAR Climate Data Gateway). I made a few attempts to pass this token to `fsspec` with various iterations of `{"headers": {"Authorization": "<type> <credentials>"}}`, none of which worked. I could do a more thorough job of documenting the Tracebacks I received if it would be useful, but perhaps a further clue resides in the fact that, when attempting to navigate to one of the source data urls __*without*__ appending a query string, we are redirected to a GUI login page; as seen, e.g., here: https://tds.ucar.edu/thredds/fileServer/datazone/campaign/cesm/collections/ASD/v5_rel04_BC5_ne30_g16/ocn/proc/tseries/daily/v5_rel04_BC5_ne30_g16.pop.h.nday1.HMXL_2.00010101-01661231.nc

This does not seem to match the usual pattern described in the MDN Web Docs, where:

> usually, but not necessarily, after the server has responded with a 401 Unauthorized status and the WWW-Authenticate header.

cisaacstern:
If we do want to move forward with a version of this feature, the implementation could alternatively be as an `AddAPIParamsCache` subclass of `CacheFSSpecTarget`, which appends the query string to the filename before caching. Assuming this does also work, it has the advantage of not requiring the divergent `meta.yaml` pattern described above.

rabernat:
Let's talk this through at today's upcoming call.

rabernat:
From my point of view, it seems much simpler to define the FilePattern without these secrets in the query string, and then figure out a way to pass extra secrets (key-value pairs) to be encoded into the URL.

cisaacstern:
> From my point of view, it seems much simpler to define the FilePattern without these secrets in the query string, and then figure out a way to pass extra secrets (key-value pairs) to be encoded into the URL.

Agreed. I *think* a subclass of `CacheFSSpecTarget` is the best way to do this, but open to all suggestions.

rabernat:
An alternative would be to put this in the `file_opener` function.

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/e1ef5756ced85e5f9b029e1c40291e133a73329b/pangeo_forge_recipes/storage.py#L172-L178

via an extra argument like `query_string_secrets`.

rabernat:
> All remote file opening goes through that function

This is wrong. We also open files here:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/e1ef5756ced85e5f9b029e1c40291e133a73329b/pangeo_forge_recipes/storage.py#L136-L139

cisaacstern:
@rabernat, this PR is ready for your re-review. Following reflection on your https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167#issuecomment-882801989, I *believe* this feature can be implemented with the very light touch represented here. AFAICT, declaration of [`CacheFSSpecTarget.cache_file.input_opener`](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/files#diff-1b0c20425676e0f245bdd49a8db3355e0e45139497bf42748276b7fd8705e5b1R161-R162) is actually the only line where we need to add in query string secrets, so defining an `input_fname` to be passed to that line only, looks like it allows us to authenticate in the source file context, without secrets leaking to `pangeo-forge-recipes` logs or cached filenames.

This PR required some small changes to the test suite, to get `test_storage.py` to pass. This is my first foray into the tests, so certainly won't be surprised if you have some notes there.

I do wonder if there may be a way to bring some of the new names into closer alignment with the spirit of ["Call things the same thing"](https://gist.github.com/tmcw/35849b7e9b86bb0c125972b2bb275bc7#call-things-the-same-thing), as I have ended up with

- `query_string_secrets` as the user-facing parameter in `XarrayZarrRecipe` (for expressiveness)
- `secrets` as the inner implementation of that parameter in `storage.py` and `test_storage.py` (for conciseness)
- `fake_secrets` for the fixture name in `conftest.py` (seemed descriptive)

...but maybe these are sufficient justifications for the naming heterogeneity.

And a final note: defining an `_add_query_string_secrets` function in `storage.py` is obviously a bit heavy-duty just to concatenate two strings, but I figured it's worth having so that we have a place to build out other ways of adding query string secrets in the future. In general, I think (?) it should not matter what order query string parameters are passed, so even if a particular url specifies other parameters aside from secrets, it will be fine to simply append the secrets at the end.

tests/test_storage.py
```diff
@@ -40,25 +40,41 @@ def test_metadata_target(tmp_metadata_target):
...
+    cache_first,
+    dask_cluster,
+    use_dask,
+    use_xarray,
+    use_query_string_secrets,
 ):
     all_paths, _ = file_paths
     path = str(all_paths[0])
     cache = tmp_cache if use_cache else None
+    secrets = fake_secrets if use_query_string_secrets and file_paths[0][-1] == "?" else None
```
rabernat, line 70: (I forgot to submit this comment in my review.)

I think `file_paths[0][-1] == "?"` will always evaluate to False, because the last character of the filename is never `?`.

pangeo_forge_recipes/storage.py
```diff
@@ -240,3 +241,7 @@ def _slugify(value: str) -> str:
     value = unicodedata.normalize("NFKD", value).encode("ascii", "ignore").decode("ascii")
     value = re.sub(r"[^.\w\s-]+", "_", value.lower())
     return re.sub(r"[-\s]+", "-", value).strip("-_")
+
+
+def _add_query_string_secrets(fname: str, secrets: str) -> str:
+    return fname + secrets
```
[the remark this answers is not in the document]
cisaacstern: I like this! Was not aware of `urllib.parse`

cisaacstern:
Thanks for these thoughtful suggestions. For clarity, I've converted this back to a draft, and will re-mark it as 'Ready for review' once I've incorporated this round of feedback.

tests/conftest.py
```diff
@@ -160,6 +160,13 @@ def teardown():
     return all_urls, items_per_file
 
 
+@pytest.fixture(scope="session")
+def netcdf_http_paths_with_secrets(netcdf_http_paths, fake_secrets):
+    all_urls, items_per_file = netcdf_http_paths
+    all_urls = [url + fake_secrets for url in all_urls]
```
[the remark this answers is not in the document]
cisaacstern: > Couldn't you just do url + "?a-pretend-api-token" and keep it a bit simpler?

I'm not clear how to implement this simpler variation. Ultimately, I think we want `netcdf_http_paths_with_secrets` to be a distinct parameterization of `test_file_opener`, as shown [here](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/files#diff-97a131ad6711fd2807ed014b7f48f566d7e07b56f29631077d6495bed8c4f459R43-R48)? Is there a way to achieve that *without* `netcdf_http_paths_with_secrets` being a fixture?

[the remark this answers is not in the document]
cisaacstern: ... oh, I think I see what you mean: we don't need a new `fake_secrets` fixture. But I think we *do* still need the new `netcdf_http_paths_with_secrets` fixture, correct?

tests/test_storage.py
cisaacstern: This confusion *should* ideally be resolved by https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/commits/c42151e6ddce7123e4349748ee54b059f9fb8fb0

tests/test_storage.py
```diff
@@ -40,25 +40,41 @@ def test_metadata_target(tmp_metadata_target):
...
 
     def do_actual_test():
         if cache_first:
-            cache.cache_file(path)
+            cache.cache_file(path, secrets)
```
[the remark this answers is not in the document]
cisaacstern: Good point. I think I've resolved this with https://github.com/pangeo-forge/pangeo-forge-recipes/commit/c42151e6ddce7123e4349748ee54b059f9fb8fb0

pangeo_forge_recipes/storage.py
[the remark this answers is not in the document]
cisaacstern: ✅  Implemented in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/commits/3680a34f8a729851fe9927ca3af906ae856d2658

pangeo_forge_recipes/storage.py
```diff
@@ -147,7 +147,7 @@ def _full_path(self, path: str) -> str:
 class CacheFSSpecTarget(FlatFSSpecTarget):
     """Alias for FlatFSSpecTarget"""
 
-    def cache_file(self, fname: str, **open_kwargs) -> None:
+    def cache_file(self, fname: str, secrets: Optional[str], **open_kwargs) -> None:
```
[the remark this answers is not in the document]
cisaacstern: ✅  Implemented in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/commits/78e2af3ef2d84c3bc60c22415b56531f84d4f7a8 plus https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/commits/d276d1ad06625fb2f092f80c65c8e79ea2e2d767

cisaacstern:
@rabernat, this is once again ready for re-review. I believe all of your inline comments have been addressed by the current changes. The current usage pattern would be:

```python
from example_recipe import recipe
recipe.query_string_secrets = "api_token=foobar"  # append secrets at runtime
# etc...
```
This has improved greatly with each iteration, so certainly welcome further changes if necessary.

tests/conftest.py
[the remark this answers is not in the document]
rabernat: Yes that makes sense.

tests/test_storage.py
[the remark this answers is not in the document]
rabernat: Yes, now I think secrets will be populated when we want it populated.

However, I'm still not convinced we are really testing that the _correct_ URL is being generated. the base test fixture `netcdf_http_paths` creates an http server with files at the specified URLs. This server doesn't care about the query string at all--it basically ignores it. So we could pass any `secrets`, or no `secrets`, and this test would still pass. So that means we can't count on this test to ensure that the secrets feature is working correctly. We could break it in the future and the test would not notify us.

The right thing to do is probably to refactor [http_auth_server.py](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/master/tests/http_auth_server.py) to add an additional option for a query string secret. You would have to
1. Refactor the command line arguments to allow for keyword (rather than position-based) arguments here 
   https://github.com/pangeo-forge/pangeo-forge-recipes/blob/c4867719c4866ec46436b5d53d95ceb470a6f72e/tests/http_auth_server.py#L8-L11
   [Clic](https://click.palletsprojects.com/en/8.0.x/) would probably be the best choice.
2. Add a new option, e.g. `--required-query-string=foobar`
3. Refactor the `netcdf_http_paths` test a bit to work with this new syntax
4. While you're at it, you could also add a test for standard username / password, which is currently possible but not actually implemented in a test.

rabernat (commented):
Great progress Charles!

We are almost there, but I think we are still falling short of actually testing whether the secrets feature works or not.

pangeo_forge_recipes/storage.py
```diff
@@ -240,3 +242,15 @@ def _slugify(value: str) -> str:
     value = unicodedata.normalize("NFKD", value).encode("ascii", "ignore").decode("ascii")
     value = re.sub(r"[^.\w\s-]+", "_", value.lower())
     return re.sub(r"[-\s]+", "-", value).strip("-_")
+
+
+def _add_query_string_secrets(fname: str, secrets: str) -> str:
+    parsed = urlparse(fname)
+    query = secrets if len(parsed.query) == 0 else f"{parsed.query}&{secrets}"
+    parsed = parsed._replace(query=query)
```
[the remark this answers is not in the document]
cisaacstern: Great point which I hadn't looked into before pushing, though strangely it seems this *is* the recommended API for named tuples?

"To prevent conflicts with field names, the method and attribute names start with an underscore."
https://docs.python.org/3/library/collections.html#collections.somenamedtuple._replace

"As is the case with all named tuples, the subclass has a few additional methods and attributes that are particularly useful. One such method is `_replace()`. The `_replace()` method will return a new ParseResult object replacing specified fields with new values.'
https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse

Reviewing these docs turned up [`parse_qs`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.parse_qs) and [`urlencode`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode), which will allow users to pass query string params as a dictionary. I'll include this upgrade in the forthcoming review-ready revision.

cisaacstern:
@rabernat, this is still a WIP, but wanted to get your 👀 on it before the tests refactor goes too deep.

First, a quick follow-up on https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167#discussion_r682956797. Extra `urllib` methods I found when tracking down the answer to your question about `_replace` have now made it so users can pass an arbitrary number of `query_string_secrets` as a dict, such as:
```python
from example_recipe import recipe
recipe.query_string_secrets = {"token": "abcd"}
```
All of the commits between your last review and https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/commits/c530e8e91df3a875e50a7cce9debd249aaf3a6f0 were aimed at implementing that change, plus refactoring `http_auth_server.py` (with a few associated tweaks in `conftest.py`) to allow the existing test suite to run with the addition of `click` for passing command line args to `http_auth_server.py`.

This was, in a sense, all set-up for actually adding a `--required-query-string=foobar` argument to the `netcdf_http_paths` fixture. When I got to that step, however, I noticed the following two TODO comments:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/c4867719c4866ec46436b5d53d95ceb470a6f72e/tests/conftest.py#L109-L111

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/c4867719c4866ec46436b5d53d95ceb470a6f72e/tests/conftest.py#L124-L127

...and it seemed like it'd be worth taking care of those TODOs first, rather than adding functionality to `netcdf_http_paths` which might just have to be changed again when that refactor happened. To begin that process, I refactored `netcdf_local_paths` and `netcdf_local_paths_by_variable` into a single fixture called [`netcdf_paths`](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R110-R126), and parameterized that fixture with two additional fixtures, `items_per_file` and `file_splitter`. This approach to getting a "product" of parameterizations is referenced from https://stackoverflow.com/a/22390931.  I then began the process of updating the other test modules to reflect this change.

In their current refactored forms, `test_storage.py` and `test_fixtures.py` both pass when run individually. `test_recipes.py` does not currently pass, however, and I wanted to get a quick go-no-go take on this whole approach before working any further on it.

tests/conftest.py
```diff
@@ -95,37 +95,42 @@ def _split_up_files_by_variable_and_day(ds, day_param):
...
-    items_per_file = {"D": 1, "2D": 2}[request.param]
-    path_format = str(tmp_path) + "/{variable}_{time:03d}.nc"
-    return full_paths, items_per_file, fnames_by_variable, path_format
+    items_per_file = {"D": 1, "2D": 2}[items_per_file]
+
+    if len(file_splitter_tuple) == 2:
+        return full_paths, items_per_file
+    else:
+        path_format = str(tmp_path) + "/{variable}_{time:03d}.nc"
+        return full_paths, items_per_file, fnames_by_variable, path_format
```
rabernat, line 126: It seems fragile to have a fixture return a variable number of items.

tests/conftest.py
```diff
@@ -95,37 +95,42 @@ def _split_up_files_by_variable_and_day(ds, day_param):
...
 
-# TODO: refactor to allow netcdf_local_paths_by_variable to be passed without
-# duplicating the whole test.
 @pytest.fixture(scope="session")
-def netcdf_http_paths(netcdf_local_paths, request):
-    paths, items_per_file = netcdf_local_paths
+def netcdf_http_paths(netcdf_paths, request):
+    paths, items_per_file = netcdf_paths[:2]
+    if len(netcdf_paths) == 4:
+        fnames_by_variable, path_format = netcdf_paths[2:]
```
rabernat, line 133: A comment would be really helpful to clarify what is happening here.

tests/conftest.py
```diff
@@ -185,10 +206,27 @@ def tmp_metadata_target(tmpdir_factory):
 
 @pytest.fixture
 def netCDFtoZarr_sequential_recipe(
-    daily_xarray_dataset, netcdf_local_paths, tmp_target, tmp_cache, tmp_metadata_target
+    daily_xarray_dataset, netcdf_paths, tmp_target, tmp_cache, tmp_metadata_target
```
rabernat, line 209: This is a big change. Now all tests that accepted the old  `netcdf_local_paths` have to also work with multi-variable paths.

rabernat (commented):
A few random comments.

It would be really helpful to understand better what tests are failing. I looked at the [log output](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/checks?check_run_id=3247287808) but it was a bit overwhelming.

Clearly the refactoring has broken some tests. Do you have any sense of what is broken and why?

rabernat:
> I thought it best to get your go-no-go take on this whole approach before working any further on it.

I think you are on the right track here. However, certain tests are written specifically for certain fixtures. We need to get a better understanding of what tests have been broken by this refactoring and why.

rabernat:
Here is the full list of failing tests

```
FAILED tests/test_recipes.py::test_recipe[manual-netCDFtoZarr_sequential_subset_recipe-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_recipe[python-netCDFtoZarr_sequential_subset_recipe-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_recipe[dask-netCDFtoZarr_sequential_subset_recipe-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_recipe[prefect-netCDFtoZarr_sequential_subset_recipe-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_recipe[prefect-dask-netCDFtoZarr_sequential_subset_recipe-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_process[manual-netCDFtoZarr_sequential_subset_recipe-None-None-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_process[manual-netCDFtoZarr_sequential_subset_recipe-incr_date-None-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_process[manual-netCDFtoZarr_sequential_subset_recipe-None-incr_date-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_process[manual-netCDFtoZarr_sequential_subset_recipe-incr_date-incr_date-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_process[python-netCDFtoZarr_sequential_subset_recipe-None-None-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_process[python-netCDFtoZarr_sequential_subset_recipe-incr_date-None-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_process[python-netCDFtoZarr_sequential_subset_recipe-None-incr_date-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_process[python-netCDFtoZarr_sequential_subset_recipe-incr_date-incr_date-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_process[dask-netCDFtoZarr_sequential_subset_recipe-None-None-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_process[dask-netCDFtoZarr_sequential_subset_recipe-incr_date-None-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_process[dask-netCDFtoZarr_sequential_subset_recipe-None-incr_date-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_process[dask-netCDFtoZarr_sequential_subset_recipe-incr_date-incr_date-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_process[prefect-netCDFtoZarr_sequential_subset_recipe-None-None-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_process[prefect-netCDFtoZarr_sequential_subset_recipe-incr_date-None-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_process[prefect-netCDFtoZarr_sequential_subset_recipe-None-incr_date-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_process[prefect-netCDFtoZarr_sequential_subset_recipe-incr_date-incr_date-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_process[prefect-dask-netCDFtoZarr_sequential_subset_recipe-None-None-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_process[prefect-dask-netCDFtoZarr_sequential_subset_recipe-incr_date-None-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_process[prefect-dask-netCDFtoZarr_sequential_subset_recipe-None-incr_date-2D-_split_up_files_by_variable_and_day]
FAILED tests/test_recipes.py::test_process[prefect-dask-netCDFtoZarr_sequential_subset_recipe-incr_date-incr_date-2D-_split_up_files_by_variable_and_day]
```

rabernat:
Ok, I think I found the problem. I think the issue here:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/92078e923f7da55bf613162912735226205897c5/tests/conftest.py#L239-L246

You can't use `pattern_from_file_sequence` to define the FilePattern when there is a `MergeDim`. So just like `netCDFtoZarr_sequential_recipe`, this fixture needs to define the FilePattern different based on the contents of `netcdf_paths`. Perhaps the optimal general solution is to make the FilePattern itself a fixture?

tests/conftest.py
```diff
@@ -184,11 +218,11 @@ def tmp_metadata_target(tmpdir_factory):
...
 
 @pytest.fixture
-def netCDFtoZarr_sequential_recipe(
-    daily_xarray_dataset, netcdf_local_paths, tmp_target, tmp_cache, tmp_metadata_target
+def netCDFtoZarr_recipe(
+    daily_xarray_dataset, netcdf_paths, tmp_target, tmp_cache, tmp_metadata_target
 ):
-    paths, items_per_file = netcdf_local_paths
-    file_pattern = pattern_from_file_sequence([str(path) for path in paths], "time", items_per_file)
+    file_pattern = _make_file_pattern(netcdf_paths)
```
rabernat, line 224: Could potentially go one step further and make file_pattern itself the fixture.

cisaacstern:
I realized the one aspect I'm not yet testing here is that all of the auth kwargs are being correctly passed through to the various function calls in `xarray_zarr.py`. Started working up a test for that and already found what appears to be an issue. Will check back shortly with a follow-up.

pangeo_forge_recipes/storage.py
```diff
@@ -240,3 +242,17 @@ def _slugify(value: str) -> str:
...
+    query = parse_qs(parsed.query)
+    for k, v in secrets.items():
+        query.update({k: v})
+    parsed = parsed._replace(query=urlencode(query, doseq=True))
+    return urlunparse(parsed)
+
+
+def _get_opener(fname: str, secrets: Optional[dict], **open_kwargs):
+    fname = fname if not secrets else _add_query_string_secrets(fname, secrets)
+    return fsspec.open(fname, mode="rb", **open_kwargs)
```
rabernat, line 258: I like this way of doing it. :+1:

tests/http_auth_server.py
```diff
@@ -1,33 +1,47 @@
...
+                    self.send_response(401)
+                    self.send_header("WWW-Authenticate", "Basic")
+                    self.end_headers()
+                    return
+            if required_query_string:
+                query = urlparse(self.path).query
+                if query != required_query_string:
+                    self.send_response(400)
+                    self.end_headers()
+                    return
```
rabernat, line 38: :+1:

tests/test_recipes.py
```diff
@@ -65,11 +112,50 @@ def test_prune_recipe(recipe_fixture, execute_recipe, nkeep):
...
+    """Test that caching and copying from http to local file work."""
+
+    RecipeClass, file_pattern, kwargs, ds_expected, target = netCDFtoZarr_http_recipe
+
+    if len(file_pattern.merge_dims) != 0:
+        pytest.skip(
+            "This test ensures auth kwargs are correctly passed to the recipe class."
+            "`netcdf_http_file_pattern` fixture currently does not return a properly"
+            " formatted `path_format` for recipes with a `MergeDim.`"
+        )
```
rabernat, line 148: Is this something we can fix in the future?

tests/test_recipes.py
```diff
@@ -65,11 +112,50 @@ def test_prune_recipe(recipe_fixture, execute_recipe, nkeep):
...
+
+    first_url = list(file_pattern.items())[0][1]
+    r = requests.get(first_url)
+    if r.status_code == 401:
+        fsspec_open_kwargs = dict(auth=aiohttp.BasicAuth("foo", "bar"))
+        kwargs.update({"fsspec_open_kwargs": fsspec_open_kwargs})
+    elif r.status_code == 400:
+        query_string_secrets = {"foo": "foo", "bar": "bar"}
+        kwargs.update({"query_string_secrets": query_string_secrets})
```
rabernat, line 157: This seems like a very roundabout way to figure out if we need to use secrets. Perhaps we should be embedding the secrets _inside_ the FilePattern object.

tests/test_storage.py
```diff
@@ -40,29 +42,45 @@ def test_metadata_target(tmp_metadata_target):
...
 ):
-    all_paths, _ = file_paths
+    all_paths = file_paths[0]
     path = str(all_paths[0])
     cache = tmp_cache if use_cache else None
 
+    if path.split(":")[0] == "http":
+        r = requests.get(path)
+        open_kwargs = dict(auth=aiohttp.BasicAuth("foo", "bar")) if r.status_code == 401 else {}
+        secrets = {"foo": "foo", "bar": "bar"} if r.status_code == 400 else {}
```
rabernat, line 70: Same.

rabernat (commented):
This looks great Charles! This was a complex and difficult feature to implement, and I'm convinced your implementation works well and is well tested.

We could definitely merge this as is. However, looking at the code, I think I see an opportunity for a major simplification that will make things more readable / maintainable.

_What if all of these secrets live with the `FilePattern` object?_ Not as part of the URLs, but as a separate attribute, e.g. `username`, `password`, `query_string_secrets`. That would save as needing to keep passing these attributes around as keyword args in `XarrayZarrRecipe`. It would also make the testing easier, since these could be set by the FilePattern fixtures.

What do you think? I think you've already figured out the hard parts here, so hopefully making this change would be straightforward.

tests/test_recipes.py
cisaacstern: Yep! Fixed by https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/commits/ba9f7904fd79fd74dcf8f0bf8e62c094efe3af3c

rabernat:
To summarize the outcome of our discussion yesterday, the path towards the refactor I suggested would be as follows.

- [ ] Add additional attributes to the [FilePattern](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/690594a86d1115efe2ab229d7c46f464396a7110/pangeo_forge_recipes/patterns.py#L121) class to contain the additional attributes related to file opening / access / secrets. Basically, the file pattern should contain all information needed to actually open the files. Note we may want to consider turning it back into a dataclass.
- [ ] Add a test for these new attributes in `test_patterns.py`. Can be very basic, but still needs to exist.
- [ ] Update the narrative documentation for patterns to explain how to use the new attributes.
- [ ] Refactor the XarrayZarrRecipe class to access these attributes from the FIlePattern, removing some of the corresponding attributes from XarrayZarrRecipe itself
- [ ] Update `conftest.py` and `test_recipes.py` to leverage this information, passing the secrets via the FilePattern fixtures

cisaacstern:
Thanks for the summary, @rabernat. I'm making good headway on this and should have something for your review later today.

A few questions:

1. What's the advantage of FilePattern being a dataclass?
2. Would you agree that if we're putting `fsspec_open_kwargs` and `query_string_secrets` into FilePattern, that `is_opendap` should go in there too? (It's another source-file attribute which get's passed around a lot in XarrayZarrRecipe.)
3. Re: `is_opendap`, while I'm at it should I write a test for this by spinning up a `pydap` server, like we do for http?

rabernat:
> 1. What's the advantage of FilePattern being a dataclass?

Basically less boilerplate. With a dataclass, you would only have to add a new class attribute to the class and it would just work. For a regular class, you have to add arguments to `__init__` and then assign them explicitly. Not a big deal either way really.

> 2\. Would you agree that if we're putting `fsspec_open_kwargs` and `query_string_secrets` into FilePattern, that `is_opendap` should go in there too? (It's another source-file attribute which get's passed around a lot in XarrayZarrRecipe.)

Agree.

> 3\. Re: `is_opendap`, while I'm at it should I write a test for this by spinning up a `pydap` server, like we do for http?

This would be a lot of work. I would pass on it for now but make a note in a separate issue that we should do it some day.

cisaacstern:
I believe all of the items from https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167#issuecomment-909456236 are now complete. A few notes:

> Add additional attributes to the FilePattern class ... related to file opening / access / secrets. 

AFAICT without changing the implementation of `*combine_dims` as *args in the `FilePattern.__init__` call signature, our options for adding these kwargs are:
1. Declare them explicitly by name, requiring that they go before  `*combine_dims` in the call signature (doesn't seem right; `*combine_dims` should come first)
2. Pass them after `*combine_dims` as **kwargs

I've opted for the latter, but reject all but the three supported kwargs, as shown [here](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/files?file-filters%5B%5D=.cfg&file-filters%5B%5D=.ipynb&file-filters%5B%5D=.py#diff-3cac813ae555a03fca1cc91d89db56980bbbf8d67ad8c9d967e0eeb9a880dfd0R140-R154). Given that these are passed as **kwargs, is it inadvisable to list them as function params, as I've done [here](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/files?file-filters%5B%5D=.cfg&file-filters%5B%5D=.ipynb&file-filters%5B%5D=.py#diff-3cac813ae555a03fca1cc91d89db56980bbbf8d67ad8c9d967e0eeb9a880dfd0R132-R137)? From the user perspective, these act like explicit parameters, but from a Python internals standpoint, that may be a bit of a misnomer.

> Add a test for these new attributes in test_patterns.py. 

I'm testing this within `test_file_pattern_concat_merge` [here](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/files?file-filters%5B%5D=.cfg&file-filters%5B%5D=.ipynb&file-filters%5B%5D=.py#diff-a675dbfa8f98feaef8a03fc4b3507922770846c09062ea9faa0ccd4c28195fe9R141-R153) to ensure that instantiating and/or updating these kwargs isn't affecting the core functionality of FilePattern.

>  Update the narrative documentation for patterns to explain how to use the new attributes.

☑️ [here](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/files#diff-e3da1598b7ee355f6180101729239375d2b70aba980da5fc41a65681c5fd55dfR116-R136)

> Refactor the XarrayZarrRecipe class to access these attributes from the FIlePattern

☑️ [here](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/files?file-filters%5B%5D=.cfg&file-filters%5B%5D=.ipynb&file-filters%5B%5D=.py#diff-e12c886cc124886c5cfa5313d760a36c39649af9da845077c663e6feab8487b5)

> Update conftest.py and test_recipes.py to leverage this information, passing the secrets via the FilePattern fixtures

- **conftest.py**: path fixtures (`netcdf_local_paths` and `netcdf_http_paths`) now return 5-tuples, with the added item being these new kwargs. They are always empty for local paths and [populated according to the server's parametrization](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/files?file-filters%5B%5D=.ipynb&file-filters%5B%5D=.py#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R212-R215) for http paths. These are then passed to the FilePattern instance [here](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/files?file-filters%5B%5D=.py#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R121).
- **test_recipes.py**: the beauty of putting this all into FilePattern is that the footprint in test_recipes.py is essentially nonexistent. All of the recipes already get a FilePattern, of course. And if that object comes from the [`netcdf_http_file_pattern` fixture](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/files?file-filters%5B%5D=.py#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R160-R162) (as opposed to the local pattern fixture) and requires some form of authentication, then it comes prepackaged with all the information it needs. (Whether that be `fsspec_open_kwargs` or `query_string_secrets`.)

tests/test_recipes.py
```diff
@@ -268,8 +272,13 @@ def test_chunks(
...
 
 
-def test_lock_timeout(netCDFtoZarr_sequential_recipe, execute_recipe):
-    RecipeClass, file_pattern, kwargs, ds_expected, target = netCDFtoZarr_sequential_recipe
+def test_lock_timeout(netCDFtoZarr_recipe, execute_recipe):
+    RecipeClass, file_pattern, kwargs, ds_expected, target = netCDFtoZarr_recipe
+
+    # `netCDFtoZarr_recipe` fixture is parametrized. We don't need to run this test more than once.
+    if len(file_pattern.merge_dims) != 0:
+        pytest.skip("It's redundant to run this test more than once.")
```
rabernat, line 280: In general it's not redundant to run the tests more than once. Each `netCDFtoZarr_recipe` instance is different. Running the same test on different inputs usually gives us better coverage. We are skipping here  because we think there are actually not meaningful differences between each parametrized run; that probably indicates that we are using the wrong fixtures in the first place.

tests/test_recipes.py
```diff
@@ -129,11 +110,34 @@ def test_prune_recipe(recipe_fixture, execute_recipe, nkeep):
...
+    execute_recipe(rec)
+    ds_actual = xr.open_zarr(target.get_mapper()).load()
+    xr.testing.assert_identical(ds_actual, ds_expected)
+
+
+@pytest.mark.parametrize("cache_inputs", [True, False])
+@pytest.mark.parametrize("copy_input_to_local_file", [True, False])
+def test_recipe_http_caching_copying(
+    netCDFtoZarr_http_recipe, execute_recipe, cache_inputs, copy_input_to_local_file
+):
```
rabernat, line 136: This test seems nearly identical to the one above. Could we use a `lazy_fixture` to combine `netCDFtoZarr_recipe` and `netCDFtoZarr_http_recipe` into a single fixture for a single `test_recipe_caching_copying`?

tests/test_references.py
```diff
@@ -15,7 +15,9 @@
 
 @pytest.mark.parametrize("with_intake", [True, False])
 def test_single(netcdf_local_paths, tmpdir, with_intake):
-    full_paths, items_per_file = netcdf_local_paths
+    full_paths, items_per_file, fnames_by_variable = netcdf_local_paths[:3]
+    if fnames_by_variable:
+        pytest.skip("This does not test merging operations.")
```
rabernat, line 20: Now that we have #174 merged _and_ we have FilePatterns as fixtures, we should refactor these test to use the FilePattern fixtures (rather than `netcdf_local_paths`).

tests/test_references.py
```diff
@@ -47,7 +49,9 @@ def test_single(netcdf_local_paths, tmpdir, with_intake):
 
 @pytest.mark.parametrize("with_intake", [True, False])
 def test_multi(netcdf_local_paths, tmpdir, with_intake):
-    full_paths, items_per_file = netcdf_local_paths
+    full_paths, items_per_file, fnames_by_variable = netcdf_local_paths[:3]
+    if fnames_by_variable:
+        pytest.skip("This does not test merging operations.")
```
rabernat, line 54: Same; let's use the FilePattern fixtures here.

tests/test_references.py
rabernat: We might have to define a new FilePattern fixture that contains just a single file for this test. That should be done in `conftest.py`.

rabernat (commented):
This is really great Charles! We are super close on this very complicated and much needed test refactor.

I just have one more round of comments and then I'm sure we'll be ready to merge. Thanks for your patience and perseverance here.

docs/recipe_user_guide/file_patterns.md
```diff
@@ -113,7 +113,28 @@ and type of combine dimensions they support.
...
 
-### Specifying `nitems_per_input` in a `ConcatDim`
+### Extra keyword arguments for `FilePattern`
+
+`FilePattern` objects carry all of the information needed to open source files. The following additional keyword
+arguments may passed to `FilePattern` instances as appropriate:
+
+- `fsspec_open_kwargs`: A dictionary of kwargs to pass to `fsspec.open` to aid opening of source files. For example,
+`{"block_size": 0}` may be passed if an HTTP source file server does not permit range requests. Authentication for
+`fsspec`-compatible filesystems may be handled here as well.
```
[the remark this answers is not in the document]
cisaacstern: ☑️  https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/commits/f4b2bb42c8a5062ca95db45b55e78c7b270a3db4

pangeo_forge_recipes/patterns.py
```diff
@@ -129,11 +129,29 @@ class FilePattern:
...
 
-    def __init__(self, format_function: Callable, *combine_dims: CombineDim):
+    def __init__(
+        self, format_function: Callable, *combine_dims: CombineDim, **kwargs,
+    ):
         self.format_function = format_function
         self.combine_dims = combine_dims
+        self.fsspec_open_kwargs = kwargs.pop("fsspec_open_kwargs", {})
+        self.query_string_secrets = kwargs.pop("query_string_secrets", {})
+        self.is_opendap = kwargs.pop("is_opendap", False)
```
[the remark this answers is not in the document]
cisaacstern: ☑️ https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/commits/efbf3354bf3a05ee31800ed12c28c6aaeff58c1f

Oh! I hadn't realized that this was an allowable signature:

```python
Signature: func(positional_arg, *args, default_kwarg=None)
```

(I'd mistakenly thought *args needed to come after default kwargs.)

Agreed that this is *way* better.

I've set both the `fsspec_open_kwargs` and `query_string_secrets` defaults to an empty dict (rather than None), because there may be scenarios where we want to call, e.g., `fsspec_open_kwargs.update()` at runtime.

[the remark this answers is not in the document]
rabernat: > defaults to an empty dict (rather than None), because there may be scenarios where we want to call, e.g., `fsspec_open_kwargs.update()` at runtime.

This is actually a common python pitfall, as the following code will illustrate.

```python
class Foo
    def __init__(self, bar={}):
        self.bar = bar

f1 = Foo()
f2 = Foo()
f1.bar.update({'a': 1})
print(f2.bar) # what do you think it will be?
```

This is [solved in dataclasses](https://docs.python.org/3/library/dataclasses.html#dataclasses.field) by using the `default_factory=dict`. If we don't go with dataclasses, you need to do something like
```python
if fsspec_open_kwargs is None:
    fsspec_open_kwargs = {}
```
inside `__init__`.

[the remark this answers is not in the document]
cisaacstern: 🤯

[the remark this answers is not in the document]
cisaacstern: For a dataclass to support variable numbers of `*args` (i.e., `*combine_dims`), we need to set `@dataclass(init=False)` and override the default `__init__`: https://www.python.org/dev/peps/pep-0557/#custom-init-method

Expanding on your toy example, that would look like this, for our use case (with incorrect types, for illustration):
```python
from dataclasses import dataclass

@dataclass(init=False)
class FilePattern:
    format_function: str
    combine_dims: int
    bar: dict
    def __init__(self, format_function, *combine_dims, bar=None):
        self.format_function = format_function
        self.combine_dims = combine_dims
        self.bar = bar if bar else {}

f1 = FilePattern("func", 0, 1)
f2 = FilePattern("func", 0)
f1.bar.update({'a': 1})
print(
    f"f1: {vars(f1)}\n"
    f"f2: {vars(f2)}"
)
```
```
f1: {'args': (0, 1), 'bar': {'a': 1}}
f2: {'args': (0,), 'bar': {}}
```
This means we can't leverage `field=(default_factory=dict)` (which is passed to the default dataclass `__init__`) and instead need to fall back to what you've proposed for non-dataclasses, i.e. `self.bar = bar if bar else {}`.

The only problem with this is that the dataclass type hints don't make it into the `__init__` signature (unless we put them there, redundantly), and will therefore be absent from the auto-generated API documentation for this class. I checked this with `make html`, but you can see it more simply here
```python
FilePattern?
```
```python
Init signature: FilePattern(format_function, *combine_dims, bar=None)
Docstring:      FilePattern(format_function, *combine_dims, bar=None)
Type:           type
Subclasses:   
```
As such, I suggest we go with the more concise non-dataclass approach:

```python
class FilePattern:
    def __init__(self, format_function: str, *combine_dims: int, bar: dict=None):
        self.format_function = format_function
        self.combine_dims = combine_dims
        self.bar = bar if bar else {}

f1 = FilePattern("func", 0, 1)
f2 = FilePattern("func", 0)
f1.bar.update({'a': 1})
print(
    f"f1: {vars(f1)}\n"
    f"f2: {vars(f2)}"
)
```
```
f1: {'format_function': 'func', 'combine_dims': (0, 1), 'bar': {'a': 1}}
f2: {'format_function': 'func', 'combine_dims': (0,), 'bar': {}}
```
Which preserves type hints that are propagated to sphinx:
```
FilePattern?
```
```
Init signature: FilePattern(format_function: str, *combine_dims: int, bar: dict = None)
Docstring:      <no docstring>
Type:           type
Subclasses: 
```

Thoughts?

tests/test_recipes.py
cisaacstern: 👍 https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167/commits/72da50fba4e023bbe170bf9e52ac86211c80934a

pangeo_forge_recipes/patterns.py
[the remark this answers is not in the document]
rabernat: > For a dataclass to support variable numbers of `*args` (i.e., `*combine_dims`), we need to set `@dataclass(init=False)` and override the default `__init__`

Gah I did not realize this. Maybe not worth it then.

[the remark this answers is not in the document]
rabernat: > As such, I suggest we go with the more concise non-dataclass approach:

:+1:

rabernat (approved):

PR #168, opened 2021-07-19
remove ad-hoc stage skipping

rabernat:
The `BaseRecipe` class includes some logic for skipping the `cache_inputs` stage of recipe execution when the `cache_inputs` property is set to True on the recipe instance.

This PR removes that logic. If merged, the `cache_inputs` stage will always run.

### Why?

Two reasons:

1. It feels very ad-hoc. We should replace it with a formal mechanism for recipes to register which stages should be run. Ideally the number of stages could be arbitrary.
2. The logic is not actually correct. Consider the case where we don't want to cache inputs (so set `cache_inputs` to False) but DO need to cache input metdata. Because input metadata caching is implemented within the `cache_input` function, we still need to run `cache_input` on each input, even if we don't actually want to cache the file itself. This could also be solved with a separate stage (e.g. make `cache_input_metdata` a top-level stage of the recipe execution). I discovered this bug in developing a tutorial using OpenDAP inputs in #166.

### Will this cause a problem?

That depends. Is anyone (e.g. @cisaacstern or @TomAugspurger) relying on the fact that  `cache_inputs=False` causes the entire stage to be skipped. The worst case scenario here is that we will now run a bunch of tasks that do nothing for each input.

### DIFF
--- a/pangeo_forge_recipes/recipes/base.py
+++ b/pangeo_forge_recipes/recipes/base.py
@@ -85,8 +85,8 @@ class BaseRecipe(ABC):
             FutureWarning,
         )
         pipeline = []  # type: MultiStagePipeline
-        if getattr(self, "cache_inputs", False):  # TODO: formalize this contract
-            pipeline.append(Stage(self.cache_input, list(self.iter_inputs())))
+        # TODO: allow recipes to customize which stages to run
+        pipeline.append(Stage(self.cache_input, list(self.iter_inputs())))
         pipeline.append(Stage(self.prepare_target))
         pipeline.append(Stage(self.store_chunk, list(self.iter_chunks())))
         pipeline.append(Stage(self.finalize_target))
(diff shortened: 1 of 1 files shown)


cisaacstern:
> Is anyone relying on the fact that cache_inputs=False causes the entire stage to be skipped. 

Not me.

> (e.g. make cache_input_metdata a top-level stage of the recipe execution)

Once some version of #166 is merged, re-caching metadata for pre-cached input files will briefly become a more common practice (for me, at least). This can be accomplished without `cache_input_metadata` being a recipe method, but including it as its own recipe step does make sense to me. I can imagine it being useful in other contexts. (e.g., rebuilding recipes in contexts with different amounts of memory from the first build, therefore requiring different metadata for the #166 logic)

TomAugspurger (approved):

PR #171, opened 2021-07-20
use xr.open_dataset context manager

### LINKED ISSUE #170: Upstream `h5py` closed file error, `ValueError: I/O operation on closed file.`
A minimal reproducible example is provided via the binder link here: https://github.com/cisaacstern/io-on-closed-file. (This example follows @rabernat's suggestion to repeat the same file name twice in a FilePattern.)

This issue was most recently noted in a real execution setting in relation to #166 ([comment here](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/166#issuecomment-882972419)) and was previously discussed in #163. Unlike the latter case, the real example connected to #166 is not resolvable by restarting the kernel, because the cached file is intentionally opened multiple times by the subsetting logic.

@martindurant, the most interesting thing I realized while developing the above-linked binder example, is that this closed file issue does not appear to arise if the input cache is on an `fsspec.LocalFileSystem`; it occurs only on `gcsfs.GCSFileSystem` (I haven't yet tried s3fs). Is this expected?

It's possible, therefore, that setting `copy_input_to_local_file=True` in the recipe may be present a sort-term (if inefficient) workaround. I will try this tomorrow and report back.

rabernat:
Fixes #170?

Rather than doing `ds = xr.open_dataset()`, we now use `xr.open_dataset` as a contextmanager. This _should_ fix the "I/O operation on a closed file" problem.

Now that we have moved to a more functional structure for the code (post #160), it would be great to have a more atomic set of unit tests for the different functions. As is, it is hard to write a tests to cover this specific issue.

### DIFF
--- a/pangeo_forge_recipes/recipes/xarray_zarr.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr.py
@@ -199,23 +199,23 @@ def open_input(
     cache = input_cache if cache_inputs else None
     with file_opener(fname, cache=cache, copy_to_local=copy_input_to_local_file) as f:
         with dask.config.set(scheduler="single-threaded"):  # make sure we don't use a scheduler
-            logger.debug(f"about to call xr.open_dataset on {f}")
             kw = xarray_open_kwargs.copy()
             if "engine" not in kw:
                 kw["engine"] = "h5netcdf"
-            ds = xr.open_dataset(f, **kw)
-            logger.debug("successfully opened dataset")
-            ds = fix_scalar_attr_encoding(ds)
+            logger.debug(f"about to enter xr.open_dataset context on {f}")
+            with xr.open_dataset(f, **kw) as ds:
+                logger.debug("successfully opened dataset")
+                ds = fix_scalar_attr_encoding(ds)
 
-            if delete_input_encoding:
-                for var in ds.variables:
-                    ds[var].encoding = {}
+                if delete_input_encoding:
+                    for var in ds.variables:
+                        ds[var].encoding = {}
 
-            if process_input is not None:
-                ds = process_input(ds, str(fname))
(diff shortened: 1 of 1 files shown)


cisaacstern:
Awesome. Confirming that installing from here does resolve the `I/O on closed file` error in the minimal "mock" recipe https://github.com/cisaacstern/io-on-closed-file

And installing from [cisaacstern/pangeo-forge-recipes@subset-plus-contexts-fix](https://github.com/cisaacstern/pangeo-forge-recipes/tree/subset-plus-contexts-fix) (a merge of current commits in #166 and #171) resolves the `I/O on closed file` error in the actual eNATL60 recipe https://github.com/cisaacstern/subset-enatl60

Two questions:

1. Are these additional contexts something you'd want to remove eventually, if/when the root cause of this issue is identified & fixed; if so, perhaps we should flag that with an inline comment?
2. Could you share anything further about why you see it as "hard to write a test to cover this specific issue."?

Otherwise, this is ready to merge IMO

cisaacstern (approved):

rabernat:
> 1. re these additional contexts something you'd want to remove eventually

No. I think it's better to keep them. As @martindurant has pointed out, they are a bit unnecessary for purely HTTP-based filesystems, but there are also people who will run pangeo_forge_recipes against actual files on a disk. These context managers help us maintain consistent file state through all the different functions.

> 2\. Could you share anything further about why you see it as "hard to write a test to cover this specific issue."?

We do not have a unit test for the `open_input` function. We just have big integrations tests that are really complex and slow. Ideally we would be testing each function individually in an isolated way. Then we could simply write a test where we call `open_input` twice in a row and verify that it works. As is, we would have to define a whole new recipe fixture.

PR #175, opened 2021-07-22
Strided streaming transfer logs

cisaacstern:
As described in https://github.com/pangeo-forge/staged-recipes/pull/56#issuecomment-885295982, with the `"pangeo_forge_recipes"` logging level at DEBUG, streaming transfers (`fsspec_open_kwargs = {"block_size": 0}`) have the potential to write so many lines to `stdout` so quickly (>1000 lines/second) that Jupyter issues a warning

<details>
    <summary> IOPub message rate exceeded </summary>

```
IOPub message rate exceeded.
The notebook server will temporarily stop sending output
to the client in order to avoid crashing it.
To change this limit, set the config variable
`--NotebookApp.iopub_msg_rate_limit`.

Current values:
NotebookApp.iopub_msg_rate_limit=1000.0 (msgs/sec)
NotebookApp.rate_limit_window=3.0 (secs)
```

</details>

and temporarily stops rendering output. This is, of course, undesirable when manually executing a recipe, though I can't imagine producing that much logging on a Bakery worker would be so great either. Of course, one solution is just to turn the logging level to INFO, but long file transfers can easily be measured in hours, and it is best if there can be some debug logging during that process, to ensure it is not hanging.

This PR proposes "strided logging" for the case of streaming downloads. Default range request DEBUG logs are unchanged 
```
pangeo_forge_recipes.storage - DEBUG - _copy_btw_filesystems reading data
pangeo_forge_recipes.storage - DEBUG - _copy_btw_filesystems copying block of 10000000 bytes
```
but default streaming logs now look like this
```
pangeo_forge_recipes.storage - DEBUG - copying block 0 of ~10000000 bytes
pangeo_forge_recipes.storage - DEBUG - copying block 1 of ~10000000 bytes
pangeo_forge_recipes.storage - DEBUG - copying block 2 of ~10000000 bytes
```
and the user can also adjust the stride (i.e. frequency) of these streaming logs with an environment variable
```python
import os
os.environ["PANGEO_FORGE_BLOCK_SIZE"] = "50_000_000"
```
```
pangeo_forge_recipes.storage - DEBUG - copying block 0 of ~50000000 bytes
pangeo_forge_recipes.storage - DEBUG - copying block 1 of ~50000000 bytes
pangeo_forge_recipes.storage - DEBUG - copying block 2 of ~50000000 bytes
```
the latter option is useful in situations such as https://github.com/pangeo-forge/staged-recipes/pull/56, in which 10 MB logging increments are still unnecessarily fast/frequent.

(Of course, the addition of this environment variable hypothetically provides the option to adjust the default `BLOCK_SIZE` for range request transfers, which I don't believe was previously possible, though I have not actually tested it in this capacity.)

@martindurant, can you think of any downsides to this logging approach?


pangeo_forge_recipes/storage.py
```diff
@@ -25,11 +25,15 @@ def _get_url_size(fname, **open_kwargs):
     return size
 
 
-def _copy_btw_filesystems(input_opener, output_opener, BLOCK_SIZE=10_000_000):
+def _copy_btw_filesystems(input_opener, output_opener, BLOCK_SIZE=10_000_000, **kwargs):
+    streaming = True if kwargs.pop("block_size", True) == 0 else False
```
rabernat, line 29: Would prefer an explicit keyword argument rather than generic `**kwargs`.

rabernat: Also find it confusing to have both bases `BLOCK_SIZE` and `block_size` appearing in the arguments. Are these treated differently?

pangeo_forge_recipes/storage.py
```diff
@@ -25,11 +25,15 @@ def _get_url_size(fname, **open_kwargs):
...
+    if os.getenv("PANGEO_FORGE_BLOCK_SIZE"):
+        BLOCK_SIZE = int(os.getenv("PANGEO_FORGE_BLOCK_SIZE"))
```
rabernat, line 31: This needs to be documented in the documentation. Perhaps we need a page on environment variables?

Your PR description makes it sound like `PANGEO_FORGE_BLOCK_SIZE` only governs logging behavior, but here is seems like it changes the actual block size for reads.

pangeo_forge_recipes/storage.py
```diff
@@ -25,11 +25,15 @@ def _get_url_size(fname, **open_kwargs):
...
     with input_opener as source:
         with output_opener as target:
+            count = summed_bytes = 0
             while True:
-                logger.debug("_copy_btw_filesystems reading data")
+                logger.debug("_copy_btw_filesystems reading data") if not streaming else None
```
rabernat, line 36: I would rather have one single if / else block for the `streaming` parameter, so maybe we can just remove this line completely?

pangeo_forge_recipes/storage.py
```diff
@@ -39,7 +43,14 @@ def _copy_btw_filesystems(input_opener, output_opener, BLOCK_SIZE=10_000_000):
                     ) from e
                 if not data:
                     break
-                logger.debug(f"_copy_btw_filesystems copying block of {len(data)} bytes")
+                # strided logging pattern for streaming transfers to avoid excessive logs
+                if streaming:
+                    summed_bytes += len(data)
+                    if summed_bytes // BLOCK_SIZE >= count:
+                        logger.debug(f"copying block {count} of ~{BLOCK_SIZE} bytes")
```
rabernat, line 50: Would be useful if we could say `copying block {count}/{total}`

But more fundamentally, I don't see how this logic actually reduces the frequency of output. Isn't `summed_bytes // BLOCK_SIZE >= count` always true, since we read data in in `BLOCK_SIZE` increments?

rabernat (commented):
Thanks for working on this. I can see why it is important for large files.

I find myself a bit confused about the meaning of the different parameters. I don't understand what the `streaming` flag means. I guess it's related to logging behavior, but the variable name sounds like something more fundamental about how i/o works.

It seems like we should have _separate_ parameters that control:
- the actual block size of reads (this is what `BLOCK_SIZE` currently does). This could be tuned to give better i/o performance in different scenarios (e.g. file vs network)
- the frequency of the logging output. That should be clearly separate from the former. For, example, we might want `BLOCK_SIZE=10` but `LOG_TRANSFER_BLOG_SIZE=100` or something like that. Currently, I can't distinguish the two.

Further, I can't see how the code actually reduces the frequency of log output, but  I'm probably missing something obvious.

pangeo_forge_recipes/storage.py
cisaacstern: > Would prefer an explicit keyword argument rather than generic **kwargs.

Makes sense; that's an easy fix.

> Also find it confusing to have both bases BLOCK_SIZE and block_size appearing in the arguments. Are these treated differently?

Yeah, agree that's not ideal. Here, `BLOCK_SIZE` is the actual block size to read (as it's always been), and `block_size` is the keyword arg passed to `fsspec_open_kwargs` to trigger streaming reads.

pangeo_forge_recipes/storage.py
cisaacstern: > This needs to be documented in the documentation. Perhaps we need a page on environment variables?

👍 

> Your PR description makes it sound like PANGEO_FORGE_BLOCK_SIZE only governs logging behavior, but here is seems like it changes the actual block size for reads.

In the case of range requests, yes, it would theoretically do that, though I'd need to test it more in that context to confirm. But in the case of streaming downloads, my understanding is `fsspec_open_kwargs = {"block_size": 0}` actually forces
```python
data = source.read(BLOCK_SIZE)
```
to ignore `BLOCK_SIZE` in favor of many smaller streaming reads. So for this PR, the notion I'm trying to capture in the env variable is perhaps closer to `PANGEO_FORGE_LOGGING_INTERVAL_FOR_STREAMING_READS`, but that seemed like an oddly specific concept to convey in an env variable (even if it's name were abbreviated), so I ended up here.

In this form, it does have the potential to change the block size for range request reads, but as I've imagined it thus far, that use case is not intended. (If we keep some form of this implementation, that could be noted in the docs.) As for how it actually works for changing the logging interval, I'll address that in the answer to your question below.

pangeo_forge_recipes/storage.py
cisaacstern: > Would be useful if we could say copying block {count}/{total}

Yes this would be ideal, but in the case of streaming reads, we may not know the total size of the source file. This is the case for the motivating recipe https://github.com/pangeo-forge/staged-recipes/pull/56, where I've switched to streaming reads not because the HTTP server doesn't support range requests, but rather because the Content Length headers are incorrect, so range request reads terminate at the wrong place in the file.

> Isn't summed_bytes // BLOCK_SIZE >= count always true, since we read data in in BLOCK_SIZE increments?

As mentioned above, in the case of streaming reads, `data = source.read(BLOCK_SIZE)` actually ignores `BLOCK_SIZE` in favor of many smaller streaming reads. So perhaps this would be clearer as:
```python
if summed_bytes // LOGGING_INTERVAL >= count:
```
which is actually the behavior it exhibits when run.

cisaacstern:
Having worked through the inline comments, now thinking that changing the env variable to `PANGEO_FORGE_STREAMING_LOGS_INTERVAL` or similar, and thereby decoupling it from `BLOCK_SIZE`, could be a good next step? Would love to think of a more concise name for capturing this concept.

cisaacstern:
Moving this back to draft as I incorporate Ryan's suggestions.

rabernat:
Stepping back a bit, I'm wondering if there is an even simpler way to deal with this. Do we really need to introduce new config to handle this?

Could we just be using [log levels](https://docs.python.org/3/library/logging.html#levels) more carefully? 

Or, if the goal is to limit the _frequency_ of logging, should we target that directly, for example, by using the time module explicitly to decide when to log? For example, we could just say we will emit a logging message every 5 seconds, period.

rabernat:
```python
import time

start = time.time()
interval = 5 # seconds
while True:
    now = time.time() 
    if now - start > interval:
        print("log")
        start = now
```

Prints every 5 seconds. If we have time information and bytes read, we could use this to log _throughput_! 🎉 

```python
throughput = bytes_read / (now - start) # in bytes / second
```

cisaacstern:
> Prints every 5 seconds. If we have time information and bytes read, we could use this to log throughput! 🎉

This is a great solution, and not only should work regardless of the I/O circumstance, it should also result in an *improvement* to the quality of logging information for default range request I/O. I will rework this PR to reflect this idea.

cisaacstern:
@rabernat, this now reflects a working version of the timed logging you proposed in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/175#issuecomment-887729589.  

`timed_logging.ipynb` in https://github.com/cisaacstern/timed-logging-example shows how this looks for the user in a few different use cases, and is runnable via the included Binder link.

The one key use case missing from that notebook is streaming a large file over HTTP, i.e. the motivating scenario from https://github.com/pangeo-forge/staged-recipes/pull/56. For some reason, the NCAR Climate Data Gateway is down right now, so I couldn't include that one, and off the top of my head, couldn't think of another a large source file to stream over HTTP.

pangeo_forge_recipes/storage.py
```diff
@@ -39,8 +42,16 @@ def _copy_btw_filesystems(input_opener, output_opener, BLOCK_SIZE=10_000_000):
                     ) from e
                 if not data:
                     break
-                logger.debug(f"_copy_btw_filesystems copying block of {len(data)} bytes")
                 target.write(data)
+                bytes_read += len(data)
+                elapsed = time.time() - start
+                throughput = int(bytes_read / elapsed)
```
rabernat, line 48:
```suggestion
                throughput = bytes_read / elapsed
```

Why does it have to be an int?

pangeo_forge_recipes/storage.py
```diff
@@ -39,8 +42,16 @@ def _copy_btw_filesystems(input_opener, output_opener, BLOCK_SIZE=10_000_000):
...
+                if elapsed // interval >= log_count:
+                    logger.debug(f"_copy_btw_filesystems total bytes copied: {bytes_read}")
+                    logger.debug(
+                        f"avg throughput over {round(elapsed/60, 2)} min: ~{throughput} bytes/sec"
```
rabernat, line 52:
```suggestion
                        f"avg throughput over {elapsed/60:.2f)} min: {throughput/1_000_000:.2f} MB/sec"
```

Better to use string formatting rather than rounding to get the desired result.

rabernat (changes requested):
Just a few tiny nits.

pangeo_forge_recipes/storage.py
cisaacstern: It doesn't. When I was logging throughput as bytes/sec (as opposed to MB/sec), it seemed odd to include fractional bytes, and this was the simplest way I could think of to truncate the value. (In retrospect, string formatting might've been better.) Anyway, not needed either way anymore.

cisaacstern:
Changed `1_000_000` to `1e6` in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/175/commits/f09782b50ea911bd190ac01b0b7901dcdb7591ad to get us under the 100 character limit. If you prefer the former style, we could alternatively:

1. wrap the debug message to a second line (seems like a lot for saving a few characters) 
2. change the name of the `throughput` variable to something shorter, like `rate` or `io_rate`
3. shorten `min` and `sec` to `m` and `s`

But if `1e6` is okay, then this is good to merge IMO.

FWIW, just re-ran the example Binder using the latest commit, and I really like how the logging looks. `MB/sec` is much more intuitive to grasp than bytes/sec. Here's a bit of logging from the notebook, for NOAA-OISST:

```
pangeo_forge_recipes.storage - DEBUG - _copy_btw_filesystems total bytes copied: 1714749
pangeo_forge_recipes.storage - DEBUG - avg throughput over 0.01 min: 2.63 MB/sec
```
and for the large file example (HYCOM):
```
pangeo_forge_recipes.storage - DEBUG - _copy_btw_filesystems total bytes copied: 10000000
pangeo_forge_recipes.storage - DEBUG - avg throughput over 0.02 min: 10.43 MB/sec
pangeo_forge_recipes.storage - DEBUG - _copy_btw_filesystems total bytes copied: 50000000
pangeo_forge_recipes.storage - DEBUG - avg throughput over 0.09 min: 9.18 MB/sec
pangeo_forge_recipes.storage - DEBUG - _copy_btw_filesystems total bytes copied: 100000000
pangeo_forge_recipes.storage - DEBUG - avg throughput over 0.18 min: 9.44 MB/sec
```

cisaacstern:
Also, I'm a bit confused by GitHub's UI. I committed both of your suggestions, but it's still showing:

Screen Shot 2021-07-28 at 9 51 41 AM

rabernat:
> I committed both of your suggestions, but it's still showing:

GitHub is not smart enough to know if your commits actually resolved suggestions. 😆  You always need to manually change the PR approval status, which I will do now.

rabernat (approved):

PR #179, opened 2021-08-09
Replace `_copy_btw_filesystems` with `fsspec.cp`

cisaacstern:
The source file server for https://github.com/pangeo-forge/staged-recipes/pull/60 exhibits the same incompatibility with fsspec's FTP range request method described in https://github.com/intake/filesystem_spec/issues/641. Specifically, fsspec's attempt to halt an in-progress transfer with the [`.abort`](https://github.com/intake/filesystem_spec/blob/10164f780909a827e88b84f69b0e01b10f1d9316/fsspec/implementations/ftp.py#L310) method raises an error. This can be reproduced without `pangeo-forge-recipes`, as follows:

```python
import fsspec

fname = (
    "ftp://ftp.mercator-ocean.fr/download/users/cbricaud/ORCA36-T404/"
    "ORCA36-T404_1hAV_20140201-20140430_gridT_GulfStream.nc"
)
BLOCK_SIZE = 10_000_000
with fsspec.open(fname, mode="rb") as source:
    while True:
        data = source.read(BLOCK_SIZE)
        if not data:
            break
```
<details>
    <summary> Traceback </summary>

```
TransferDone                              Traceback (most recent call last)
~/.pyenv/versions/anaconda3-2019.10/envs/pangeo-forge3.8/lib/python3.8/site-packages/fsspec/implementations/ftp.py in _fetch_range(self, start, end)
    292         try:
--> 293             self.fs.ftp.retrbinary(
    294                 "RETR %s" % self.path,

~/.pyenv/versions/anaconda3-2019.10/envs/pangeo-forge3.8/lib/python3.8/ftplib.py in retrbinary(self, cmd, callback, blocksize, rest)
    429                     break
--> 430                 callback(data)
    431             # shutdown ssl layer

~/.pyenv/versions/anaconda3-2019.10/envs/pangeo-forge3.8/lib/python3.8/site-packages/fsspec/implementations/ftp.py in callback(x)
    284                 if end < self.size:
--> 285                     raise TransferDone
    286             else:

TransferDone: 

During handling of the above exception, another exception occurred:

error_proto                               Traceback (most recent call last)
~/.pyenv/versions/anaconda3-2019.10/envs/pangeo-forge3.8/lib/python3.8/site-packages/fsspec/implementations/ftp.py in _fetch_range(self, start, end)
    301                 # stop transfer, we got enough bytes for this block
--> 302                 self.fs.ftp.abort()
    303                 self.fs.ftp.getmultiline()

~/.pyenv/versions/anaconda3-2019.10/envs/pangeo-forge3.8/lib/python3.8/ftplib.py in abort(self)
    268         if resp[:3] not in {'426', '225', '226'}:
--> 269             raise error_proto(resp)
    270         return resp

error_proto: 451 Transfer aborted. Broken pipe

During handling of the above exception, another exception occurred:

AttributeError                            Traceback (most recent call last)
<ipython-input-1-12f123af293b> in <module>
      8 with fsspec.open(fname, mode="rb") as source:
      9     while True:
---> 10         data = source.read(BLOCK_SIZE)
     11         if not data:
     12             break

~/.pyenv/versions/anaconda3-2019.10/envs/pangeo-forge3.8/lib/python3.8/site-packages/fsspec/spec.py in read(self, length)
   1481             # don't even bother calling fetch
   1482             return b""
-> 1483         out = self.cache._fetch(self.loc, self.loc + length)
   1484         self.loc += len(out)
   1485         return out

~/.pyenv/versions/anaconda3-2019.10/envs/pangeo-forge3.8/lib/python3.8/site-packages/fsspec/caching.py in _fetch(self, start, end)
    151             part = b""
    152         end = min(self.size, end + self.blocksize)
--> 153         self.cache = self.fetcher(start, end)  # new block replaces old
    154         self.start = start
    155         self.end = self.start + len(self.cache)

~/.pyenv/versions/anaconda3-2019.10/envs/pangeo-forge3.8/lib/python3.8/site-packages/fsspec/implementations/ftp.py in _fetch_range(self, start, end)
    303                 self.fs.ftp.getmultiline()
    304             except Error:
--> 305                 self.fs.ftp._connect()
    306 
    307         return b"".join(out)

AttributeError: 'FTP' object has no attribute '_connect'
```

</details>

> **Note 1:** the last exception here is actually an fsspec `AttributeError`, the solution to which @martindurant previously identified in https://github.com/intake/filesystem_spec/issues/641#issuecomment-842747716. This is corrected by https://github.com/intake/filesystem_spec/pull/722.)

> **Note 2:** One workaround for this incompatibility is to read the source file in a single block. Many of the source files for the above-linked recipe are too large for memory, however, so this is not a viable solution.

The current PR refactors `_copy_btw_filesystems` to allow an alternate pathway for FTP transfers which accesses the low-level methods of Python's `ftplib` directly. By wrapping both this alternate pathway and the default pathway with the same `_write` decorator, writing and logging behaviors are identical regardless of which path is followed.

Of course, the alternative to merging this PR would be to implement a version of this approach upstream in `fsspec`. Martin, if that's something you're interested in exploring, let's discuss further. As you'll see here, I've applied the internal logic of `ftplib.FTP.retrbinary` to make this work, with the key point being that both reading *and writing* need to happen within the same (uninterrupted) [connection context](https://github.com/pangeo-forge/pangeo-forge-recipes/compare/master...cisaacstern:call-ftplib-directly?expand=1#diff-1b0c20425676e0f245bdd49a8db3355e0e45139497bf42748276b7fd8705e5b1R60-R66). 

> **Note 3:** While an upstream feature addition is always preferable, I question whether it's possible in this specific situation, given that a fundamental assumption of the `source.read(BLOCK_SIZE)` design pattern is that a transfer can be interrupted. In this scenario, the FTP server does not permit that.  `ftplib.FTP.retrbinary` gets around that by allowing the user to provide a single-parameter function to be called on each block of data as it is read. This is insufficient for our current situation, because uniform logging behavior here requires a wrapper around reading, parts of which are called before the read, and parts after. (This is why I've gone with even lower level `ftplib.FTP.transfercmd` instead of `ftplib.FTP.retrbinary`.) Noting these concerns in hopes there is some way to contribute upstream, but somewhat inclined to think our implementation here will just have to be a one-off downstream solution.

### DIFF
--- a/ci/py3.7.yml
+++ b/ci/py3.7.yml
@@ -11,7 +11,6 @@ dependencies:
   - dask
   - distributed
   # - fsspec>=2021.6.0
-  - git+https://github.com/cisaacstern/filesystem_spec@cp_file_x
   - h5netcdf
   - h5py
   - hdf5
@@ -36,3 +35,4 @@ dependencies:
   - zarr>=2.6.0
   - pip:
     - pytest-timeout
+    - git+https://github.com/cisaacstern/filesystem_spec@cp_file_x
--- a/ci/py3.7.yml
+++ b/ci/py3.7.yml
@@ -10,7 +10,8 @@ dependencies:
   - codecov
   - dask
   - distributed
-  - fsspec>=2021.6.0
+  # - fsspec>=2021.6.0
+  - git+https://github.com/cisaacstern/filesystem_spec@cp_file_x
   - h5netcdf
   - h5py
   - hdf5
--- a/ci/py3.8.yml
+++ b/ci/py3.8.yml
@@ -11,7 +11,6 @@ dependencies:
   - dask
   - distributed
   # - fsspec>=2021.6.0
-  - git+https://github.com/cisaacstern/filesystem_spec@cp_file_x
   - h5netcdf
   - h5py
   - hdf5
@@ -38,3 +37,4 @@ dependencies:
   - pip
   - pip:
     - pytest-timeout
+    - git+https://github.com/cisaacstern/filesystem_spec@cp_file_x
--- a/ci/py3.8.yml
+++ b/ci/py3.8.yml
@@ -10,7 +10,8 @@ dependencies:
   - codecov
   - dask
   - distributed
-  - fsspec>=2021.6.0
+  # - fsspec>=2021.6.0
+  - git+https://github.com/cisaacstern/filesystem_spec@cp_file_x
   - h5netcdf
   - h5py
   - hdf5
--- a/ci/py3.9.yml
+++ b/ci/py3.9.yml
@@ -10,7 +10,8 @@ dependencies:
   - codecov
   - dask
   - distributed
-  - fsspec>=2021.6.0
+  # - fsspec>=2021.6.0
+  - git+https://github.com/cisaacstern/filesystem_spec@cp_file_x
   - h5netcdf
   - h5py
   - hdf5
--- a/ci/py3.9.yml
+++ b/ci/py3.9.yml
@@ -11,7 +11,6 @@ dependencies:
   - dask
   - distributed
   # - fsspec>=2021.6.0
-  - git+https://github.com/cisaacstern/filesystem_spec@cp_file_x
   - h5netcdf
   - h5py
   - hdf5
@@ -37,3 +36,4 @@ dependencies:
   - zarr>=2.6.0
   - pip:
     - pytest-timeout
+    - git+https://github.com/cisaacstern/filesystem_spec@cp_file_x
--- a/docs/development/release_notes.md
+++ b/docs/development/release_notes.md
@@ -1,6 +1,9 @@
 # Release Notes
 
-## 0.5.0 - unreleased
+## v0.5.1 - Unreleased
+
+
+## v0.5.0 - 2021-07-11
 
 - Added `subset_inputs` option to `XarrayZarrRecipe`. {issue}`93`, {pull}`166`
 - Fixed file opening to eliminate HDF errors related to closed files. {issue}`170`, {pull}`171`
--- a/pangeo_forge_recipes/recipes/xarray_zarr.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr.py
@@ -125,6 +125,7 @@ def cache_input_metadata(
     delete_input_encoding: bool,
     process_input: Optional[Callable[[xr.Dataset, str], xr.Dataset]],
     is_opendap: bool,
+    call_ftplib_directly: bool,
 ) -> None:
     if metadata_cache is None:
         raise ValueError("metadata_cache is not set.")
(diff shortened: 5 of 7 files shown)


rabernat:
I see a lot of work has gone into this PR. Thanks so much @cisaacstern for digging into it so deeply.

> the alternative to merging this PR would be to implement a version of this approach upstream in `fsspec`

I must admit that I favor this path rather strongly. The code in this PR introduces a lot of complexity to work around the fsspec issue. That is technical debt for our project. I'm keen to hear from Martin whether this would be viable to implement in fsspec itself.

martindurant:
It occurred to me that we can use the "callback" mechanism recently added in fsspec to do this. The method would be `get`, and the FTP module would define a default callback which writes each chunk to an open local file (the normal process), but the callback could be overridden to write elsewhere (or do something else with the data).  I think this might be relatively easy to implement. 

We are still a fair way from a class for arbitrary cross-filesystem operations, but perhaps `cp` is the one and only such method we want, there is an argument to introduce a `cp_file_x` method in the spec meant to do this and a top level function `copy`/`cp` which takes two arbitrary URLs or lists of URLs. The default implementation would be what you had before
```
def cp_file_x(self, filea, fileb: OpenFile, blocksize, callback=None):
    with self.open(filea) as a:
        with fileb as b:
             while True:
                 data = a.read(chunksize)
                 if not data:
                      break
                 b.write(data)
```

cisaacstern:
I hadn't known of fsspec callbacks; they do indeed seem like the missing link for migrating out `_copy_btw_filesystems` to a new `fsspec.copy` function, without loosing our desired logging pattern.

I'd love to take a crack at the `cp_file_x` + top level `copy`/`cp` implementation. I'll aim to have a draft PR for discussion by sometime today or tomorrow.

This could result in a dramatic simplification of our `storage.py`. Very exciting. Thank you both for the feedback.

martindurant:
> they do indeed seem like the missing link

The tricky thing will be, that for FTP, alone, the callback will handle the writing, whereas for other implementations, callbacks just update some value, e.g., for progress bars. Probably there will need to be some special-casing at least somewhere.

> I'd love to take a crack at the cp_file_x + top level copy/cp implementation

Certainly! I have some company stuff this week, I wasn't planning to work on it.

rabernat (approved):
This looks great to me!

I would recommend removing the `_copy_btw_filesystems` function completely if its no longer used.

Let's wait until fsspec is released with the needed functionality before we merge this.

cisaacstern:
> Let's wait until fsspec is released with the needed functionality before we merge this.

@martindurant, Ryan's response reminds me to ask if you've gotten a chance to look at https://github.com/intake/filesystem_spec/pull/723#issuecomment-899829355?

cisaacstern:
Per https://github.com/pangeo-forge/pangeo-forge-recipes/pull/247#issuecomment-981893091, @martindurant will be working on a generic copy function for `fsspec` in the near term. We can reopen a new PR for implementing that when it's available but we won't need these commits so closing this.

PR #192, opened 2021-08-28
Go back to rechunker Pipeline executors

rabernat:
As discussed at our last dev meeting, we may want to try going back to using the rechunker Pipeline executors internally.

This PR tries that, plus moves towards a more flexible execution model. Rather than having to define the four specific stages, Recipes just have to define a private method `_to_pipelines()`. That way they can customize their own stages.

This seems to work (tests pass). The big question is whether the Dask and Prefect graphs are still as efficient as we need them to be. My hope is yes, because the underlying functions in the pipelines are still based on the more efficient `partial` functions that @TomAugspurger defined in #160.


pangeo_forge_recipes/recipes/reference_hdf_zarr.py
```diff
@@ -76,6 +77,15 @@ class HDFReferenceRecipe(BaseRecipe, FilePatternRecipeMixin):
...
     xarray_concat_args: dict = field(default_factory=dict)
 
+    def _to_pipelines(self) -> ParallelPipelines:
+        """Translate recipe to pipeline for execution.
+        """
+        pipeline = []  # type: MultiStagePipeline
+        pipeline.append(Stage(self.store_chunk, list(self.iter_chunks())))
+        pipeline.append(Stage(self.finalize_target))
+        pipelines = [pipeline]  # type: ParallelPipelines
+        return pipelines
```
rabernat, line 87: Here note that the compilation of the flow to pipelines is now implemented by each recipe class, not BaseRecipe. This gives recipes a lot more flexibility.

rabernat:
If this branch works, it will help things a lot going forward. We need an answer.

I would like to find a way to test this at scale with the workflows that were previously running out of memory. Questions for @sharkinsspatial:

- Is there a way to run a flow at scale from a test bakery against a development branch?
- If so, could you try to run the memory-crashing flows from a while back on this branch?  (https://github.com/rabernat/pangeo-forge/tree/refactor-executors)

sharkinsspatial:
@rabernat I can have a go at this.  I already have a rudimentary version of https://github.com/pangeo-forge/pangeo-forge-prefect/issues/25 which I was using for testing purposes which will handle this.  Do you have a specific recipe which would be the best candidate for testing?

rabernat:
> Do you have a specific recipe which would be the best candidate for testing?

Whichever recipe was causing things to crash before. I guess the noaa-oisst one?

sharkinsspatial:
@rabernat I attempted to run `noaa-oisst` recipe using the branch and encountered an OOM termination of the scheduler before completion.  I suspect based on my quick research noted here https://github.com/pangeo-forge/gpm-imerge-hhr-feedstock/issues/1 that this is a Prefect issue rather than a serialization issue with `pangeo-forge-recipes`.  

As another test, I attempted to run https://github.com/pangeo-forge/gpm-imerge-hhr-feedstock.  Running against `pangeo-forge-recipes==0.5.0` this recipe runs correctly (until, as noted it encounters memory issues).  However, when run using this branch (which requires moving the `fsspec_open_kwargs` from the recipe to the file pattern) I receive a `FileNotFoundError` for each `cache_inputs` task.  Reviewing this with @cisaacstern, he ran a quick local test https://gist.github.com/cisaacstern/d685d2453c613be60b363d44e22e000c demonstrating no issues with the recipe itself.  So it appears that there may be a regression with `to_pipelines` and `fsspec_open_kwargs`.

For next steps,
1. I'll run a test with `noaa-oisst` and the `DaskExecutor` agains a remote cluster to try and verify that serialization issues are Prefect specific with respect to this branch.
2. I'll run a test of `gpm-imerge-hhr` which requires a file pattern using the `DaskExecutor` so that you can have reproducible example of the `FileNotFoundError` which does not require Prefect infrastructure.

Let me know if there might be any other considerations here.

cisaacstern:
@sharkinsspatial, if the error is `FileNotFoundError`, then I think the issue is probably not with `fsspec_open_kwargs`? Because if the issue was `fsspec_open_kwargs` authentication, then the server should return `401 - Unauthorized`, not `FileNotFoundError`?

rabernat:
> I receive a `FileNotFoundError` for each `cache_inputs` task.

Do you have DEBUG logs from the failed run we can log at? Just as a general point, exposing these logs for public view seems like a really important priority for the bakeries.

cisaacstern:
> Just as a general point, exposing these logs for public view seems like a really important priority for the bakeries.

@rabernat, at our last Coordination Meeting, there was discussion of routing these logs to a "third party logging service" to get them out from behind the Prefect credential wall, and make them more easily searchable. Does that still seem like a viable option and if so, is there a particular service or implementation from another project, which we could use as a model?

rabernat:
> is there a particular service or implementation from another project, which we could use as a model?

Based on [this post](https://www.quora.com/What-is-the-best-logging-system-for-a-SaaS-startup), one free path would be 

The path could be:
- Create a [loggly free account](https://www.loggly.com/plans-and-pricing/)
- Set up the [loggly syslog agent](https://documentation.solarwinds.com/en/success_center/loggly/content/getting-started/gsg-send_data.htm) inside the prefect docker image
- Watch the logs roll in!

sharkinsspatial:
@cisaacstern Your assumption was correct that this is authentication issue.  See the logs below

rabernat:
> this authentication issue does not occur with `pangeo-forge-recipes==0.5.0`

0.5.0 is over a month old. We are already on 0.6.0 (see [release notes](https://pangeo-forge.readthedocs.io/en/latest/development/release_notes.html)). It seems most likely that this regression does not have to do with this PR (which doesn't touch the related code) but rather is due to #167 (refactor of input authentication and secret handling).

Automatic building of bakery images (https://github.com/pangeo-forge/pangeo-forge-bakery-images/issues/7), coupled with bakery integration testing, might have helped catch this issue earlier.

How / where are the secrets getting passed to the flow?

Issue #182, opened 2021-08-16
Disambiguate naming of `store_chunk`/`iter_chunks` from `target_chunks`


cisaacstern:
`XarrayZarrRecipe` uses variations of the name `chunks` to refer to two different recipe attributes. `iter_chunks` is a generator which yields inputs for the `store_chunk` execution step. `target_chunks` is a dictionary defining zarr chunk sizes. It may be worth considering re-naming `iter_chunks` (and any/all associated recipe attributes) to make it clear that this method is not related to zarr chunks.

rabernat:
Possible alternate names

- fragments
- shards
- blocks
- segments 
- write fragments / blocks / shards / segments

cisaacstern:
> fragments
shards
blocks
segments

shards or blocks are appealing to me simply for being shorter, and between them, shards seems less connotative of bytes (at least to me). so maybe shards?

cisaacstern:
Would like to add an endorsement for following the `target_chunks` naming pattern (`<adjective>_<noun>`) in the renaming of `store_chunk`/`iter_chunks`. Naming in storage.py defines a `source`/`target` relationship, so perhaps `store_chunk` could become something like ... `write_source_shard` ? Which would make `iter_chunks` ... `iter_source_shards`.

cisaacstern:
For anyone looking to grok the somewhat hard-to-grasp notion of the "source chunks" (under discussion for renaming here), reading https://github.com/pangeo-forge/pangeo-forge-recipes/pull/205#issuecomment-923200879 was an "a-ha" moment for me.

cisaacstern:
This issue relates to the pre-beam-refactor `XarrayZarrRecipe` which we are no longer developing, so closing. As an aside, I think the current `StoreToZarr` naming does a good job of addressing this prior naming overload, by distinguishing between `target_chunks` (i.e. zarr chunks), and `fragments` (which are used to create zarr chunks, but are not 1:1 with zarr chunks).

PR #198, opened 2021-09-02
Optimize tests following #167 slowdown

### LINKED ISSUE #199: Tests slowed by > 2x following merge of #167
[Checks](https://github.com/pangeo-forge/pangeo-forge-recipes/actions/runs/1183164556) for #174 took < 15 mins:
Screen Shot 2021-09-02 at 1 49 33 PM

[Checks](https://github.com/pangeo-forge/pangeo-forge-recipes/actions/runs/1192494723) for #167 took > 30 mins:
Screen Shot 2021-09-02 at 1 53 01 PM

Output of [`pytest --durations=10 --durations-min=1.0`](https://docs.pytest.org/en/6.2.x/usage.html#profiling-test-execution-duration):
```
========================================= slowest 10 durations ==========================================
5.37s teardown tests/test_utils.py::test_calc_subsets
2.74s call     tests/test_recipes.py::test_chunks[prefect-netCDFtoZarr_recipe-target_chunks6-False-error_expectation6-1-subset_inputs1-netcdf_local_file_pattern_sequential_multi_variable-D]
2.67s setup    tests/test_recipes.py::test_recipe_caching_copying[manual-netCDFtoZarr_http_recipe-True-True-netcdf_http_paths0-netcdf_local_paths_sequential-D]
2.33s call     tests/test_recipes.py::test_chunks[prefect-netCDFtoZarr_recipe-target_chunks6-False-error_expectation6-1-subset_inputs0-netcdf_local_file_pattern_sequential_multi_variable-D]
2.28s setup    tests/test_fixtures.py::test_fixture_http_files[netcdf_http_paths1-netcdf_local_paths_sequential_multi_variable-D]
2.28s setup    tests/test_fixtures.py::test_fixture_http_files[netcdf_http_paths2-netcdf_local_paths_sequential_multi_variable-D]
2.25s call     tests/test_recipes.py::test_chunks[prefect-netCDFtoZarr_recipe-target_chunks5-False-error_expectation5-1-subset_inputs0-netcdf_local_file_pattern_sequential_multi_variable-D]
2.25s setup    tests/test_fixtures.py::test_fixture_http_files[netcdf_http_paths0-netcdf_local_paths_sequential_multi_variable-D]
2.23s call     tests/test_recipes.py::test_chunks[prefect-netCDFtoZarr_recipe-target_chunks5-False-error_expectation5-1-subset_inputs1-netcdf_local_file_pattern_sequential_multi_variable-D]
2.17s setup    tests/test_fixtures.py::test_fixture_http_files[netcdf_http_paths2-netcdf_local_paths_sequential-D]
```
Setup for `tests/test_fixtures.py::test_fixture_http_files` is [obviously time-consuming](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/e748febdb9a5111d43a9ba7cf837812614894f27/tests/conftest.py#L227), but there are only 8 total tests run from `test_fixtures.py` and they cumulatively take < 30 seconds to run:
```
$ pytest test_fixtures.py
Results (28.12s):
       8 passed
       8 skipped
```
The slow `tests/test_recipes.py::test_chunks` calls are of greater interest, because with current fixturing, that test gets called *a lot*, as shown by the output of [`python parametrization_counter.py details plot`](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/198/files#diff-0e7d92f16205081086bc83ea6d0b8d0611e07e7b201061d5f76b7daf42962d7eR10):
test_counts

cisaacstern:
To start I've just added a module `parametrization_counter.py` which returns diagnostic information on test runs.

More to follow. Closes #199 when complete.


cisaacstern:
https://github.com/pangeo-forge/pangeo-forge-recipes/issues/199#issuecomment-912153004 documents speed-ups in `test_storage.py::test_file_opener` after limiting input fixtures to sequential file paths only. Picking up that thread here with some insights on `test_recipes.py::test_recipe_caching_copying` following the same change.

From our current `master`
```
$ pytest test_recipes.py -k test_recipe_caching_copying
Results (252.91s):
     320 passed
     835 deselected
```
And from https://github.com/pangeo-forge/pangeo-forge-recipes/pull/198/commits/076ed1645924865432730e63ea66665c197c9475
```
$ pytest test_recipes.py -k test_recipe_caching_copying
Results (103.09s):
     160 passed
     835 deselected
```

cisaacstern:
@rabernat, this PR shaves 15 mins off **3.8-build** and 10 mins of the **3.9-build** compared to `master` (where each is ~30min):

Screen Shot 2021-09-02 at 7 07 22 PM


We get there by refactoring the fixtures so we can use only sequential files/paths/patterns for the tests of moving files around, thereby cutting out a few hundred unnecessary tests. Here are the numbers of calls in each module, by PR:

| module | #174 |  #167 | #198 
|---|---|---|---|
| test_chunk_grid | 6 | 8 | 8
| test_fixtures| 6 | 16| 22
| test_locking| 6 | 6| 6
| test_patterns| 6 | 22| 22
| test_recipes|803 | 1155| 995 
| test_references| 8 | 8| 8
| test_storage| 99 | 387| 195
| test_utils| 2 | 2| 2

And a visual look at test calls from the current PR (xref plot of test calls for `master` in https://github.com/pangeo-forge/pangeo-forge-recipes/issues/199#issue-987172655):

optimize-tests



`test_recipes.py::test_chunks` is still the major outlier, but there doesn't seem to be a way to reduce those calls by changing the recipe fixture, since it only uses local paths at it is, and it doesn't seem appropriate to deprive that test of multivariable files.


This seems to be good enough progress for now, without making this into a multi-day exercise, so I'm good to merge if you are. Also happy to dig a little deeper, though, if you see any other obvious places for improvement.

If we merge, do you want to commit [`parametrization_counter.py`](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/198/files?file-filters%5B%5D=.py#diff-0e7d92f16205081086bc83ea6d0b8d0611e07e7b201061d5f76b7daf42962d7e), which I used to make those plots? I'd probably use it if it was there, but if not I'll drop in in a gist. (It wasn't clear to me if any pytest or third-party options already existed for making visual diagnostics like that.)

cisaacstern:
On further reflection, I thought of a way to possibly cut the file transfer testing time in half again. I'll push that update later tonight or first thing tomorrow.

cisaacstern:
We're now basically back to where we were before #167 in terms of testing time and IMHO this is now ready to merge:

Screen Shot 2021-09-03 at 1 17 26 AM

Previously when I mentioned testing only the "sequential" files, both the `"D"` and `"2D"` interval files were included in that. Considering each http run requires three auth parameterizations (no auth, basic auth, query string auth), even that little bit of redundancy was adding up.

In the last few commits, I refactored the fixtures again to make this `items_per_file` option explicit in the local paths, thereby exposing just the `"D"` interval files to make our http path/pattern/recipe fixtures. I think I'm convinced that we don't need to test any other files aside from these over http, because the purpose of the http tests is really to ensure file opening/caching/authentication/etc. works, and we actually don't care about the content of the files in those tests (only that they get where they're going intact).

Similarly, I did change `test_storage.py::test_file_opener` so that it only runs with the local sequential files. This is a marginal speed increase over including all of the local files, but thought I'd go for it while I was at it. Again, IIUC, for file opening/transfer/etc. tests, diversity of internal contents in the files being opened and transferred shouldn't matter. (At least, for files of roughly the same size and format.)

> do you want to commit parametrization_counter.py

Moved this out of the PR to https://github.com/cisaacstern/pytest-param-counter. And just because I like that plot, here's were we stand:

optimize-chunks-fewer-http-tests

cisaacstern:
Noting also that this also turns on duration profiling for github pytest runs (per Ryan's suggestion) and that all 10 of the slowest tests are now some version of `tests/test_recipes.py::test_chunks[prefect...`, which was not the case prior to the work in this PR, as documented in https://github.com/pangeo-forge/pangeo-forge-recipes/issues/199#issue-987172655. Working on this is out-of-scope for this PR, IMO, but good to keep in mind for further optimization down the line:

```
============================= slowest 10 durations =============================
3.30s call     tests/test_recipes.py::test_chunks[prefect-netCDFtoZarr_recipe-target_chunks6-False-error_expectation6-1-subset_inputs1-netcdf_local_file_pattern_sequential_multivariable-netcdf_local_paths_sequential_multivariable_1d]
3.17s call     tests/test_recipes.py::test_chunks[prefect-netCDFtoZarr_recipe-target_chunks5-False-error_expectation5-1-subset_inputs1-netcdf_local_file_pattern_sequential_multivariable-netcdf_local_paths_sequential_multivariable_1d]
3.17s call     tests/test_recipes.py::test_chunks[prefect-netCDFtoZarr_recipe-target_chunks6-False-error_expectation6-1-subset_inputs0-netcdf_local_file_pattern_sequential_multivariable-netcdf_local_paths_sequential_multivariable_1d]
3.14s call     tests/test_recipes.py::test_chunks[prefect-netCDFtoZarr_recipe-target_chunks5-False-error_expectation5-1-subset_inputs0-netcdf_local_file_pattern_sequential_multivariable-netcdf_local_paths_sequential_multivariable_1d]
2.97s call     tests/test_recipes.py::test_chunks[prefect-netCDFtoZarr_recipe-target_chunks2-True-error_expectation2-1-subset_inputs1-netcdf_local_file_pattern_sequential_multivariable-netcdf_local_paths_sequential_multivariable_1d]
2.94s call     tests/test_recipes.py::test_chunks[prefect-netCDFtoZarr_recipe-target_chunks1-True-error_expectation1-1-subset_inputs0-netcdf_local_file_pattern_sequential_multivariable-netcdf_local_paths_sequential_multivariable_1d]
2.90s call     tests/test_recipes.py::test_chunks[prefect-netCDFtoZarr_recipe-target_chunks1-True-error_expectation1-1-subset_inputs1-netcdf_local_file_pattern_sequential_multivariable-netcdf_local_paths_sequential_multivariable_1d]
2.90s call     tests/test_recipes.py::test_chunks[prefect-dask-netCDFtoZarr_recipe-target_chunks5-False-error_expectation5-1-subset_inputs1-netcdf_local_file_pattern_sequential_multivariable-netcdf_local_paths_sequential_multivariable_1d]
2.88s call     tests/test_recipes.py::test_chunks[prefect-netCDFtoZarr_recipe-target_chunks0-True-error_expectation0-1-subset_inputs1-netcdf_local_file_pattern_sequential_multivariable-netcdf_local_paths_sequential_multivariable_1d]
2.87s call     tests/test_recipes.py::test_chunks[prefect-netCDFtoZarr_recipe-target_chunks2-True-error_expectation2-1-subset_inputs0-netcdf_local_file_pattern_sequential_multivariable-netcdf_local_paths_sequential_multivariable_1d]
========== 1041 passed, 70 skipped, 465 warnings in 696.19s (0:11:36) ==========
```

cisaacstern:
@rabernat, is there a reason why we'd want each one of these param combinations of `test_chunks` https://github.com/pangeo-forge/pangeo-forge-recipes/blob/e748febdb9a5111d43a9ba7cf837812614894f27/tests/test_recipes.py#L188-L205

to be run with each of the 4 executors

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/e748febdb9a5111d43a9ba7cf837812614894f27/tests/conftest.py#L331-L332

I believe chunking/subsetting behavior should be independent of executor? If we run `test_chunks` with just a single executor (`manual`, maybe, or `python`), I think we'd see dramatic speed ups in the test suite.

rabernat:
This `test_chunks` test actually does two important things:
- It tests chunking options in XarrayZarrRecipe, verifying that the various options and combinations of parameters work as expected.
- Since some of these options involve different `store_chunk` tasks writing to the same zarr chunks (again, ambiguous terminology collision 💥 ; #182), it verifies that our distributed locking mechanisms work properly (and indeed, these tests write corrupted data when locking doesn't work, so that is very useful)

But it is an expensive test, and I agree that we should eliminate some parameterization. I recommend three steps.
- Refactor the `execute_recipe` fixture to be four separate fixtures (allowing us to apply executors more selectively in the tests, rather than always using all of them), and then make combinations of them using `lazy_fixture`
- Make `test_chunks` only execute with one executor (probably python, the simplest)
- Create a new test called `test_chunks_distributed_locking` that is explicitly aimed at verifying the locking. This doesn't need parameterization; just use the option set `({"lon": 12, "time": 3}, False, does_not_raise())`. (This produces write conflicts because the task chunks [either size 1 or 2] are smaller than the Zarr chunks [3])

cisaacstern:
I believe all the steps in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/198#issuecomment-912695964 are now complete.

We're now at the point where almost half the test time is setting up miniconda. Will look into https://github.com/marketplace/actions/setup-miniconda#caching now to reduce that.

rabernat (approved):
Awesome work Charles!

cisaacstern:
To close the loop, here's where we landed with this PR, in terms of calls per test (plot below).

@rabernat,
https://github.com/pangeo-forge/pangeo-forge-recipes/blob/c4a4ed84027ee6d2c372d8f403a3c3f3109a13d1/tests/test_recipes.py#L189-L190
is now the outlier. Just checked locally and if we run it with the new `execute_recipe_python` fixture only, we can save more than a minute, going from

```
Results (87.57s):
     120 passed
```
to
```
Results (10.40s):
      24 passed
```
any reason `test_process` needs to be run with all the executors?

PR-198

cisaacstern:
and the same could be asked of 
https://github.com/pangeo-forge/pangeo-forge-recipes/blob/c4a4ed84027ee6d2c372d8f403a3c3f3109a13d1/tests/test_recipes.py#L136-L137

rabernat:
Very true. Our default approach for testing is basically to run every test with every recipe and every executor. This obviously creates a lot of tests. But the good thing is that we really verify that all features _work_ with all the different executors. We could make the test suite a lot slimmer by removing all this executor parameterization...if we are sure we can trust the executors.

I would feel more comfortable doing that if we formalized the interface with the executors more. The pipelines model (#192) allows us to do that. If all of the recipes export pipelines, and then we trust the executors to execute the pipelines faithfully, then we can remove this parameterization. This would require the pipelines framework to be very well tested. Ideally, we would put pipelines into its own package, which both pangeo-forge-recipes and rechunker would depend on.

We can do these things and they will help. But they are probably not your top priority right now.

cisaacstern:
Makes a lot of sense. Thanks for the detailed reflections. If the dependencies caching looks like low-hanging fruit after a bit more reading, I may throw that in today, given just how long the miniconda setup takes. But I'll leave the parameter tweaking here for now, as I fully agree the path forward is through #192, not fiddling with the existing setup.

PR #201, opened 2021-09-03
create FilePatternRecipeMixin

rabernat:
Now that we have two distinct Recipe classes, we can start refactoring to eliminate duplication between them.

This starts in that direction by creating a new mixin that all Recipes that use FilePatterns as an input can use. One benefit of this is that `HDFReferenceRecipe` will now get the `copy_pruned` method for free.


cisaacstern (approved):
Very cool, Ryan. And timely, as I just learned about the concept of mixins from @TomNicholas.

As an aside, from glancing at the docs build for this PR it seems that Sphinx is aware of inherited attributes, and so still lists file pattern in the XarrayZarrRecipe API, even though it's no longer defined directly in the subclass. This is expected, of course, but still cool. 😄

PR #202, opened 2021-09-03
implemented a preprocessing argument for reference_hdf_zarr.py

jbusecke:
@cisaacstern and I had a little hack and found the need to pipe a preprocessing function into `HDFReferenceRecipe`.

### DIFF
--- a/tests/recipe_tests/test_HDFReferenceRecipe.py
+++ b/tests/recipe_tests/test_HDFReferenceRecipe.py
@@ -73,3 +73,77 @@ def test_multi(netcdf_local_file_pattern_sequential, tmpdir, with_intake):
         )
         ds = xr.open_dataset(m, engine="zarr", chunks={}, consolidated=False)
     assert (ds.foo == expected.foo).all()
+
+
+def _drop_lon(ds):
+    return ds.drop_vars(["lon"])
+
+
+@pytest.mark.parametrize("process", [None, _drop_lon])
+@pytest.mark.parametrize("with_intake", [True, False])
+def test_process(netcdf_local_file_pattern_sequential, tmpdir, with_intake, process):
+    file_pattern = netcdf_local_file_pattern_sequential
+    paths = [f for _, f in list(file_pattern.items())]
+    expected = xr.open_mfdataset(paths, engine="h5netcdf")
+    if process:
+        expected = _drop_lon(expected)
+
+    recipe = HDFReferenceRecipe(file_pattern, process_input=process)
+
+    # make sure assigning storage later works
+    out_target = FSSpecTarget(fs=fsspec.filesystem("file"), root_path=str(tmpdir))
+    metadata_cache = MetadataTarget(fs=fsspec.filesystem("file"), root_path=tempfile.mkdtemp())
+    recipe.target = out_target
+    recipe.metadata_cache = metadata_cache
+
+    recipe.to_dask().compute(scheduler="sync")
+
+    assert out_target.exists("reference.json")
+    assert out_target.exists("reference.yaml")
+
+    if with_intake:
+        cat = intake.open_catalog(out_target._full_path("reference.yaml"))
+        ds = cat.data.read()
+    else:
+        m = fsspec.get_mapper(
+            "reference://",
+            fo=out_target._full_path("reference.json"),
+            target_protocol="file",
+            remote_protocol="file",
+            skip_instance_cache=True,
+        )
+        ds = xr.open_dataset(m, engine="zarr", chunks={}, consolidated=False)
+    assert (ds.foo == expected.foo).all()
+
+
+# # function passed to preprocessing
+# def incr_date(ds, filename=""):
+#     # add one day
+#     t = [d + int(24 * 3600e9) for d in ds.time.values]
+#     ds = ds.assign_coords(time=t)
+#     return ds
+
+
+# @pytest.mark.parametrize(
+#     "process_input, process_chunk",
+#     [(None, None), (incr_date, None), (None, incr_date), (incr_date, incr_date)],
+# )
+# @pytest.mark.parametrize("recipe_fixture", all_recipes)
+# def test_process(recipe_fixture, execute_recipe, process_input, process_chunk):
+#     """Check that the process_chunk and process_input arguments work as expected."""
+
+#     RecipeClass, file_pattern, kwargs, ds_expected, target = recipe_fixture
+#     kwargs["process_input"] = process_input
+#     kwargs["process_chunk"] = process_chunk
+#     rec = RecipeClass(file_pattern, **kwargs)
+#     execute_recipe(rec)
+#     ds_actual = xr.open_zarr(target.get_mapper()).load()
+
+#     if process_input and process_chunk:
+#         assert not ds_actual.identical(ds_expected)
+#         ds_expected = incr_date(incr_date(ds_expected))
+#     elif process_input or process_chunk:
+#         assert not ds_actual.identical(ds_expected)
+#         ds_expected = incr_date(ds_expected)
+
+#     xr.testing.assert_identical(ds_actual, ds_expected)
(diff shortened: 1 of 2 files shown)


martindurant (commented):
Yes, good to include this, and some examples outside of pangeo-forge have already needed to use this.

rabernat:
Keep in mind that this recipe doesn't actually copy any data variables, so your preprocessor function is limited to touching metadata and the coordinates that get written directly to the json file.

cisaacstern:
> limited to touching metadata and the coordinates that get written directly to the json file.

Yes, this was our use case: deleting conflicting variable encodings, as is done in XarrayZarrRecipe.

jbusecke:
> Keep in mind that this recipe doesn't actually copy any data variables, so your preprocessor function is limited to touching metadata and the coordinates that get written directly to the json file.

I can mention that in the docstring, or is there another way of restricting modifications to the metadata?

cisaacstern:
> I can mention that in the docstring, or is there another way of restricting modifications to the metadata?

Docstring mention makes sense. Functionality-wise, we don't need to do anything, as the metadata restriction is enforced by `fsspec-reference-maker`.

I wonder though, should we call the `HDFReferenceRecipe` kwarg `process_input_metadata`, rather than `process_input`, so users are more likely to be aware of this limitation?

jbusecke:
I think that is a good idea. Also does anyone have some input on how to structure a test for this?

martindurant:
> Also does anyone have some input on how to structure a test for this?

Dropping a coordinate seems the simplest thing to do

cisaacstern:
The test for this should go in [pangeo-forge-recipes/tests/recipe_tests/test_HDFReferenceRecipe.py](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/master/tests/recipe_tests/test_HDFReferenceRecipe.py)

This test of `XarrayZarrRecipe.process_input` and `XarrayZarrRecipe.process_chunk` may be a useful reference, but your test will obviously differ given that (1) you're only testing one method, not two; and (2) HDFReferenceRecipe outputs differ from XarrayZarrRecipe outputs in ways which will affect how you structure the test

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/8720b48bfe208319ab80135500192e23351eece9/tests/recipe_tests/test_XarrayZarrRecipe.py#L176-L206

Let me know if any further clarification would be helpful

jbusecke:
I implemented a simple test based on @martindurant  suggestion. Due to the fact that we actually can do more than changing the metadata (like dropping variables) I left the argument name at `process_input` and added a note in the docstring. Thoughts?

cisaacstern:
Something strange is happening here where https://github.com/pangeo-forge/pangeo-forge-recipes/pull/202/files is showing changes for this PR which were already merged as part of https://github.com/pangeo-forge/pangeo-forge-recipes/pull/201

jbusecke:
Ughhh this again. Sorry, this has happened multiple times to me now. Something went bananas, when I tried to rebase. Is there a way to force clean the git commits?

jbusecke:
Ok i rebased again and force pushed. It seems to have worked this time.

tests/recipe_tests/test_HDFReferenceRecipe.py
```diff
@@ -73,3 +73,44 @@ def test_multi(netcdf_local_file_pattern_sequential, tmpdir, with_intake):
...
+
+
+@pytest.mark.parametrize("process", [None, _drop_lon])
+@pytest.mark.parametrize("with_intake", [True, False])
+def test_process(netcdf_local_file_pattern_sequential, tmpdir, with_intake, process):
+    file_pattern = netcdf_local_file_pattern_sequential
+    paths = [f for _, f in list(file_pattern.items())]
+    expected = xr.open_mfdataset(paths, engine="h5netcdf")
+    if process:
+        expected = _drop_lon(expected)
```
cisaacstern, line 89:
```suggestion
        expected = process(expected)
```
This function is assigned to the name process in line 82, so you should be able to call it by this name.

cisaacstern (changes requested):
@jbusecke, I can't seem to get this test to fail. For example, when I clone this PR and change 

```diff
if process:
-    expected = process(expected)
+    expected = expected
```
We would expect
```
assert (ds.foo == expected.foo).all()
```
to fail, right? Because `expected` has ostensibly not been modified in the same manner as `ds` was by the recipe.

But that's not what I'm seeing: the test passes either way.

Are you able to make this test fail by changing something? Apologies if I'm missing something obvious here.

jbusecke:
Ah fantastic catch @cisaacstern. The assertion at the end was not strict enough (because I am dropping 'lon' and 'foo' does not have lon as a coordinate 🙈). Using `xr.testing.assert_equal(ds, expected)` works. I can confirm that the test fails if I comment 
```python
    if process:
        expected = process(expected)
```

tests/recipe_tests/test_HDFReferenceRecipe.py
```diff
@@ -73,3 +73,44 @@ def test_multi(netcdf_local_file_pattern_sequential, tmpdir, with_intake):
...
+    else:
+        m = fsspec.get_mapper(
+            "reference://",
+            fo=out_target._full_path("reference.json"),
+            target_protocol="file",
+            remote_protocol="file",
+            skip_instance_cache=True,
+        )
+        ds = xr.open_dataset(m, engine="zarr", chunks={}, consolidated=False)
+    xr.testing.assert_equal(ds, expected)
```
rabernat, line 116: Instead of creating a new test by copy-pasting the previous one, let's add the `@pytest.mark.parametrize("process", [None, _drop_lon])` decorator (and associated logic) to `test_multi`. This would be a lot more compact and would help keep our test suite running more quickly (fewer tests run total). I don't see what we gain by running this test with `process=None`.

rabernat (changes requested):
Thanks so much Julius and Charles for working on this important feature!

Just one request that will help slim down the testing code.

tests/recipe_tests/test_HDFReferenceRecipe.py
jbusecke: Sounds good to me!

tests/recipe_tests/test_HDFReferenceRecipe.py
```diff
@@ -39,15 +39,23 @@ def test_single(netcdf_local_file_pattern_sequential, tmpdir, with_intake):
         )
         m = fs.get_mapper("")
         ds = xr.open_dataset(m, engine="zarr", chunks={}, consolidated=False)
-    assert (ds.foo == expected.foo).all()
+    xr.testing.assert_equal(ds, expected)
```
rabernat, line 42: I changed this just to scratch an itch, and interestingly, it turned up a test failure.

```
>       xr.testing.assert_equal(ds, expected)
E       AssertionError: Left and right Dataset objects are not equal
E       Differing dimensions:
E           (time: 10, lat: 18, lon: 36) != (time: 2, lat: 18, lon: 36)
E       Differing coordinates:
E       L * time     (time) datetime64[us] 2010-01-01 2010-01-02 ... 2010-01-10
E       R * time     (time) datetime64[ns] 2010-01-01 2010-01-02
E       R   foo      (time, lat, lon) float64 ...
E       L   bar      (time, lat, lon) float64 9.0 4.0 3.0 2.0 2.0 ... 3.0 3.0 9.0 5.0
E       R   bar      (time, lat, lon) int64 9 4 3 2 2 8 0 0 4 8 ... 1 9 4 5 4 3 1 0 5 3
```

@martindurant - any idea what is going on here?

cisaacstern:
It looks like we didn't end up finishing this before the same thing was achieved in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/326:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/092e45078c1c36a22cbc320a78b2aa56d751149a/pangeo_forge_recipes/recipes/reference_hdf_zarr.py#L144

Closing this now.

Issue #209, opened 2021-09-22
Ensure dimension coordinates are not chunked?


TomAugspurger:
I'm not sure if this is specific to the [gridMET feedstock](https://github.com/pangeo-forge/gridmet-feedstock), but posting here for visibility anyway.

The variables in gridMET have dimensions `(time, lon, lat)`, and are chunked by `(time,)`. Currently, opening the Zarr store is quite slow: From outside Azure (requires adlfs to be installed)

```python
In [1]: import fsspec, xarray as xr

In [2]: store = fsspec.get_mapper("abfs://gridmet/gridmet.zarr", account_name="ai4edataeuwest")

In [3]: %time ds = xr.open_zarr(store, consolidated=True)
CPU times: user 6.9 s, sys: 624 ms, total: 7.52 s
Wall time: 2min 34s
```

I *think* this is because the array for one of the coordinates, `time`, is chunked:

```python
In [5]: g = zarr.open_consolidated(store)

In [6]: %time g = zarr.open_consolidated(store)
CPU times: user 15.6 ms, sys: 0 ns, total: 15.6 ms
Wall time: 263 ms

In [7]: print(g.time.info)
Name               : /time
Type               : zarr.core.Array
Data type          : int64
Shape              : (15341,)
Chunk shape        : (30,)
Order              : C
Read-only          : False
Compressor         : Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)
Store type         : zarr.storage.ConsolidatedMetadataStore
Chunk store type   : fsspec.mapping.FSMap
No. bytes          : 122728 (119.9K)
Chunks initialized : 512/512
```

So a few questions:

1. Does pangeo-forge want to ensure that coordinates arrays are not chunked? At least by default, I think not chunking makes sense, because xarray eagerly loads them into memory.
2. Is this an issue in pangeo-forge-recipes itself, or in how gridMET is using it?

cc @sharkinsspatial

rabernat:
There are two issues here:

1. If fsspec async magic is working properly, it should definitely not take 2 minutes to read 512 chunks. This suggests an issue with adlfs. I am nearly certain that s3fs and gcsfs are not nearly so slow on this. Does adlfs mapper implement `getitems`?
[...]

PR #210, opened 2021-09-22
Consolidate dimension coordinates

TomAugspurger:
This consolidates dimension coordinates to address the performance
issues from having many small coordinate chunks.

Closes https://github.com/pangeo-forge/pangeo-forge-recipes/issues/209

The tests will fail right now with

```python
/home/taugspurger/src/pangeo-forge/pangeo-forge-recipes/tests/recipe_tests/test_XarrayZarrRecipe.py:282: AssertionError: Left and right Dataset objects are not identical


>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
E   AssertionError: Left and right Dataset objects are not identical

    Differing coordinates:
    L * time     (time) datetime64[ns] NaT 2010-01-02 ... 2010-01-09 2010-01-10
    R * time     (time) datetime64[ns] 2010-01-01 2010-01-02 ... 2010-01-10
    Differing data variables:
    L   foo      (time, lat, lon) float64 0.417 0.7203 0.0001144 ... 0.1179 0.3748
        long_name: Fantastic Foo
    R   foo      (time, lat, lon) float64 0.417 0.7203 0.0001144 ... 0.1179 0.3748
        long_name: Fantastic Foo
    L   bar      (time, lat, lon) int64 9 4 3 2 2 8 0 0 4 8 ... 9 8 4 4 6 0 3 3 9 5
        long_name: Beautiful Bar
    R   bar      (time, lat, lon) int64 9 4 3 2 2 8 0 0 4 8 ... 9 8 4 4 6 0 3 3 9 5
        long_name: Beautiful Bar
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> entering PDB >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
```

it's a bit hard to see, but the `time` value changed. The value of `0` is being interpreted as a NaT by xarray now. I'm sure I'm missing something basic with metadata when rewriting the variable, but I haven't found it yet. Posting this before I figure that out in case someone knows it offhand.

### DIFF
--- a/pangeo_forge_recipes/recipes/xarray_zarr.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr.py
@@ -621,9 +621,25 @@ def store_chunk(
                 zarr_array[zarr_region] = data
 
 
-def finalize_target(target: CacheFSSpecTarget, consolidate_zarr: bool) -> None:
+def finalize_target(
+    target: CacheFSSpecTarget,
+    consolidate_zarr: bool,
+    consolidate_dimension_coordinates: bool = True,
+) -> None:
     if target is None:
         raise ValueError("target has not been set.")
+
+    if consolidate_dimension_coordinates:
+        logger.info("Consolidating dimension coordinate arrays")
+        target_mapper = target.get_mapper()
+        ds = xr.open_zarr(target_mapper)  # Probably a better way to get the dimension coords?
+        group = zarr.open(target_mapper)
+        for dim in ds.dims:
+            attrs = dict(group[dim].attrs)
+            data = group[dim][:]
+            group[dim] = data
+            group[dim].attrs.update(attrs)
+
     if consolidate_zarr:
         logger.info("Consolidating Zarr metadata")
         target_mapper = target.get_mapper()
@@ -661,6 +677,9 @@ class XarrayZarrRecipe(BaseRecipe, FilePatternRecipeMixin):
       ``xr.open_dataset``. This is required for engines that can't open
       file-like objects (e.g. pynio).
     :param consolidate_zarr: Whether to consolidate the resulting Zarr dataset.
+    :param consolidate_dimension_coordinates: Whether to rewrite coordinate variables as a
+        single chunk. We recommend consolidating coordinate variables to avoid
+        many small read requests to get the coordinates in xarray.
     :param xarray_open_kwargs: Extra options for opening the inputs with Xarray.
     :param xarray_concat_kwargs: Extra options to pass to Xarray when concatenating
       the inputs to form a chunk.
@@ -685,6 +704,7 @@ class XarrayZarrRecipe(BaseRecipe, FilePatternRecipeMixin):
     cache_inputs: Optional[bool] = None
     copy_input_to_local_file: bool = False
     consolidate_zarr: bool = True
+    consolidate_dimension_coordinates: bool = True
     xarray_open_kwargs: dict = field(default_factory=dict)
     xarray_concat_kwargs: dict = field(default_factory=dict)
     delete_input_encoding: bool = True
@@ -851,7 +871,10 @@ class XarrayZarrRecipe(BaseRecipe, FilePatternRecipeMixin):
     @property
     def finalize_target(self) -> Callable[[], None]:
         return functools.partial(
-            finalize_target, target=self.target, consolidate_zarr=self.consolidate_zarr
+            finalize_target,
+            target=self.target,
+            consolidate_zarr=self.consolidate_zarr,
+            consolidate_dimension_coordinates=self.consolidate_dimension_coordinates,
         )
 
     def iter_inputs(self) -> Iterator[InputKey]:
--- a/pangeo_forge_recipes/recipes/xarray_zarr.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr.py
@@ -3,6 +3,7 @@ A Pangeo Forge Recipe
 """
 
 import functools
+import itertools
 import logging
 import os
 import warnings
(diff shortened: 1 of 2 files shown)


pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -621,9 +621,25 @@ def store_chunk(
...
         raise ValueError("target has not been set.")
+
+    if consolidate_dimension_coordinates:
+        logger.info("Consolidating dimension coordinate arrays")
+        target_mapper = target.get_mapper()
+        ds = xr.open_zarr(target_mapper)  # Probably a better way to get the dimension coords?
+        group = zarr.open(target_mapper)
+        for dim in ds.dims:
+            attrs = dict(group[dim].attrs)
+            data = group[dim][:]
```
TomAugspurger, line 639: This assumes the coordinate fits in-memory on a single machine. I don't know if we're assuming that anywhere else (probably in the tests), but it's probably a safe assumption for now.

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -621,9 +621,25 @@ def store_chunk(
...
+    consolidate_zarr: bool,
+    consolidate_dimension_coordinates: bool = True,
+) -> None:
     if target is None:
         raise ValueError("target has not been set.")
+
+    if consolidate_dimension_coordinates:
+        logger.info("Consolidating dimension coordinate arrays")
+        target_mapper = target.get_mapper()
+        ds = xr.open_zarr(target_mapper)  # Probably a better way to get the dimension coords?
```
rabernat, line 635: You could make a set out of the [`_ARRAY_DIMENSIONS`](http://xarray.pydata.org/en/stable/internals/zarr-encoding-spec.html) attribute on each array in the group.

pangeo_forge_recipes/recipes/xarray_zarr.py
rabernat: I think it's a safe assumption, since the dimension coordinates are guaranteed to be 1D. We could note this caveat in the docs on `consolidate_dimension_coordinates`.

It also assumes that `dim in group`. But that's not always the case. Xarray Datasets can have dimensions with no corresponding coordinate, which would give a `KeyError` here.

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -621,9 +621,25 @@ def store_chunk(
...
+        group = zarr.open(target_mapper)
+        for dim in ds.dims:
+            attrs = dict(group[dim].attrs)
+            data = group[dim][:]
+            group[dim] = data
```
rabernat, line 640: AFAICT this will not actually change the chunking. Since the array already exists, this statement will stripe the data over the existing chunks.

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -621,9 +621,25 @@ def store_chunk(
...
+            group[dim].attrs.update(attrs)
```
rabernat, line 641: Why are you updating the attrs here? It doesn't look like they could have changed at all.

rabernat (commented):
I'm confused about how this implementation actually changes the chunks. To me it looks like you're just reading the data and writing it back to the same Zarr Array. But I am probably missing something.

rabernat:
> The value of `0` is being interpreted as a NaT by xarray now.

This may indicate that the encoding has changed. To debug, I would open the target data with `decode_times=False` and see what is in the `time` array. Then call `xr.decode_cf` and look at the `.encoding` attribute on `time`.

TomAugspurger:
> To me it looks like you're just reading the data and writing it back to the same Zarr Array

IIUC, the difference is that `group["key"][:] = value` writes with the same structure (chunks, metadata, etc.), while `group["key"] ` value` completely overwrites the old array, so it gets new chunks.

```python
In [3]: group = zarr.group()

In [4]: group["a"] = zarr.ones(10, chunks=(2,))

In [5]: data = group["a"][:]

In [6]: group["a"][:] = data

In [7]: group["a"].chunks
Out[7]: (2,)

In [8]: group["a"] = data

In [9]: group["a"].chunks
Out[9]: (10,)
```

Which should answer https://github.com/pangeo-forge/pangeo-forge-recipes/pull/210#discussion_r714387952 and https://github.com/pangeo-forge/pangeo-forge-recipes/pull/210#discussion_r714388132.

However, writing an array with attrs to a group like `group[key] = array_with_attrs` doesn't actually result in attrs. I haven't looked to see if this is a bug or not.

b2fcf0dd4531d47f22d068a4dadee1776894248c has a fix that passes tests by using `group.array(...)`.

pangeo_forge_recipes/recipes/xarray_zarr.py
TomAugspurger: Thanks. Done in 1c36e2155c6e86ae633930d541a727713b62efb6.

tests/recipe_tests/test_XarrayZarrRecipe.py
```diff
@@ -340,6 +345,19 @@ def test_chunks_distributed_locking(
     )
 
 
+def test_no_consolidate_dimension_coordinates(netCDFtoZarr_recipe):
+    RecipeClass, file_pattern, kwargs, ds_expected, target = netCDFtoZarr_recipe
+
+    rec = RecipeClass(file_pattern, **kwargs)
+    rec.consolidate_dimension_coordinates = False
```
rabernat, line 352: Perfect...my one last comment was going to be that we needed a test for both options, but that's already here.

PR #213, opened 2021-09-27
Tweak file opening

rabernat:
This PR is in preparation for making XarrayZarrRecipe use fsspec-references for opening NetCDF files.

I have moved some of the reference stuff into its own module `reference.py` and call this from `reference_hdf_zarr.py`.

I have also changed `reference_hdf_zarr` to use the `storage.file_opener` function. This revealed some inconsistencies with return types of that function. Specifically, `file_opener` was previously yielding and `_io.BufferedReader` object in some cases and an `fsspec.OpenFile` object in others (specifically, when opening from cache). Now it will _always_ yield `fsspec.OpenFile`s.

### DIFF
--- a/pangeo_forge_recipes/reference.py
+++ b/pangeo_forge_recipes/reference.py
@@ -1,0 +1,26 @@
+"""
+Functions related to creating fsspec references.
+"""
+
+import fsspec
+from fsspec_reference_maker.hdf import SingleHdf5ToZarr
+
+
+def create_hdf5_reference(fname: str, inline_threshold: int = 300, **netcdf_storage_options):
+    with fsspec.open(fname, **netcdf_storage_options) as f:
+        h5chunks = SingleHdf5ToZarr(f, _unstrip_protocol(fname, f.fs), inline_threshold=300)
+        reference_data = h5chunks.translate()
+    return reference_data
+
+
+def _unstrip_protocol(name, fs):
+    # should be upstreamed into fsspec and maybe also
+    # be a method on an OpenFile
+    if isinstance(fs.protocol, str):
+        if name.startswith(fs.protocol):
+            return name
+        return fs.protocol + "://" + name
+    else:
+        if name.startswith(tuple(fs.protocol)):
+            return name
+        return fs.protocol[0] + "://" + name
--- a/pangeo_forge_recipes/reference.py
+++ b/pangeo_forge_recipes/reference.py
@@ -2,14 +2,12 @@
 Functions related to creating fsspec references.
 """
 
-import fsspec
 from fsspec_reference_maker.hdf import SingleHdf5ToZarr
 
 
-def create_hdf5_reference(fname: str, inline_threshold: int = 300, **netcdf_storage_options):
-    with fsspec.open(fname, **netcdf_storage_options) as f:
-        h5chunks = SingleHdf5ToZarr(f, _unstrip_protocol(fname, f.fs), inline_threshold=300)
-        reference_data = h5chunks.translate()
+def create_hdf5_reference(fp, fname: str, inline_threshold: int = 300, **netcdf_storage_options):
+    h5chunks = SingleHdf5ToZarr(fp, _unstrip_protocol(fname, fp.fs), inline_threshold=300)
+    reference_data = h5chunks.translate()
     return reference_data
 
 
--- a/pangeo_forge_recipes/reference.py
+++ b/pangeo_forge_recipes/reference.py
@@ -2,23 +2,27 @@
 Functions related to creating fsspec references.
 """
 
+from typing import Dict, Tuple, Union
+
 from fsspec_reference_maker.hdf import SingleHdf5ToZarr
 
 
-def create_hdf5_reference(fp, fname: str, inline_threshold: int = 300, **netcdf_storage_options):
-    h5chunks = SingleHdf5ToZarr(fp, _unstrip_protocol(fname, fp.fs), inline_threshold=300)
+def create_hdf5_reference(
+    fp, fname: str, url: str, inline_threshold: int = 300, **netcdf_storage_options
+) -> Dict:
+    h5chunks = SingleHdf5ToZarr(fp, url, inline_threshold=inline_threshold)
     reference_data = h5chunks.translate()
     return reference_data
 
 
-def _unstrip_protocol(name, fs):
+def unstrip_protocol(name: str, protocol: Union[str, Tuple[str, ...]]) -> str:
     # should be upstreamed into fsspec and maybe also
     # be a method on an OpenFile
-    if isinstance(fs.protocol, str):
-        if name.startswith(fs.protocol):
+    if isinstance(protocol, str):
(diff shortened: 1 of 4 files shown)


pangeo_forge_recipes/reference.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def unstrip_protocol(name: str, protocol: Union[str, Tuple[str, ...]]) -> str:
+    # should be upstreamed into fsspec and maybe also
+    # be a method on an OpenFile
+    if isinstance(protocol, str):
+        if name.startswith(protocol):
+            return name
+        return protocol + "://" + name
+    else:
```
rabernat, line 25: @martindurant - could you explain this `else` block? Under what circumstances would `protocol` _not_ be a string?

rabernat:
> Is there any possibility that this alone would help with #177?

Yes it's possible. I wish there were an easy way to answer this.

cisaacstern:
> I wish there were an easy way to answer this.

I have an idea for how to investigate this through the reproducer work I'm doing on #177. Will check back here if/when I find anything of note.

pangeo_forge_recipes/storage.py
```diff
@@ -181,6 +181,10 @@ def getitems(self, keys: Sequence[str]) -> dict:
         return {k: json.loads(raw_bytes) for k, raw_bytes in all_meta_raw.items()}
 
 
+# A note about something frustrating about this context manager:
+# Sometimes it yields an fsspec.OpenFile object
+# Other times it yields something lower level, like an _io.BufferedReader
+# This can lead to unpredictable behavior
```
rabernat, line 187: This comment is outdated and needs to be removed.

pangeo_forge_recipes/reference.py
martindurant: It can be a tuple of strings, like `("gcs", "gs")`

pangeo_forge_recipes/storage.py
```diff
@@ -116,14 +115,15 @@ def size(self, path: str) -> int:
...
         logger.debug(f"entering fs.open context manager for {full_path}")
-        with self.fs.open(full_path, **kwargs) as f:
-            logger.debug(f"FSSpecTarget.open yielding {f}")
-            yield f
-            logger.debug("FSSpecTarget.open yielded")
+        of = self.fs.open(full_path, **kwargs)
+        logger.debug(f"FSSpecTarget.open yielding {of}")
+        yield of
+        logger.debug("FSSpecTarget.open yielded")
+        of.close()
```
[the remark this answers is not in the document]
martindurant: Correction: self.fs.open returns a file-like object (io.IOBase subclass, AbstractBufferedFile for s3, gcs, etc.)

For an AbstractBufferedFile, entering the context is a no-op yielding the same object, and exiting calls `.close()`. For other file types, it's something else, with the expectation that exiting the context closes. For example, LocalFileSystem.open -> LocalFileOpener ; LocalFileSystem.__enter__ -> _io.BufferedReader (and exit calls the actual file's exit, calling close()). You *could* do `with` on the file-like.

For fsspec.open, you get OpenFile objects, which can contain more than one file-like layer, for compression and text mode. You must use this with a context, to get the outermost file-like object back.  Leaving the context calls close on the file-like objects in reverse order. If you, instead, call `.open` of the OpenFile, you get the same file-like object, but its .close method has been patched to execute the same chain of close calls, since this object might in this case outlive the OpenFile instance that made it.

pangeo_forge_recipes/storage.py
cisaacstern:
```suggestion

```

pangeo_forge_recipes/storage.py
```diff
@@ -116,14 +115,15 @@ def size(self, path: str) -> int:
         return self.fs.size(self._full_path(path))
 
     @contextmanager
-    def open(self, path: str, **kwargs) -> Iterator[None]:
+    def open(self, path: str, **kwargs) -> Iterator[fsspec.core.OpenFile]:
```
cisaacstern, line 118:
```suggestion
    def open(self, path: str, **kwargs) -> Iterator[io.IOBase]:
```
@martindurant, do I understand [your comment here](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/213/files/c001e710e0a49d2f55476d0a7900819cb26acc2f#r717563338) correctly to mean that the return type of `self.fs.open` is not always an `AbstractBufferedFile`, and therefore this annotation should be changed as proposed here, to be compatible with reading from certain implementations (maybe LocalFileSystem, e.g.)?

According to https://github.com/python/mypy/issues/2984#issuecomment-285721489, I believe mypy should accept subclasses of `BaseClass` for an `Iterator[BaseClass]` annotation.

(If so, we'll need to `import io` in this module, of course.)

martindurant: > the return type of self.fs.open is not always an AbstractBufferedFile

Correct. It is a subclass of AbstractBufferedFile for s3, gcsfs, abfs, ftp, http and others, but can also be LocalFileOpener (io.IOBase), SMBFileOpener (object), SFTPFile (paramiko.BufferedFile), ...

cisaacstern:
@rabernat, I wonder if we should rename `_get_opener` to `_get_source_opener` or similar? Reviewing this PR made me realize that `_get_opener` is perhaps too generic a name for a function which doesn't do all opening in the module. (Your tweaked `FSSpecTarget.open` handles target opening.)

pangeo_forge_recipes/storage.py
rabernat: However, the following assertion fails, so I do see your point

```python
 with file_opener('/etc/hosts') as fp:
     assert isinstance(fp, fsspec.core.OpenFile)
```

So my question is this: __is there some other type besides `fsspec.core.OpenFile` we can use which captures the fact that this is an `ffspec`-related object__.

I added this test, basically asserting that whatever comes out of `file_opener` has an `fs` attribute:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/c001e710e0a49d2f55476d0a7900819cb26acc2f/tests/test_storage.py#L101

This is required by some of the reference-maker code, e.g. here

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/c001e710e0a49d2f55476d0a7900819cb26acc2f/pangeo_forge_recipes/recipes/reference_hdf_zarr.py#L155

martindurant: Maybe better
```
of = file_opener(..)
assert isinstance(of, fsspec.core.OpenFile)
with of as fp:
   ...
```

rabernat: But that's not how we use `file_opener` in Pangeo Forge. We also invoke it as a context manager.

martindurant: There is no fool-proof method to tell whether the output of the context was created by fsspec. You could check for the few concrete classes that we expect, like AbstractBufferedFile, LocalFileOpener.

Issue #207, opened 2021-09-21
0.6.0 release causes authentication issue with fsspec_open_kwargs used in FilePattern


sharkinsspatial:
Moving the discussion of this issue over from https://github.com/pangeo-forge/pangeo-forge-recipes/pull/192 for clarity as per @rabernat 's comment https://github.com/pangeo-forge/pangeo-forge-recipes/pull/192#issuecomment-923497234.  `cache_input` tasks fail when using `fsspec_open_kwargs` fail in a `FilePattern`.  The same recipe code using `pangeo-forge-recipes==0.5.0` with `fsspec_open_kwargs` directly in the recipe args works.  I reproduced this issue using `pangeo-forge-recipes==0.6.0` (potentially introduced via https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167) with the below recipe code

```
import aiohttp
import pandas as pd

from pangeo_forge_recipes.patterns import ConcatDim, FilePattern
from pangeo_forge_recipes.recipes import XarrayZarrRecipe

# TODO: replace with ENV vars
username = password = "pangeo@developmentseed.org"


def make_filename(time):
    input_url_pattern = (
        "https://arthurhouhttps.pps.eosdis.nasa.gov/gpmdata/{yyyy}/{mm}/{dd}/"
        "imerg/3B-HHR.MS.MRG.3IMERG.{yyyymmdd}-S{sh}{sm}00-E{eh}{em}59.{MMMM}.V06B.HDF5"
    ).format(
        yyyy=time.strftime("%Y"),
        mm=time.strftime("%m"),
        dd=time.strftime("%d"),
        yyyymmdd=time.strftime("%Y%m%d"),
        sh=time.strftime("%H"),
        sm=time.strftime("%M"),
        eh=time.strftime("%H"),
        em=(time + pd.Timedelta("29 min")).strftime("%M"),
        MMMM=f"{(time.hour*60 + time.minute):04}",
    )
    return input_url_pattern


dates = pd.date_range("2000-06-01T00:00:00", "2021-05-31T23:59:59", freq="30min")
time_concat_dim = ConcatDim("time", dates, nitems_per_file=1)
pattern = FilePattern(
    make_filename,
    time_concat_dim,
    fsspec_open_kwargs={"auth": aiohttp.BasicAuth(username, password)},
)

recipe = XarrayZarrRecipe(
    pattern,
[...]

PR #215, opened 2021-09-30
Pass kwargs through to pruned FilePatterns

cisaacstern:
@sharkinsspatial, I believe this closes https://github.com/pangeo-forge/pangeo-forge-recipes/issues/207.

Maybe it's worth running a `/run-recipe-test` for the recipe described in that Issue to be sure? I'm pretty confident the tests added here now cover that bug, but running this against the real world Issue would certainly be a good gut check, if it's not too hard for you to do.

### DIFF
--- a/pangeo_forge_recipes/patterns.py
+++ b/pangeo_forge_recipes/patterns.py
@@ -268,9 +268,10 @@ def prune_pattern(fp: FilePattern, nkeep: int = 2) -> FilePattern:
         else:  # pragma: no cover
             assert "Should never happen"
 
-    sig = inspect.signature(fp.__init__)
+    sig = inspect.signature(fp.__init__)  # type: ignore
     kwargs = {
-        param: getattr(fp, param) for param in sig.parameters.keys()
-        if param not in ['format_function', 'combine_dims']
+        param: getattr(fp, param)
+        for param in sig.parameters.keys()
+        if param not in ["format_function", "combine_dims"]
     }
     return FilePattern(fp.format_function, *new_combine_dims, **kwargs)
--- a/pangeo_forge_recipes/patterns.py
+++ b/pangeo_forge_recipes/patterns.py
@@ -267,4 +267,10 @@ def prune_pattern(fp: FilePattern, nkeep: int = 2) -> FilePattern:
         else:  # pragma: no cover
             assert "Should never happen"
 
-    return FilePattern(fp.format_function, *new_combine_dims)
+    return FilePattern(
+        fp.format_function,
+        *new_combine_dims,
+        fsspec_open_kwargs=fp.fsspec_open_kwargs,
+        query_string_secrets=fp.query_string_secrets,
+        is_opendap=fp.is_opendap,
+    )
--- a/pangeo_forge_recipes/patterns.py
+++ b/pangeo_forge_recipes/patterns.py
@@ -2,6 +2,7 @@
 Filename / URL patterns.
 """
 
+import inspect
 from dataclasses import dataclass, field, replace
 from enum import Enum
 from itertools import product
@@ -267,10 +268,9 @@ def prune_pattern(fp: FilePattern, nkeep: int = 2) -> FilePattern:
         else:  # pragma: no cover
             assert "Should never happen"
 
-    return FilePattern(
-        fp.format_function,
-        *new_combine_dims,
-        fsspec_open_kwargs=fp.fsspec_open_kwargs,
-        query_string_secrets=fp.query_string_secrets,
-        is_opendap=fp.is_opendap,
-    )
+    sig = inspect.signature(fp.__init__)
+    kwargs = {
+        param: getattr(fp, param) for param in sig.parameters.keys()
+        if param not in ['format_function', 'combine_dims']
+    }
+    return FilePattern(fp.format_function, *new_combine_dims, **kwargs)
--- a/tests/test_patterns.py
+++ b/tests/test_patterns.py
@@ -29,13 +29,8 @@ def make_concat_merge_pattern(**kwargs):
     def format_function(time, variable):
         return f"T_{time}_V_{variable}"
 
-    if "fsspec_open_kwargs" in kwargs.keys() and "is_opendap" in kwargs.keys():
-        with pytest.raises(ValueError):
-            fp = FilePattern(format_function, merge, concat, **kwargs)
-            return
-    else:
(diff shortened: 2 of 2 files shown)


pangeo_forge_recipes/patterns.py
```diff
@@ -267,4 +267,10 @@ def prune_pattern(fp: FilePattern, nkeep: int = 2) -> FilePattern:
...
             assert "Should never happen"
 
-    return FilePattern(fp.format_function, *new_combine_dims)
+    return FilePattern(
+        fp.format_function,
+        *new_combine_dims,
+        fsspec_open_kwargs=fp.fsspec_open_kwargs,
+        query_string_secrets=fp.query_string_secrets,
+        is_opendap=fp.is_opendap,
+    )
```
rabernat, line 276: Just noting that, _if_ `FilePattern` were a dataclass, we could just call `replace` here. Because it is not, we have to remember to update this method every time we change the signature of `FilePattern.__init__`.

tests/test_patterns.py
```diff
@@ -146,8 +149,25 @@ def test_file_pattern_concat_merge(runtime_secrets, pickle, concat_merge_pattern
 
 
 @pytest.mark.parametrize("nkeep", [1, 2])
-def test_prune(nkeep, concat_merge_pattern):
-    fp = concat_merge_pattern[0]
+def test_prune(nkeep, concat_merge_pattern_with_kwargs, runtime_secrets):
+    if not concat_merge_pattern_with_kwargs:
+        # if `fsspec_open_kwargs` are passed with `is_opendap`, `FilePattern.__init__` raises
+        # ValueError and `concat_merge_pattern_with_kwargs` returns None, so nothing to test
+        return
```
rabernat, line 156: I'm not a fan of this behavior (having a fixture sometimes return `None`). It feels hacky and fragile. 

Best solution: refactor the fixtures so that this situation (`fsspec_open_kwargs` are passed with `is_opendap`) never happens.

Acceptable solution: make `concat_merge_pattern_with_kwargs` call `pytest.skip` when this condition is met.

rabernat (changes requested):
Thanks for the quick turnaround here.

Your inline comment suggests to me that we need to tweak our fixtures a little bit.

pangeo_forge_recipes/patterns.py
```diff
@@ -267,10 +268,9 @@ def prune_pattern(fp: FilePattern, nkeep: int = 2) -> FilePattern:
...
-        *new_combine_dims,
-        fsspec_open_kwargs=fp.fsspec_open_kwargs,
-        query_string_secrets=fp.query_string_secrets,
-        is_opendap=fp.is_opendap,
-    )
+    sig = inspect.signature(fp.__init__)
+    kwargs = {
+        param: getattr(fp, param) for param in sig.parameters.keys()
+        if param not in ['format_function', 'combine_dims']
+    }
```
rabernat, line 275: 👌

tests/test_patterns.py
```diff
@@ -93,19 +89,20 @@ def test_pattern_from_file_sequence():
...
             if not fp.is_opendap:
                 fp.fsspec_open_kwargs.update(runtime_secrets["fsspec_open_kwargs"])
             else:
-                return
+                pytest.skip(
+                    "`fsspec_open_kwargs` should never be used in combination with `is_opendap`. "
+                    "This is checked in `FilePattern.__init__` but not when updating attributes. "
+                    "Proposed changes to secret handling will obviate the need for runtime updates"
+                    " to attributes in favor of encryption. So for now, we'll just skip this."
+                )
```
rabernat, line 105: 👌

tests/test_patterns.py
```diff
@@ -148,12 +145,15 @@ def test_file_pattern_concat_merge(runtime_secrets, pickle, concat_merge_pattern
         assert fp.fsspec_open_kwargs == {}
 
 
+def test_incompatible_kwargs():
+    kwargs = dict(fsspec_open_kwargs={"block_size": "foo"}, is_opendap=True)
+    with pytest.raises(ValueError):
+        make_concat_merge_pattern(**kwargs)
+        return
```
rabernat, line 152: 👌

rabernat (approved):

pangeo_forge_recipes/patterns.py
cisaacstern: Ok phew 😅  wasn't sure if this was getting a little too tricky, glad to get your 👌 

I agree having to change the prune function every time the signature of FilePattern.__init__ changes is asking for problems... but as a follow-on to our discussion in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/167#discussion_r700633315, it looks like [dataclasses.replace](https://docs.python.org/3/library/dataclasses.html#dataclasses.replace) doesn't work out-of-the-box for dataclasses with custom __init__s (which are required for variable numbers of *args):

> Be forewarned about how init=False fields work during a call to replace(). They are not copied from the source object, but rather are initialized in __post_init__(), if they’re initialized at all. It is expected that init=False fields will be rarely and judiciously used. If they are used, it might be wise to have alternate class constructors, or perhaps a custom replace() (or similarly named) method which handles instance copying.

So this was my workaround for that

rabernat:
Looks like you still haven't set up [pre-commit hooks for git](https://pre-commit.com/#quick-start)? Otherwise it would not be possible for you to push a commit that fails pre-commit.

cisaacstern:
🤦  I have, but looks like I have to `pre-commit install` every time I activate my conda development environment? I don't always remember to do this and haven't gotten in the habit of looking for the pre-commit run after every commit... so this one slipped through the cracks

rabernat:
> but looks like I have to `pre-commit install` every time I activate my conda development environment?

No. I definitely don't have to do that. You should only have to install once per environment.

cisaacstern:
> You should only have to install once per environment.

Oh I think I realize what happened ... I made a new clone of `pangeo-forge-recipes` before working on this branch (had a lot of branches in various states that I didn't need, and re-cloning seemed easier than cleaning up my local copy), so the pre-commit was not in `.git/hooks/pre-commit`

pangeo_forge_recipes/patterns.py
cisaacstern: Further discussion of this in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/215#discussion_r719624247 (noting for anyone who comes across this in the future)

Issue #214, opened 2021-09-29
KeyError raised by `consolidate_dimension_coordinates`


cisaacstern:
In re-building portions of https://github.com/pangeo-forge/staged-recipes/pull/24, the [`consolidate_dimension_coordinates`](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/244fe1d5df48081eeb94bfbcfc66505ec0725708/pangeo_forge_recipes/recipes/xarray_zarr.py#L640) pathway added in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/210 raised a KeyError on the manual call to `finalize_target`, with the following Traceback:

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-10-7e80b39f4210> in <module>
----> 1 rec.finalize_target()

/srv/conda/envs/notebook/lib/python3.8/site-packages/pangeo_forge_recipes/recipes/xarray_zarr.py in finalize_target(target, consolidate_zarr, consolidate_dimension_coordinates)
    644         dims = _gather_coordinate_dimensions(group)
    645         for dim in dims:
--> 646             arr = group[dim]
    647             attrs = dict(arr.attrs)
    648             new = group.array(

/srv/conda/envs/notebook/lib/python3.8/site-packages/zarr/hierarchy.py in __getitem__(self, item)
    347                          synchronizer=self._synchronizer)
    348         else:
--> 349             raise KeyError(item)
    350 
    351     def __setitem__(self, item, value):

KeyError: 'y'
```

`finalize_target` worked fine with `consolidate_dimension_coordinates=False`. The dataset which raised this error can be accessed as follows:
```python
import fsspec
import xarray as xr

path = "s3://Pangeo/pangeo-forge/swot_adac/eNATL60/Region03/interior_daily/aso.zarr"
mapper = fsspec.get_mapper(
    path, anon=True, client_kwargs={'endpoint_url': 'https://ncsa.osn.xsede.org'},
)
ds = xr.open_zarr(mapper, consolidated=True)
print(ds)
```

<details>

```
<xarray.Dataset>
Dimensions:       (y: 675, x: 574, deptht: 107, depthu: 107, depthv: 107, depthw: 107, time_counter: 92)
Coordinates: (12/18)
    depth         (y, x) float32 dask.array<chunksize=(15, 574), meta=np.ndarray>
[...]

Issue #211, opened 2021-09-27
Potential refactor of Recipe classes to use shared context


rabernat:
Trying to capture a suggestion by @alxmrs at today's meeting.

### Current Situation

Presently, our recipe classes use partials to bind the numerous recipe configuration parameters into functions with zero or one arguments, e.g. for `XarrayZarrRecipe.store_chunk`

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/7311c3e7ac1d32710c52221f92a08eef48a5d59d/pangeo_forge_recipes/recipes/xarray_zarr.py#L865-L873

we can see that we are basically just copying over most of the attributes from `self` to the keyword arguments of the class.

This function is invoked with a single argument (`chunk_key`) by the executors: 

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/7311c3e7ac1d32710c52221f92a08eef48a5d59d/pangeo_forge_recipes/recipes/base.py#L173-L175

The actual work happens in a standalone function:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/7311c3e7ac1d32710c52221f92a08eef48a5d59d/pangeo_forge_recipes/recipes/xarray_zarr.py#L527-L533

...from which we also need to pass around a bunch of similar arguments to other functions:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/7311c3e7ac1d32710c52221f92a08eef48a5d59d/pangeo_forge_recipes/recipes/xarray_zarr.py#L551-L556

Not only is this very hard to read, maintain, and extend; it also _might_ be related to the persistent challenges related to serialization.

### Alternative: use a single object to hold recipe configuration

We could put all of this config into a single object (Alex called it a "context"). This was actually proposed during the previous refactor (https://github.com/pangeo-forge/pangeo-forge-recipes/pull/160#discussion_r652762000) but we didn't end up going that route.

Calling code might look like this:
```python
recipe_config = XarrayZarrRecipeConfig(
    nitems_per_input=10,
    ...
)

recipe = XarrayZarrRecipe(recipe_config)
```

The config would be an explicit argument on all of the recipe functions. Invoking it from prefect might look like
```python
 store_task = store_chunk_task.map( 
     chunk_key=list(self.iter_chunks()),
     recipe_config=unmapped(recipe_config)
     upstream_tasks=[unmapped(prepare_task)], 
 ) 
```

### Questions

[...]

PR #216, opened 2021-09-30
finalize_target handles coordinateless dimensions

TomAugspurger:
Closes https://github.com/pangeo-forge/pangeo-forge-recipes/issues/214

I'm not sure if this was the easiest way to get a test case. I wanted to avoid doubling the size of the test suite just for this issue, so I made a set of fixture similar to the existing ones, but whose dataset has a coordinateless dimension.

### DIFF
--- a/pangeo_forge_recipes/recipes/xarray_zarr.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr.py
@@ -641,7 +641,10 @@ def finalize_target(
         logger.info("Consolidating dimension coordinate arrays")
         target_mapper = target.get_mapper()
         group = zarr.open(target_mapper, mode="a")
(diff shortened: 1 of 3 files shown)


cisaacstern:
> I wanted to avoid doubling the size of the test suite just for this issue, so I made a set of fixture similar to the existing ones, but whose dataset has a coordinateless dimension.

Makes sense to me

rabernat:
Thanks Tom. This looks great!

It feels like a lot of boilerplate to write in order to get to this test. But I don't really see a way around it.

In a future hypothetical refactor related to #211, we would gain the ability to write better units tests for individual functions within recipes, rather than having to test everything via this heavily nested, integration-style tests. I think we could revisit this then.

PR #217, opened 2021-10-01
Fixed memory usage for IMERG recipe

TomAugspurger:
This registers a few custom objects with dask.sizeof.sizeof. Previously,
Dask was underestimating the size of these objects, causing it to
schedule too many tasks on a single worker, overwhelming its memory.

I'm thinking through tests for this, but not coming up with much. In particular, I'm worried that adding new custom objects in the future will cause similar issues.

cc @sharkinsspatial

### DIFF
--- a/pangeo_forge_recipes/patterns.py
+++ b/pangeo_forge_recipes/patterns.py
@@ -2,7 +2,7 @@
 Filename / URL patterns.
 """
 
-from dataclasses import dataclass, field, replace
+from dataclasses import asdict, dataclass, field, replace
 from enum import Enum
 from itertools import product
 from typing import (
@@ -18,6 +18,8 @@ from typing import (
     Tuple,
     Union,
 )
+
+import dask.sizeof
 
 
 class CombineOp(Enum):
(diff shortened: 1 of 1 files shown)


pangeo_forge_recipes/patterns.py
```diff
@@ -268,3 +270,22 @@ def prune_pattern(fp: FilePattern, nkeep: int = 2) -> FilePattern:
             assert "Should never happen"
 
     return FilePattern(fp.format_function, *new_combine_dims)
+
+
+# Register custom objects with Dask, to ensure memory usage is tracked properly
+
+
+@dask.sizeof.sizeof.register(FilePattern)
```
TomAugspurger, line 278: Actually, these should use the bulitin `__sizeof__` protocol since we control these objects.

TomAugspurger:
Well, now this fix isn't working in my manual testing...  Let me keep messing with it.

TomAugspurger:
Going to close this for now. It's going stale and it's unclear if it's necessary.

PR #218, opened 2021-10-02
Let XarrayZarrRecipe use fsspec references for opening netCDF inputs

### LINKED ISSUE #177: Hanging during `store_chunk` loop
While manually executing https://github.com/pangeo-forge/staged-recipes/pull/66, I encountered recurring hangs at multiple places within this loop:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/c4867719c4866ec46436b5d53d95ceb470a6f72e/pangeo_forge_recipes/recipes/xarray_zarr.py#L590-L626

In order to pinpoint the hang, I added [these additional logs](https://github.com/cisaacstern/pangeo-forge-recipes/compare/master..extra-write-loop-logs) to `xarray_zarr.py`. Anecdotally, running with these additional logs indicated that the hang occurred most often at the call to `np.asarray`, but __*not exclusively*__ at that line. None of the source file variable arrays for this recipe exceed much more than ~60 MB, and I was running a **Large (12 GB - 16 GB)** notebook server on https://us-central1-b.gcp.pangeo.io/hub, so it's hard to imagine it was a memory issue. (Note also that input file to target chunk mapping of this recipe is 1:1, and the input files are about ~80 MB in size.)

Hanging within the `store_chunk` loop arose at unpredictable intervals: usually after writing  between 5 and 25 chunks, but sometimes after writing as many as a few hundred. Again anecdotally, it seemed that the hangs became more frequent once I'd reached around the 1500th loop or so.

`KeyboardInterrupt`ing and restarting the `store_chunk` loop from the hang location (without restarting the kernel) **always resolved the issue** and allowed the write to continue for another 5-25 writes before hanging again.

Ultimately, I restarted the loop like this a few dozen times, and eventually got all 2117 inputs written. During this process, I tried the following, to no avail:

- switching to  https://staging.us-central1-b.gcp.pangeo.io/hub
- switching back to  https://us-central1-b.gcp.pangeo.io/hub, and selectively updating all packages which could be involved in this loop to their latest versions: `xarray`, `numpy`, `dask`, `distributed`, `fsspec`, `s3fs`, and `gcsfs`
- removing an explicit `target_chunks` kwarg from the recipe in https://github.com/pangeo-forge/staged-recipes/pull/66/commits/3b9d3fa151b6e6e12b837af391ed38f159c7fd8d, in case that was somehow redundant with `nitems_per_file=1`
- changing the call to `np.asarray` to `var.to_numpy()` (which was admittedly unlikely to help, given that xarray implements `to_numpy` with `np.asarray` internally, and also because the hang was not exclusively on that line)

This feels like an environment or cloud I/O issue to me, but at this point I'm prepared to believe anything.

I've made a notebook which reproduces the execution scenario here: https://github.com/cisaacstern/pangeo-forge-debug-examples/tree/soda342-ice ... with the caveat the storage target is an `fsspec.LocalFileSystem` (since we can't include cloud creds in a public repo). The Binder link in the README does allow the notebook to be run, but it looks like Binder's local disk allotment of 500 MB (?) may fill up before the hang is reproduced.

rabernat:
Potentially closes #177 by allowing us to bypass h5py when reading netCDF inputs from cloud storage.

After #213 this is pretty simple.

The tests should fail like this

```
recipe-netcdf_local_file_pattern_sequential-netcdf_local_paths_sequential_2d-execute_recipe_manual] FAILED [  6%]

=================================================== FAILURES ===================================================
_ test_recipe_with_references[execute_recipe_no_prefect-netCDFtoZarr_recipe-netcdf_local_file_pattern_sequential-netcdf_local_paths_sequential_2d-execute_recipe_manual] _

recipe_fixture = (<class 'pangeo_forge_recipes.recipes.xarray_zarr.XarrayZarrRecipe'>, <FilePattern {'time': 5}>, {'input_cache': Cache...t 0x1801146a0>, root_path='/private/var/folders/n8/63q49ms55wxcj_gfbtykwp5r0000gn/T/pytest-of-rpa/pytest-45/target11'))
execute_recipe = <function execute_recipe_manual.<locals>.execute at 0x180e78ee0>

    @pytest.mark.parametrize("recipe_fixture", all_recipes)
    def test_recipe_with_references(recipe_fixture, execute_recipe):
        """Same as above, but use fsspec references for opening the files."""
    
        RecipeClass, file_pattern, kwargs, ds_expected, target = recipe_fixture
        rec = RecipeClass(file_pattern, open_input_with_fsspec_reference=True, **kwargs)
        execute_recipe(rec)
        ds_actual = xr.open_zarr(target.get_mapper()).load()
>       xr.testing.assert_identical(ds_actual, ds_expected)
E       AssertionError: Left and right Dataset objects are not identical
E       
E       Differing coordinates:
E       L * time     (time) datetime64[ns] 2010-01-01 2010-01-02 ... 2010-01-10
E           NAME: time
E           _Netcdf4Dimid: 0
E       R * time     (time) datetime64[ns] 2010-01-01 2010-01-02 ... 2010-01-10
E       L * lon      (lon) float64 5.0 15.0 25.0 35.0 45.0 ... 325.0 335.0 345.0 355.0
E           NAME: lon
E           _Netcdf4Dimid: 2
E           long_name: longitude
E           units: degrees_east
E       R * lon      (lon) float64 5.0 15.0 25.0 35.0 45.0 ... 325.0 335.0 345.0 355.0
E           units: degrees_east
E           long_name: longitude
E       L * lat      (lat) float64 5.0 15.0 25.0 35.0 45.0 ... 145.0 155.0 165.0 175.0
E           NAME: lat
E           _Netcdf4Dimid: 1
E           long_name: latitude
E           units: degrees_north
E       R * lat      (lat) float64 5.0 15.0 25.0 35.0 45.0 ... 145.0 155.0 165.0 175.0
E           units: degrees_north
E           long_name: latitude
E           _Netcdf4Coordinates: [0, 1, 2]
E           _Netcdf4Dimid: 0
E           long_name: Fantastic Foo
E       R   foo      (time, lat, lon) float64 0.417 0.7203 0.0001144 ... 0.1179 0.3748
E           long_name: Fantastic Foo
E       L   bar      (time, lat, lon) float64 9.0 4.0 3.0 2.0 2.0 ... 3.0 3.0 9.0 5.0
E           _Netcdf4Coordinates: [0, 1, 2]
E           _Netcdf4Dimid: 0
E           long_name: Beautiful Bar
E       R   bar      (time, lat, lon) int64 9 4 3 2 2 8 0 0 4 8 ... 9 8 4 4 6 0 3 3 9 5
E           long_name: Beautiful Bar
E       Attributes only on the left object:
E           _NCProperties: version=2,netcdf=4.8.0,hdf5=1.10.6

tests/recipe_tests/test_XarrayZarrRecipe.py:143: AssertionError
```

It looks like fsspec-reference-maker is giving more metadata attributes than the h5netcdf library.

### DIFF
--- a/pangeo_forge_recipes/recipes/xarray_zarr.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr.py
@@ -14,12 +14,14 @@ from math import ceil
 from typing import Callable, Dict, Hashable, Iterator, List, Optional, Sequence, Set, Tuple
 
 import dask
+import fsspec
 import numpy as np
 import xarray as xr
 import zarr
 
 from ..chunk_grid import ChunkGrid
 from ..patterns import CombineOp, DimIndex, FilePattern, Index
+from ..reference import create_hdf5_reference, unstrip_protocol
 from ..storage import AbstractTarget, CacheFSSpecTarget, MetadataTarget, file_opener
 from ..utils import calc_subsets, fix_scalar_attr_encoding, lock_for_conflicts
 from .base import BaseRecipe, FilePatternRecipeMixin
@@ -45,9 +47,14 @@ SubsetSpec = Dict[str, int]
 # (e.g. {'time': 5, 'depth': 2})
 
 
-def _input_metadata_fname(input_key):
+def _input_metadata_fname(input_key: InputKey) -> str:
     key_str = "-".join([f"{k.name}_{k.index}" for k in input_key])
     return "input-meta-" + key_str + ".json"
+
+
+def _input_reference_fname(input_key: InputKey) -> str:
+    key_str = "-".join([f"{k.name}_{k.index}" for k in input_key])
+    return "input-reference-" + key_str + ".json"
 
 
 def inputs_for_chunk(
@@ -147,6 +154,7 @@ def cache_input_metadata(
     xarray_open_kwargs: dict,
     delete_input_encoding: bool,
     process_input: Optional[Callable[[xr.Dataset, str], xr.Dataset]],
+    open_input_with_fsspec_reference: bool,
 ) -> None:
     if metadata_cache is None:
         raise ValueError("metadata_cache is not set.")
@@ -160,9 +168,39 @@ def cache_input_metadata(
         xarray_open_kwargs=xarray_open_kwargs,
         delete_input_encoding=delete_input_encoding,
         process_input=process_input,
+        open_input_with_fsspec_reference=open_input_with_fsspec_reference,
+        metadata_cache=metadata_cache,
     ) as ds:
         input_metadata = ds.to_dict(data=False)
         metadata_cache[_input_metadata_fname(input_key)] = input_metadata
+
+
+def make_input_reference(
+    input_key: InputKey,
+    metadata_cache: MetadataTarget,
+    file_pattern: FilePattern,
+    input_cache: Optional[CacheFSSpecTarget],
+    copy_input_to_local_file: bool,
+) -> None:
+
+    fname = file_pattern[input_key]
+    if input_cache is None:
+        protocol = fsspec.utils.get_protocol(fname)
+        url = unstrip_protocol(fname, protocol)
+    else:
+        url = unstrip_protocol(input_cache._full_path(fname), input_cache.fs.protocol)
+    with file_opener(
+        fname,
+        cache=input_cache,
+        copy_to_local=copy_input_to_local_file,
+        bypass_open=False,
+        secrets=file_pattern.query_string_secrets,
+        **file_pattern.fsspec_open_kwargs,
+    ) as fp:
+        ref_data = create_hdf5_reference(fp, url, fname)
+
+    ref_fname = _input_reference_fname(input_key)
+    metadata_cache[ref_fname] = ref_data
 
 
 def cache_input(
@@ -176,6 +214,7 @@ def cache_input(
     delete_input_encoding: bool,
     process_input: Optional[Callable[[xr.Dataset, str], xr.Dataset]],
     metadata_cache: Optional[MetadataTarget],
+    open_input_with_fsspec_reference: bool,
 ) -> None:
     if cache_inputs:
         if file_pattern.is_opendap:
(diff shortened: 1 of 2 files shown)


martindurant:
> It looks like fsspec-reference-maker is giving more metadata attributes than the h5netcdf library.

Specifically, h5py - it seems that these are attributes that h5netcdf is swallowing to apply CDF conventions. The code could clean them out, too, although they probably don't do any harm.

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -14,12 +14,14 @@
...
 
 import dask
+import fsspec
 import numpy as np
 import xarray as xr
 import zarr
 
 from ..chunk_grid import ChunkGrid
 from ..patterns import CombineOp, DimIndex, FilePattern, Index
+from ..reference import create_hdf5_reference, unstrip_protocol
```
martindurant, line 24: Note that `_unstrip_protocol` is now available in `fsspec.utils`. A more thorough one for chained FSs would be good, but not necessary for the work here.

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -160,11 +168,41 @@ def cache_input_metadata(
...
+        cache=input_cache,
+        copy_to_local=copy_input_to_local_file,
+        bypass_open=False,
+        secrets=file_pattern.query_string_secrets,
+        **file_pattern.fsspec_open_kwargs,
+    ) as fp:
+        ref_data = create_hdf5_reference(fp, url, fname)
+
+    ref_fname = _input_reference_fname(input_key)
+    metadata_cache[ref_fname] = ref_data
```
martindurant, line 203: Maybe this is a recipe detail, but it would be good to figure out where we can store these reference files for the future, perhaps together with a hash/uid of the file they were created from.

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -268,35 +320,56 @@ def open_input(
...
-                    ds = process_input(ds, str(fname))
+        with file_opener(
+            fname,
+            cache=cache,
+            copy_to_local=copy_input_to_local_file,
+            bypass_open=file_pattern.is_opendap,
+            secrets=file_pattern.query_string_secrets,
+            **file_pattern.fsspec_open_kwargs,
+        ) as f:
+            with dask.config.set(scheduler="single-threaded"):  # make sure we don't use a scheduler
```
martindurant, line 355: Why do we not want a scheduler? Because we are dealing with local files?

martindurant (commented):
Well it seems pretty simple on a quick read! 

Am I right in thinking that the input to reference-maker is always single files in this model? It may be efficient to pre-combine sets of filenames (if the combination logic is sound, of course). I would also argue, if practical, that a reference set for the whole aggregated dataset should always be created as a side-product of copying to zarr. If a future recipe wants to make a new chunking or re-copy the data (which hasn't changed), then reusing that artefact would save a lot of CPU time, as well as, of course, allow direct access to the original bytes.

(Opening many reference files with open_mf can still use a lot of memory when it is set to check the coordinates or attributes of all of the dataset, be warned; I don't suppose this is any different for input by virtual zarr versus any other)

PR #219, opened 2021-10-02
Recipe context

### LINKED ISSUE #211: Potential refactor of Recipe classes to use shared context

rabernat:
closes #211


TomAugspurger:
I'll need to take a closer look later, but I'm not 100% sure this will solve the memory issues we're seeing on workers / the scheduler when deserializing. This is partially due to not fully understanding why the memory issues are showing up in the first place, but this feels equivalent to just passing around the context on the `self` object like we did earlier.

> The choices are:

I think there is a third option, which mostly retains the current API, but refactors functions like `pangeo_forge_recipes.recipes.xarray_zarr.cache_input` that currently take `file_pattern` and instead pass the things they extract out of `file_pattern` (specifically the ConcatDim.keys in the case of IMERG).

The tricky part is that some of these things (like `fname`) depend on things like `input_key` or `chunk_key`, which we don't know until the function is actually called. So I think that one option forward is to update `iter_inputs` and `iter_chunks` to yield these additional pieces of metadata.

Changing the return value of `iter_inputs` is technically an API breaking change, but I'm less concerned with that at this stage of the project.

rabernat:
> but this feels equivalent to just passing around the context on the self object like we did earlier.

I see what you're saying here. I was hoping this is different because there are no funky partials. My hope was that the (potentially large) config object would just get serialized once, rather than N times for each task. But agreed, it may not actually work. 

However, even it it doesn't solve the memory issue, it may be worth pursuing for the beam compatibility.

rabernat:
@TomAugspurger - do you have 15 minutes today to chat about this? I think it would save us a lot of back and forth? I'm free most of the day.

rabernat:
Cool! I'll go on https://whereby.com/pangeo and hang out there for the rest of the hour.

TomAugspurger:
Summary of the brief call:

We're going to test out this change. Ryan has a small change to make to the Dask task graph construction to ensure that the config is serialized exactly once, and then I'll update my testing code from #151 to see if memory usage still looks good on this branch.

TomAugspurger:
Actually @rabernat I have a small additional test to perform. I'm going to try serializing each of the partially applied functions using the same idea (embed them once in the task graph, rather than per task). I don't know if it'll behave similarly to serializing the object context, but I'd guess it would.

We'll likely still want to pursue the refactor here regardless of the outcome.

TomAugspurger:
https://gist.github.com/TomAugspurger/792f521bf2c086fd820324d18ef349ce has my manual tests for memory usage on this branch with the updated IMERG recipe (to use this config). Things seem OK, so I think we can consider the *dask* memory issue fixed by including the large objects in the task graph exactly once (either the partially applied functions in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/219#issuecomment-933574661, or the Config in this RP).

cisaacstern (commented):
`xarray_zarr.py` is _so_ much more readable without the kwargs passing; impressive! 

> Just break the API? This would mean rewriting all our tutorials, examples, intro guide, etc.

IIUC, this is the direction we're heading?

rabernat:
> Things seem OK, so I think we can consider the _dask_ memory issue fixed by including the large objects in the task graph exactly once

:tada: 

How would we do the equivalent check for prefect?

> https://gist.github.com/TomAugspurger/792f521bf2c086fd820324d18ef349ce has my manual tests for memory usage on this branch with the updated IMERG recipe

It would be so helpful to somehow get this check into our test suite. What are the manual steps involved? Do you just watch your memory using via `top` or something like that? If so could, we automate this with a memory profiler?

---

Regarding the overall refactor, I'm inclined to move forward with this PR. Beyond the memory issue, it has the following benefits
- Simplify recipe classes by removing tons of boilerplate
- Beam compatibility down the line (@alxmrs, would love your feedback here eventually on whether we are on the right track)

Having thought about it a bit more, I think we may be able to do it without an API change. I'll keep working on that.

cisaacstern:
> Having thought about it a bit more, I think we may be able to do it without an API change. I'll keep working on that.

This would be incredible. API change was my only reservation here.

TomAugspurger:
> It would be so helpful to somehow get this check into our test suite. What are the manual steps involved?

I think no manual steps involved. I wait to see if Dask's memory monitor kills the worker for exceeding its budget. I'll put up a PR with a (failing) test on main.

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -939,37 +708,55 @@ def inputs_for_chunk(self, chunk_key: ChunkKey) -> Sequence[InputKey]:
         ninputs = self.file_pattern.dims[self.file_pattern.concat_dims[0]]
         return inputs_for_chunk(chunk_key, self.inputs_per_chunk, ninputs)
 
+
+@dataclass
+class XarrayZarrRecipe(BaseRecipe):
+
+    config: XarrayZarrRecipeConfig
```
alxmrs, line 715: If you didn't want to break the previous API: The `BaseRecipe` could take a constructor that builds the config -- that way, the use of configs would be more of an internal implementation detail.

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -939,37 +708,55 @@ def inputs_for_chunk(self, chunk_key: ChunkKey) -> Sequence[InputKey]:
...
-            target_chunks=self.target_chunks,
-            file_pattern=self.file_pattern,
-            input_cache=self.input_cache,
-            cache_inputs=self.cache_input,
-            copy_input_to_local_file=self.copy_input_to_local_file,
-            xarray_open_kwargs=self.xarray_open_kwargs,
-            delete_input_encoding=self.delete_input_encoding,
-            process_input=self.process_input,
-        ) as ds:
+        with open_chunk(self.config, chunk_key,) as ds:
```
alxmrs, line 750: nit: remove last `,`.

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -939,37 +708,55 @@ def inputs_for_chunk(self, chunk_key: ChunkKey) -> Sequence[InputKey]:
...
             yield ds
+
+    def iter_inputs(self) -> Iterator[InputKey]:
+        for input_key in self.config.iter_inputs():
```
alxmrs, line 754: I think this could also be just: 
```
yield from self.config.iter_inputs()
```
Same with `iter_chunks` below.

pangeo_forge_recipes/recipes/xarray_zarr.py
alxmrs: Some thoughts on the interface for this: 
Since python doesn't support multiple constructors, you could structure the API like 
```
def __init__(self, paramA=None, paramB=None, ..., *, config: t.Optional[XarrayZarrRecipeConfig] = None):
```
This way, the config would have to be passed in as a key-word arg.

alxmrs (commented):
Here are some general thoughts about the PR; I will respond to the main discussion in a little bit.

alxmrs:
> So I think we can judge this PR solely on how it affects the user API and maintenance burden.

Yes, I think this is the main benefit of this PR; it makes a great improvement here. 

> Beam compatibility down the line (@alxmrs, would love your feedback here eventually on whether we are on the right track)

TL;DR: I think that either way, we'll be able to have Beam support working. I think that this change helps make this much easier.

Let me summarize the takeaways from a meeting that @TomAugspurger, @cisaacstern, and I had last week: I was primarily focused on changes in Rechunker, specifically with MultiSagePipelines -- my argument was that these should have a config/context object. Before the meeting, I thought that state was often passed between workers via closures, an assumption I made having only read unit tests in that project. In the meeting, I found out that the unit test in question was pretty much the only case where this pattern was used -- mostly, to fit with pytest fixtures. We found out that the `functools.partial` approach should also work with beam (it does in fact serialize well with pickle).

That said, this PR makes supporting beam much simpler. We can basically translate the `to_python` version of the pipeline with beam idioms. It would be something like: 

```
      def to_beam():
            # ...
            pcoll |= beam.Create(self.iter_inputs()) | beam.Map(self.cache_input, config=self.config)
            pcoll |= beam.Create(self.iter_chunks()) | beam.Map(self.store_chunk, config=self.config)
            pcoll |= beam.ParDo(finalize_target, config=self.config)
            return pcoll
```

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -137,163 +137,109 @@ def chunk_position(chunk_key: ChunkKey) -> int:
...
-    cache_metadata: bool,
-    copy_input_to_local_file: bool,
-    xarray_open_kwargs: dict,
-    delete_input_encoding: bool,
-    process_input: Optional[Callable[[xr.Dataset, str], xr.Dataset]],
-    metadata_cache: Optional[MetadataTarget],
-) -> None:
-    if cache_inputs:
-        if file_pattern.is_opendap:
+def cache_input(config: XarrayZarrRecipeConfig, input_key: InputKey) -> None:
```
alxmrs, line 140: Here's an optional style suggestion: 

There's a python feature that I enjoy that seems to be a good fit for these functions ([keyword only arguments](https://www.asmeurer.com/python3-presentation/slides.html#8)):
```
def cache_input(input_key: InputKey, *, config: XarrayZarrRecipeConfig) -> None:
...
```
In order to invoke this function, the caller would have to write:
```
cache_input(my_input_key, config=config)
```


I think this structure is a good fit for these functions since: 
- it seems more natural for the first argument to imply the value being operated on, instead of a shared context
- it forces the caller to pass in the config as a key-word argument. I think this is less confusing to callers. It helps people remember to include state; error messages help point he developer in the right direction.
- it prevents another object from accidentally being passed in as the config.

rabernat:
Hi Folks! I have just pushed a big update that (I hope) fully resolves all remaining issues. I have updated my top-level comment in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/219#issue-748043942 and am looking for reviews again.

Also note the mypy question just above.

TomAugspurger:
> Is the way I am building the dask HighLevelGraph in pangeo_forge_recipes/executors/dask.py aligned with best practices? Do you see anything that could go wrong? @TomAugspurger

Looks good to me at a quick glance. Calling `recipe.to_dask().visualize()` on the terraclimate recipe makes shows that things look reasonable.

I don't have any immediate guesses on the typing issue.

alxmrs:
I left a couple of guests to address the typing issue. I'm working on a tightish deadline right now, but if I find some time, I'll pull this branch locally and see if I can find a solution.

PR #220, opened 2021-10-04
Added failing tests for #151

TomAugspurger:
This adds a tests for the memory issues from #151. It should fail with a TimeoutError after 90s.

### DIFF
--- a/tests/recipe_tests/test_XarrayZarrRecipe.py
+++ b/tests/recipe_tests/test_XarrayZarrRecipe.py
@@ -2,6 +2,7 @@ from contextlib import nullcontext as does_not_raise
 from dataclasses import replace
 from unittest.mock import patch
 
+import dask.core
 import pytest
 import xarray as xr
 import zarr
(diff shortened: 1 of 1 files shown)


TomAugspurger:
There's the timeout error: https://github.com/pangeo-forge/pangeo-forge-recipes/pull/220/checks?check_run_id=3795717468#step:8:821

In theory, it should be fixed with https://github.com/pangeo-forge/pangeo-forge-recipes/pull/220/commits/0f9ce09ba793a8845ff31eb4ee5be441269818a2.

Then we'll decide how to reconcile this PR and the larger changes in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/219 (this will create a merge conflict, so isn't worth merging if https://github.com/pangeo-forge/pangeo-forge-recipes/pull/219 will fix it anyway).

TomAugspurger:
So it turns out that my "fix" is faulty. Dask doesn't replace functions in tasks, just arguments, so putting the functions in the task graph once doesn't work.

I don't think that #219 suffers from the same problem, since the large objects will be in the context as an argument, rather than the function.

PR #225, opened 2021-11-07
Apache Beam Executor

### LINKED ISSUE #169: Support Beam as an executor for recipe pipelines
As discussed in today's [meeting], our internal users would really like to execute Recipes on [Beam] (and Dataflow). What is the best way to support this functionality? 

To this end, I mentioned using [XArray-Beam] within Pangeo in the meeting. @rabernat pointed out that this is probably not the best direction, since we don't necessarily want to introduce a dependency on that project.

[meeting]: https://docs.google.com/document/d/14FpI9vaM6TeFtmM7LP9o_d5DZaYKgQVTTzT7tFRt-Nw/edit#heading=h.u9p2mwlg964t
[Beam]: https://beam.apache.org/
[XArray-Beam]: https://github.com/google/xarray-beam

alxmrs:
The pangeo-forge pipeline model is adapted to run on Apache Beam. Here, we use dummy stage numbers ("steps") as the input and output PCollections needed to work with Beam's programming model. Where substantive inputs need to be mapped to stages with arguments, "mappables" are combined with these steps. The result is that each stage executes serially, and mappable stages execute in parallel.

Fixes #169.

### DIFF
--- a/ci/py3.8.yml
+++ b/ci/py3.8.yml
@@ -4,6 +4,7 @@ channels:
 dependencies:
   - python=3.8
   - aiohttp
+  - apache_beam
   - black
   - boto3
   - cfgrib<0.9.9.0
--- a/ci/py3.9.yml
+++ b/ci/py3.9.yml
@@ -4,6 +4,7 @@ channels:
 dependencies:
   - python=3.9
   - aiohttp
+  - apache_beam
   - black
   - boto3
   - cfgrib<0.9.9.0
--- a/ci/py3.9.yml
+++ b/ci/py3.9.yml
@@ -4,7 +4,6 @@ channels:
 dependencies:
   - python=3.9
   - aiohttp
-  - apache-beam
   - black
   - boto3
   - cfgrib<0.9.9.0
--- a/pangeo_forge_recipes/executors/beam.py
+++ b/pangeo_forge_recipes/executors/beam.py
@@ -52,6 +52,7 @@ class _SingleArgumentStage(beam.PTransform):
         return (
             pcoll
             | "Prepare" >> beam.FlatMap(self.prepare_stage)
+            | beam.Reshuffle()
             | "Execute" >> beam.MapTuple(self.exec_stage)
             | beam.combiners.ToList()
             | "Validate" >> beam.Map(self.post_validate)
--- a/pangeo_forge_recipes/executors/beam.py
+++ b/pangeo_forge_recipes/executors/beam.py
@@ -7,11 +7,11 @@ import apache_beam as beam
 from .base import Config, NoArgumentStageFunction, Pipeline, PipelineExecutor, Stage
 
 
-def _no_arg_stage(last: int, *, current: int, func: NoArgumentStageFunction, config: Config) -> int:
+def _no_arg_stage(last: int, *, current: int, fun: NoArgumentStageFunction, config: Config) -> int:
     """Execute a NoArgumentStageFunction, ensuring execution order."""
     assert (last + 1) == current, f"stages are executing out of order! On step {current!r}."
 
-    func(config=config)
+    fun(config=config)
 
     return current
 
@@ -67,7 +67,7 @@ class BeamPipelineExecutor(PipelineExecutor[beam.PTransform]):
                 pcoll |= stage.name >> _SingleArgumentStage(step, stage, pipeline.config)
             else:
                 pcoll |= stage.name >> beam.Map(
-                    _no_arg_stage, current=step, func=stage.function, config=pipeline.config
+                    _no_arg_stage, current=step, fun=stage.function, config=pipeline.config
                 )
 
         return pcoll
--- a/pangeo_forge_recipes/executors/beam.py
+++ b/pangeo_forge_recipes/executors/beam.py
@@ -1,26 +1,75 @@
-from typing import Optional, Iterable
+import itertools
+from dataclasses import dataclass
+from typing import Optional, Iterable, Tuple, Any, List
 
 import apache_beam as beam
 
-from .base import Pipeline, PipelineExecutor, Config, StageFunction
+from .base import (
+    Pipeline,
+    PipelineExecutor,
+    Config,
+    Stage,
+    NoArgumentStageFunction,
+    SingleArgumentStageFunction
+)
 
 
-def exec_ordered(last: int,
-                 *,
-                 current: int,
-                 function: StageFunction,
-                 config: Config,
-                 mappable: Optional[Iterable]) -> int:
-    """Execute a Pipeline Stage, ensuring execution order."""
-    assert (last + 1) == current, 'stages are executing out of order!'
+def _exec_no_arg_stage(
+    last: int, *, current: int, function: NoArgumentStageFunction, config: Config
+) -> int:
+    """Execute a NoArgumentStageFunction, ensuring execution order."""
+    assert (last + 1) == current, f'stages are executing out of order! On step {current!r}.'
 
-    if mappable is not None:
-        for m in mappable:
-            function(m, config=config)
-    else:
-        function(config=config)
+    function(config=config)
 
     return current
+
+
+def _no_op(arg, config=None) -> None:
+    pass
+
+
+@dataclass()
+class _SingleArgumentStage(beam.PTransform):
+    """Execute mappable stage in parallel."""
+
+    step: int
+    stage: Stage
+    config: Config
+
+    @staticmethod
+    def prepare_stage(last: int, *, current: int, mappable: Iterable) -> Iterable[Tuple[int, Any]]:
+        """Propagate current stage to Mappables for parallel execution."""
+        assert (last + 1) == current, f'stages are executing out of order! On step {current!r}.'
+        return zip(itertools.repeat(current), mappable)
+
+    @staticmethod
+    def exec_stage(
+        last: int, arg,
+        current: int = -1, function: SingleArgumentStageFunction = _no_op, config: Optional[Config] = None
+    ) -> int:
+        """Execute stage function."""
+        assert last == current, f'stages are executing out of order! On step {current!r}.'
+
+        function(arg, config=config)
+
+        return current
+
+    @staticmethod
+    def post_validate(last: List[int], *, current: int) -> int:
+        """Propagate step number for downstream stage validation."""
+        assert all([it == current for it in last]), f'stages are executing out of order! On step {current!r}.'
+        return current
+
+    def expand(self, pcoll):
+        return (
+                pcoll
+                | "Prepare" >> beam.FlatMap(self.prepare_stage, current=self.step, mappable=self.stage.mappable)
+                | "Execute" >> beam.MapTuple(self.exec_stage, current=self.step, function=self.stage.function,
+                                             config=self.config)
+                | beam.combiners.ToList()
+                | "Validate" >> beam.Map(self.post_validate, current=self.step)
+        )
 
 
 class BeamPipelineExecutor(PipelineExecutor[beam.PTransform]):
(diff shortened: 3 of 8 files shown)


alxmrs:
I was going to take some time to update the docs to reflect Apache Beam support in this CL. However, it seems like the docs need a larger overhaul anyway. So, I'll mark this ready for review.

rabernat:
This is fantastic Alex! I just activated the workflows, but I predict they will fail because beam is not in the dev environments. You'll need to add beam [here](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/master/ci/py3.8.yml) and [here](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/master/ci/py3.9.yml).

ci/py3.8.yml
```diff
@@ -4,6 +4,7 @@ channels:
 dependencies:
   - python=3.8
   - aiohttp
+  - apache_beam
```
rabernat, line 7:
```suggestion
  - apache-beam
```

ci/py3.9.yml
```diff
@@ -4,6 +4,7 @@ channels:
 dependencies:
   - python=3.9
   - aiohttp
+  - apache_beam
```
rabernat, line 7:
```suggestion
  - apache-beam
```

rabernat (commented):
I think conda forge wants a dash in the name `apache-beam`: https://github.com/conda-forge/apache-beam-feedstock

rabernat:
I merged my suggested change to the py3.8.yml environment. This turned up an error in the beam executor: https://github.com/pangeo-forge/pangeo-forge-recipes/runs/4138427592?check_suite_focus=true#step:8:2443

alxmrs:
Thanks, I should have caught that locally. Apparently, a variable name I chose to meet the linter created a type check conflict with Beam. `func` may have been used as an internal argument, and had a conflict? Anyways, I changed the parameter name to `fun`, and it meets Beam's type check and the precommit checks, locally.

rabernat:
I think this has now turned up an upstream issue with Xarray. (0.20.0 was just released) I'll make a PR for that fix.

rabernat:
For some reason, the 3.9 conda environment was taking over an hour to build. I'm restarting the build.

rabernat:
For some reason, the python 3.9 environment seems to be unsolvable. https://github.com/pangeo-forge/pangeo-forge-recipes/runs/4143328072?check_suite_focus=true#step:5:302

There must be a version conflict somewhere, but I don't know how to debug it. Any ideas?

TomAugspurger:
mamba tends to give better error messages when there are conflicts.

That said, mamba solves fine for me locally (linux). Maybe try using mamba if you aren't already? https://github.com/conda-incubator/setup-miniconda#example-6-mamba

rabernat:
> Maybe try using mamba if you aren't already?

A PR to switch the CI to Mamba would be most welcome!

alxmrs:
> For some reason, the python 3.9 environment seems to be unsolvable.

It just occurred to me: Apache Beam only officially supports Python 3.8 -- it doesn't yet have support for 3.9, though it's on the roadmap. 

- https://beam.apache.org/get-started/quickstart-py/
- https://issues.apache.org/jira/browse/BEAM-12000

rabernat:
If you merge master (now that #226 is in), the tests should pass here.

tests/test_pipelines.py
```diff
@@ -6,6 +6,7 @@
 from pytest_lazyfixture import lazy_fixture
 
 from pangeo_forge_recipes.executors.base import Pipeline, Stage
+from pangeo_forge_recipes.executors.beam import BeamPipelineExecutor
```
rabernat, line 9: Since beam will be an optional dependency, we need some sort of [importorskip](https://docs.pytest.org/en/latest/reference/reference.html#pytest-importorskip-ref) logic in this test.

We don't want to skip the whole module if beam is not importable--just the beam part.

alxmrs: TIL about this pytest feature! I'll update the tests.

alxmrs: I ended up taking a different approach to achieve the same thing. PTAL.

pangeo_forge_recipes/executors/beam.py
```diff
@@ -0,0 +1,79 @@
...
+from typing import Any, Iterable, List, Tuple, cast
+
+import apache_beam as beam
+
+from .base import Config, NoArgumentStageFunction, Pipeline, PipelineExecutor, Stage
+
+
+def _no_arg_stage(last: int, *, current: int, fun: NoArgumentStageFunction, config: Config) -> int:
+    """Execute a NoArgumentStageFunction, ensuring execution order."""
+    assert (last + 1) == current, f"stages are executing out of order! On step {current!r}."
```
rabernat, line 12: I like these checks to ensure that the stages are happening in the right order.

pangeo_forge_recipes/executors/beam.py
```diff
@@ -0,0 +1,79 @@
...
+        in_current_step = all((it == self.step for it in last))
+        assert in_current_step, f"stages are executing out of order! On step {self.step!r}."
+
+        return self.step
+
+    def expand(self, pcoll):
+        return (
+            pcoll
+            | "Prepare" >> beam.FlatMap(self.prepare_stage)
+            | "Execute" >> beam.MapTuple(self.exec_stage)
```
rabernat, line 55: @shoyer added a `shuffle` step here in rechunker for reasons I don't fully understand

https://github.com/pangeo-data/rechunker/blob/73ef80b4ec9dfb980988460f7d5290076d4835e2/rechunker/executors/beam.py#L69-L71

tests/test_pipelines.py
```diff
@@ -5,11 +5,21 @@
...
 
+# Only run Beam executor tests if Beam is in the current test environment.
+try:
+    from pangeo_forge_recipes.executors.beam import BeamPipelineExecutor
+except (ImportError, AttributeError):
+
+    class BeamPipelineExecutor(PipelineExecutor[None]):
+        @staticmethod
+        def compile(pipeline: Pipeline) -> None:
+            pytest.skip()
```
rabernat, line 21: I'm not wild about this way of getting to skip. It would be nice to have a solution that doesn't know anything about how to implement the class.

What if we used this pattern from pytest: https://docs.pytest.org/en/6.2.x/example/parametrize.html#indirect-parametrization-of-optional-implementations-imports ?

Redefine `Executor` as a parametrized fixture (rather than a parametrized argument). Something like

```python
@pytest.fixture(
    scope="session",
    params=[
        ("pangeo_forge_recipes.executors.dask", "DaskPipelineExecutor"),
        ("pangeo_forge_recipes.executors.function", "FunctionPipelineExecutor"),
        ("pangeo_forge_recipes.executors.prefect", "PrefectPipelineExecutor"),
        ("pangeo_forge_recipes.executors.beam", "BeamPipelineExecutor"),
    ]
)
def Executor(request):
    module = pytest.importorskip(request.param[0])
    return getattr(module, param[1])
```
(disclaimer: untested)

There may be an even simpler way to do it...

rabernat (commented):
Looks great Alex! Just one place I'd like to tighten up.

pangeo_forge_recipes/executors/beam.py
alxmrs: I understand why -- and this is a good suggestion. The main reason is that it helps tell tells beam not to fuse operations, which allows the autoscale algorithm to split the work across multiple workers.

tests/test_pipelines.py
alxmrs: Thanks for the suggestion. I'm not too familiar with `pytest`, and this worked with minor adjustments. I'm not sure if I know of a simpler method.

pangeo_forge_recipes/executors/beam.py
```diff
@@ -0,0 +1,80 @@
...
+        in_current_step = all((it == self.step for it in last))
+        assert in_current_step, f"stages are executing out of order! On step {self.step!r}."
+
+        return self.step
+
+    def expand(self, pcoll):
+        return (
+            pcoll
+            | "Prepare" >> beam.FlatMap(self.prepare_stage)
+            | beam.Reshuffle()
```
shoyer, line 55: I would suggest comparing this to just calling `beam.Create()` to create an in-memory collection of tasks. In my experience this can be a little more more efficient than always shuffling inputs, which can add a lot of overhead from anything being written to distributed disk.

If you use beam.Create, you can still ensure the ordering between stages by using a side input, like what I do here in xarray-beam: https://github.com/google/xarray-beam/blob/ebfdbf0275d226d86d4abb6cd47e46307c4f4036/xarray_beam/_src/zarr.py#L236

alxmrs: Thanks for calling out that there will be overhead with writing to disk in the shuffle operation. However, I don't understand how `Create` could fit in here, and am not sure how the xarray-beam example can be applied. 

I notice that when I expand the pcoll with a `Create` step, I get an assertion error that the previous step is not a `PBegin`. This makes sense to me: Creates will only work when applied to the root of the pipeline. Maybe the side-input technique solves this problem, but I'm missing exactly how.

alxmrs: Looking at the definition of `Create`, I notice that it also implements the FlatMap...Reshuffle pattern found in the above implementation:
```
  def expand(self, pbegin):
    assert isinstance(pbegin, pvalue.PBegin)
    serialized_values = [self._coder.encode(v) for v in self.values]
    reshuffle = self.reshuffle # this is true by default

    # Avoid the "redistributing" reshuffle for 0 and 1 element Creates.
    # These special cases are often used in building up more complex
    # transforms (e.g. Write).

    class MaybeReshuffle(PTransform):
      def expand(self, pcoll):
        if len(serialized_values) > 1 and reshuffle:
          from apache_beam.transforms.util import Reshuffle
          return pcoll | Reshuffle()
        else:
          return pcoll

    return (
        pbegin
        | Impulse()
        | FlatMap(lambda _: serialized_values).with_output_types(bytes)
        | MaybeReshuffle().with_output_types(bytes)
        | Map(self._coder.decode).with_output_types(self.get_output_type()))
```

tests/test_pipelines.py
```diff
@@ -50,9 +47,23 @@ def func1(arg, config=None):
...
+    params=[
+        ("pangeo_forge_recipes.executors.dask", "DaskPipelineExecutor"),
+        ("pangeo_forge_recipes.executors.function", "FunctionPipelineExecutor"),
+        ("pangeo_forge_recipes.executors.prefect", "PrefectPipelineExecutor"),
+        ("pangeo_forge_recipes.executors.beam", "BeamPipelineExecutor"),
+    ],
 )
+def Executor(request):
+    try:
+        module = pytest.importorskip(request.param[0])
```
rabernat, line 61: Sorry to keep nit-picking this section. Last time, I promise...

```suggestion
        module = importlib.import_module(request.param[0])
```

It seems redundant to use `importorskip` and also `pytest.skip`. (This will require adding `importlib` to the imports above.

pangeo_forge_recipes/executors/beam.py
rabernat: I don't understand enough about beam to know whether this conversation has been resolved.

rabernat (commented):
Ok, great! We are almost there. Just a few tiny comments then I promise we can stop iterating and merge this.

pangeo_forge_recipes/executors/beam.py
shoyer: I'm convinced by @alxmrs's argument here.

We'll still have to see how well things work in practice, but theoretically I see no reason why this work any differently from what I was suggesting :+1:

tests/test_pipelines.py
alxmrs: No problem, this nit makes sense to me.

Issue #227, opened 2021-11-08
Debugging memory issues in IMERG


TomAugspurger:
Dropping a few notes here for the data:

@sharkinsspatial and I are seeing some slowness and high memory usage on the client when running the imerg recipe. This has lots of steps in the time dimension.

I've traced something that looks like a memory leak to `azure.storage.blob`. This script just puts a bunch of 24-bytes objects:

<details>

```python
# file: async_mem.py
import os
import time
import asyncio

import azure.storage.blob.aio

# N = 10_000

data = b"\x02\x013\x08\x08\x00\x00\x00\x08\x00\x00\x00\x18\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"  # 0

# Obervation: we don't see a memory leak when just creating the bc; Actually have to put data.

async def put(acc, i):
    async with acc.get_blob_client(f"async/{i}") as bc:
        await bc.upload_blob(data=data, overwrite=True, metadata={"is_directory": "false"})



async def main():
    conn_str = os.environ["AZURE_CONNECTION_STRING"]
    async with azure.storage.blob.aio.ContainerClient.from_connection_string(conn_str, "gpm-imerg-debug") as acc:
        tracemalloc.start()

        coros = (put(acc, i) for i in range(N))

        t0 = time.time()
        await asyncio.gather(*coros)
        t1 = time.time()

        print(f"processed {N} blobs in {t1 - t0:.2f}s")


if __name__ == "__main__":
    asyncio.run(main())
```

</details>

![mem](https://user-images.githubusercontent.com/1312546/140832567-3cee1109-3803-47a4-9531-c1e90e2034de.png)

I don't see an easy workaround (I also don't really understand where the memory leak might be, and hopefully won't need to find out).

---

So making a bunch of small writes isn't great for performance, and it turns out memory for Azure. Can we avoid the small writes in the first place?

One of the batches of small writes happens at https://github.com/pangeo-forge/pangeo-forge-recipes/blob/78a274b4cceb4bc1734cddc48fa9c58f553f7684/pangeo_forge_recipes/recipes/xarray_zarr.py#L109-L112

That writes `0` for each chunk in the concat dim (`time` typically). In the case of IMERG, that's ~360,000 writes. @rabernat mentioned this was to avoid an issue with `decode_times`. For better or worse, the test suite passes with those lines removed...

[...]

PR #229, opened 2021-11-09
Consolidate concat_dim earlier in prepare_target

TomAugspurger:
Closes https://github.com/pangeo-forge/pangeo-forge-recipes/issues/227
(maybe, might be worth testing).

A few notes:

1. I'm not really happy with how I'm handling the locking (see https://github.com/pangeo-forge/pangeo-forge-recipes/pull/229/files#diff-e12c886cc124886c5cfa5313d760a36c39649af9da845077c663e6feab8487b5R245-R253). I just kind of overwrite what the "real" lock determination code does. This feels too special-casey, but I haven't dug into the how that conflict code works.
2. I haven't tested this at scale yet. Maybe @sharkinsspatial can help verify that this actually fixes things? Let me know if you're able to, or if I should get something set up.

### DIFF
--- a/pangeo_forge_recipes/recipes/xarray_zarr.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr.py
@@ -91,7 +91,12 @@ def inputs_for_chunk(
     return input_keys
 
 
-def expand_target_dim(target: FSSpecTarget, concat_dim: Optional[str], dimsize: int) -> None:
+def expand_target_dim(
+    target: FSSpecTarget,
+    concat_dim: Optional[str],
+    dimsize: int,
+    consolidate_dimension_coordinates: bool = True,
+) -> None:
     target_mapper = target.get_mapper()
     zgroup = zarr.open_group(target_mapper)
     ds = open_target(target)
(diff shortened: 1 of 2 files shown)


pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -549,20 +571,20 @@ def _gather_coordinate_dimensions(group: zarr.Group) -> List[str]:
...
-        raise ValueError("target has not been set.")
-
-    if config.consolidate_dimension_coordinates:
-        logger.info("Consolidating dimension coordinate arrays")
-        target_mapper = config.target.get_mapper()
-        group = zarr.open(target_mapper, mode="a")
-        # https://github.com/pangeo-forge/pangeo-forge-recipes/issues/214
-        # intersect the dims from the array metadata with the Zarr group
-        # to handle coordinateless dimensions.
+def _consolidate_dimension_coordinate(
```
TomAugspurger, line 574: The diff here looks strange.

All I did was

1. move the `if config.consolidate_dimension_coordinates` to its own function so that I could call it from two places (here and earlier in `prepare_target` for `concat_dim`)
2. Let the caller pass in the `dims` to check (so that we can do just `concat_dim` early on, and the other dims later; I think this is what we want).
3. Added a check that `arr.chunks != arr.shape (to avoid rewriting `concat_dim` in `finalize_target`, since it would already be consolidated)

sharkinsspatial:
@TomAugspurger I've been locked up on other things today but I'll try to run some testing with this tomorrow 👍

TomAugspurger:
As discovered by @sharkinsspatial, this implementation buggy (sorry). When we call `resize`, we end up with a "fragmented" array that has 360,000 entries and a chunksize of 1. So it's already too late. I'll explore our options (and the CI failure)

rabernat:
@TomAugspurger - over in #236, I have removed the pre-initialization of the concat dim with 0 values. It doesn't seem to break anything (including the tutorial notebooks). If that is the main bottleneck, I would prefer that (simpler) solution over this (more complex) one.

I think the real fix for #227 is upstream in fsspec, related by supporting `batch_size` more broadly across async operations.

rabernat:
p.s. If you plan to keep working on this branch, please rebase as there have been some big improvements to the test suite.

cisaacstern:
Closing as stale/out-of-date. Thanks all!

PR #231, opened 2021-11-11
Use mamba to install env in test worklow

### LINKED ISSUE #230: cache-conda doesn't seem to be working in github workflows
See e.g. https://github.com/pangeo-forge/pangeo-forge-recipes/runs/4158825199?check_suite_focus=true#step:4:9

```g u
Run actions/cache@v1
  with:
    path: ~/conda_pkgs_dir
    key: Linux-conda-0-
  env:
    PYTEST_ADDOPTS: --color=yes
    pythonLocation: /opt/hostedtoolcache/Python/3.8.12/x64
    CACHE_NUMBER: 0
Cache not found for input keys: Linux-conda-0-.
```

As a result (I think), we are spending 4 minutes setting up environments in every workflow.

rabernat:
Fixes #230 

...maybe?

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -39,6 +39,7 @@ jobs:
       - name: setup miniconda
         uses: conda-incubator/setup-miniconda@v2
         with:
+          mamba-version: "*"
           activate-environment: pangeo-forge-recipes
           environment-file: ci/py${{ matrix.python-version }}.yml
           python-version: ${{ matrix.python-version }}
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -29,13 +29,13 @@ jobs:
           python-version: ${{ matrix.python-version }}
           architecture: x64
       - name: Cache conda
-        uses: actions/cache@v1
+        uses: actions/cache@v2
         env:
           # Increase this value to reset cache if ci/py${{ matrix.python-version }}.yml has not changed
           CACHE_NUMBER: 0
         with:
           path: ~/conda_pkgs_dir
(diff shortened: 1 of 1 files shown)


cisaacstern:
🙏 🎉 will save a lot of time in testing when this works; last I checked, building the conda env was ~50% of the test time 😅

rabernat:
I'm not quite sure how much time this will shave off. I think it only applies to the downloading of packages. 

Where I am going with this is that I want to start building the tutorial notebooks in CI as a sort of integration test (see comment in https://github.com/pangeo-forge/pangeo-forge-recipes/issues/227#issuecomment-964318852). But I want to make the env setup faster first.

If this works as I hope, in my next commit I'll try mamba.

rabernat:
Well caching did not appear to speed things up, but that might have something to do with the combination of mamba. I'm gonna try disabling mamba for a minute.

rabernat:
Here are the results on the time spent by the "setup miniconda" step of the workflow

| config | time |
| -- | -- |
| [no cache + conda](https://github.com/pangeo-forge/pangeo-forge-recipes/runs/4181929885?check_suite_focus=true#step:5:1) | 4:57 |
| [cache + mamba](https://github.com/pangeo-forge/pangeo-forge-recipes/runs/4182087012?check_suite_focus=true#step:5:1) | 4:05 |
| [cache + conda](https://github.com/pangeo-forge/pangeo-forge-recipes/runs/4182263417?check_suite_focus=true#step:5:1) | 5:04 |

The good news is that mamba speeds things up. The bad news is that caching appears to have no effect.

I'm going to back up to e998912 and merge this. But I really wish we could shave down that 4 min.

PR #234, opened 2021-11-14
New workflow for tutorial building

rabernat:
This will cause all our tutorial notebooks to be run in CI.

PR #236, opened 2021-11-15
remove pre-initialization of concat-dim

rabernat:
This PR will test the new slash command workflows introduced in #234.

### DIFF
--- a/pangeo_forge_recipes/recipes/xarray_zarr.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr.py
@@ -106,11 +106,6 @@ def expand_target_dim(target: FSSpecTarget, concat_dim: Optional[str], dimsize: 
         logger.debug(f"resizing array {v} to shape {shape}")
         arr.resize(shape)
 
-    # now explicity write the sequence coordinate to avoid missing data
-    # when reopening
-    if concat_dim in zgroup:
-        zgroup[concat_dim][:] = 0
-
 
 def open_target(target: FSSpecTarget) -> xr.Dataset:
     return xr.open_zarr(target.get_mapper())


cisaacstern:
If the notebooks are run in CI, why is there a slash command?

rabernat:
> If the notebooks are run in CI, why is there a slash command?

Because I don't want the notebooks tests to run _every time you push to a PR_. I want them to run on command.

rabernat:
This is now basically working right but reveals two problems.

- The details url is wrong. This is apparently a known issue (see https://github.com/LouisBrunner/checks-action/issues/18#issuecomment-969384672 and https://github.community/t/checks-api-creating-new-check-run-can-not-set-details-url/152651)
- The text `pre-commit` is appended in front of the check name. I have no idea why.

cisaacstern:
> The text pre-commit is appended in front of the check name. I have no idea why.

🙃 ... IIUC strings w/out special characters don't need enclosing quotes in YAML but the [action example uses them](https://github.com/LouisBrunner/checks-action/blob/3d24d4813a797720cc4e2080a50bdafb3373aef1/.github/workflows/examples.yml#L1) for `name:` and seems [to get the desired result](https://github.com/LouisBrunner/checks-action/pull/1), whereas this PR [does not](https://github.com/rabernat/pangeo-forge-recipes/blob/78ac0ac49c8a713d27e597335ad7602aa1124adf/.github/workflows/tutorials.yaml#L1) 🤷

rabernat:
/run-test-tutorials

Trying with change to bot permissions from  9039dbf (recommended in https://github.com/LouisBrunner/checks-action/issues/18#issuecomment-970312052)

PR #239, opened 2021-11-18
Add marks to executors to allow selection and parallelization

rabernat:
Hopefully this will make our test suite go a lot faster by parallelizing execution across the different executors.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -65,18 +65,18 @@ jobs:
         pytest-mark: ["no_executor", "executor_python", "executor_dask", "executor_prefect", "executor_prefect_dask"]
     steps:
       - uses: actions/checkout@v2
-      - name: 🔁 Setup Python
-        uses: actions/setup-python@v2
-        with:
-          python-version: ${{ matrix.python-version }}
-          architecture: x64
-      - name: 🔁 Setup Miniconda
-        uses: conda-incubator/setup-miniconda@v2
-        with:
-          miniforge-variant: Mambaforge
-          miniforge-version: latest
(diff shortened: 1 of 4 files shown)


rabernat:
Wow this is amazing! 🤩

cisaacstern:
2 minutes! 🤯 
This is such a cool pytest feature!!

rabernat:
I think this is uncontroversial, so I'm just going to merge it.

cisaacstern:
@rabernat, thanks for the helpful in-line comments and docs additions on this.

> parallelizing execution across the different executors

I was going to ask how this part works, but then realized it's via assigning the pytest marks to a matrix. I guess the GitHub docs do imply this, by saying that a `strategy.matrix`defines ["different job configurations"](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) and elsewhere that ["Jobs run in parallel by default. "](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobs), but they're not as explicit in the `strategy.matrix` section as I might expect in emphasizing something like "matrices are a great way to parallelize your workflows." (Cf. [this blog](https://hanxiao.io/2021/01/24/Speedup-CI-Workflow-in-Github-Actions-via-Strategy-Matrix/#building-job-matrix-for-parallelization).)

(And of course, once I thought about the fact that our previous jobs ran in parallel, this became obvious.)

Thought I'd just note these links because this did confuse me for a moment, and perhaps someone else reading this will have the same question. I don't think it needs to be added to our docs, though, as its more about GitHub Actions in general than our use case specifically.

PR #238, opened 2021-11-16
Add generator executor

rabernat:
This PR adds a `.to_generator()` method to the `BaseRecipe` class. It allows one to iterate through a recipe as follows:

```python
rec = MyRecipe()
for function, args, kwargs in rec.to_generator():
    function(*args, **kwargs)
```

This will eventually replace the explicit mode of manual recipe execution and open the door to more flexible recipe patterns (e.g. #224).

PR #240, opened 2021-11-18
Prevent operator fusion for every stage of the Beam executor.

alxmrs:
In an end-to-end experiment, I noticed that stages were still being fused. This caused errors to occur, namely that the Zarr metadata file was not created before one of the later steps (I think `store_chunk`) that needed it.

By adding a `beam.Reshuffle()` step to the end of every stage, we ensure that fusion doesn't occur and that stages do execute serially.

### DIFF
--- a/pangeo_forge_recipes/executors/beam.py
+++ b/pangeo_forge_recipes/executors/beam.py
@@ -70,6 +70,7 @@ class BeamPipelineExecutor(PipelineExecutor[beam.PTransform]):
                 pcoll |= stage.name >> beam.Map(
                     _no_arg_stage, current=step, fun=stage.function, config=pipeline.config
                 )
+            pcoll |= f"Reshuffle{step:03d}" >> beam.Reshuffle()
 
         return pcoll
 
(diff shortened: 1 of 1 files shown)


rabernat:
Thanks a lot for this PR Alex! 🎉 

> In an end-to-end experiment, I noticed that stages were still being fused.

It's great you caught this. It also exposes the fact that the Beam executor is not fully covered by the tests. Over in #238, I have been refactoring the test suite so that all of the recipe integration tests will get run with all the executors, eventually including beam. I'd like to finish that first and see if it catches this same bug.

If so, we can merge #238 first and then you you can rebase this, thereby showing that it fixes the bug.

If not, we will need to add a test for this.

So stand by for a few days.

rabernat:
It looks like #238 did NOT catch the bug related to task ordering and fusion with the Beam executor.

Therefore it would be great to add to or modify the pipeline tests in such a way that we defend against this potential error.

alxmrs:
> Therefore it would be great to add to or modify the pipeline tests in such a way that we defend against this potential error.

I've been investigating ways to do this. Fusion is an implementation details of the Beam Runner (type of executor). [The documentation] on this subject suggests that the proper way to test these details is to deploy a lo-fi pipeline on GCP within CI. To check if the Dataflow service fuses steps, we can validate JSON returned from a CLI, for example. Is this an acceptable approach to take? 

There's a contingency I'm also considering: Updating the assert system in place to break the pipeline if order properties are violated, maybe by using [stateful processing]. 


[The documentation]: https://cloud.google.com/architecture/building-production-ready-data-pipelines-using-dataflow-developing-and-testing#testing_your_pipeline
[stateful processing]: https://beam.apache.org/blog/stateful-processing/

Issue #241, opened 2021-11-19
Check that metadata is already cached before re-computing to prevent local file copy.


alxmrs:
I have a `XarrayZarrRecipe` with the flag `copy_input_to_local_file=True` set. I have been debugging a pipeline while running it end-to-end, and thus re-run partially finished pipelines. 

I notice that in the `cache_input` stage, the `cache_file` step checks if the cached file exists (and is the right size) to skip re-computing. This checking is not done for the `cache_metadata` step. As a result, a stage that could be computed quickly takes a long time, since it has to copy the input file locally first. 

Proposed fix: On line ~165 in the `config.cache_metadata` block, add an if statement to check that the key is already in the `config.metadata_cache` object before re-computing.

alxmrs:
Implemented in #243.

Issue #224, opened 2021-10-25
Use more granular stages for XarrayZarrRecipe


rabernat:
Here is the current "compiler" for the XarrayZarrRecipe

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/0abcaa2438f7ba5966cc49a55f72470d551bec70/pangeo_forge_recipes/recipes/xarray_zarr.py#L586-L593

After #219, we are no longer bound to follow this same exact sequence for every recipe. We could imagine breaking things up instead like

```
     stages = [ 
         Stage(name="cache_input", function=cache_input, mappable=list(recipe.iter_inputs())), 
         Stage(name="cache_metadata", function=cache_metadata, mappable=list(recipe.iter_inputs())), 
         Stage(name="make_fsspec_reference", function=make_fsspec_reference, mappable=list(recipe.iter_inputs())), 
         Stage(name="prepare_target", function=prepare_target), 
         Stage(name="store_chunk", function=store_chunk, mappable=list(recipe.iter_chunks())), 
         Stage(name="finalize_target", function=finalize_target), 
     ] 
```

i.e. adding explicit stages for metadata caching and reference making. We could also add some conditional logic to this compiler to skip certain stages if they are not needed.

jbusecke:
@rabernat I just wanted to point out that #355 might be a relevant use case for more granular stages as proposed here.

cisaacstern:
The modularity of the Apache Beam refactor solves the issues discussed here! 🎉 So closing.

Issue #173, opened 2021-07-21
Add unit tests for recipe functions


rabernat:
As discussed in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/171#issuecomment-883546066, the recent refactoring of the XarrayZarrRecipe class has made it more modular, in the sense that most of the key functions can be run without instantiating a big Recipe class and executing the whole recipe. This means we can have [unit tests](https://en.wikipedia.org/wiki/Unit_testing) for all of these individual functions to verify that they do exactly what is expected of them. This would complement the more holistic integration-style tests we currently have for XarrayZarrRecipe. It might also provide a path for reducing our test suite cost while maintaining high coverage. (It currently takes > 10 minutes to run all the tests.)

cisaacstern:
Just noting that I'm *very* interested in expanding my testing knowledge/competency by working on this. If anyone else gets started on it before I do, please ping me as I'd love to help out with it.

PR #243, opened 2021-11-19
Skip re-computing metadata cache.

alxmrs:
A fix for #241.


cisaacstern:
xref https://github.com/pangeo-forge/pangeo-forge-recipes/issues/224

rabernat:
@alxmrs - I think this change makes sense.

To move forward, we would need two tests:
- A test for the `__contains__` function in https://github.com/pangeo-forge/pangeo-forge-recipes/blob/master/tests/test_storage.py
- A unit test for this re-computing feature. This is harder to test because our recipes tests are mostly integration-style tests which execute the whole recipe from start to finish.  I don't think it makes sense to re-execute an entire recipe just to test this feature.  Since the big refactor, it would now be much easier to implement unit tests (see #173), but we have not actually done it yet. Perhaps this could be our first unit test? 😄

rabernat:
Alex, if we can just get a test for `__contains__` I am fine with merging this as is.

alxmrs:
I'm happy to write unit tests for this – I will push a patch for this tomorrow. 

I'm investigating if an additional flag is needed for this feature – should the user be able to set a `recompute_metadata` flag to force overwriting the cache? I may have encountered a need for this in the current project that I'm working on. My current workaround is to just change the path of the metadata cache.

cisaacstern:
>  should the user be able to set a recompute_metadata flag to force overwriting the cache

@jbusecke and I are working on some derived datasets, where the inputs are already in Zarr format on the cloud.

In this case, we would not be caching inputs, and therefore need some way to call `cache_input_metadata` as a standalone Stage of the recipe. This being it's own stage might supersede the need for for a specific `recompute_metadata` flag on the `cache_inputs` stage.

xref https://github.com/pangeo-forge/pangeo-forge-recipes/issues/224

rabernat:
My plan once this is merged is to start refactoring this recipe to have several [optional] standalone stages (#224), rather than just one big `cache_inputs` stage.

Here we have a choice--another option to add to the recipe config for `recompute_metadata`, or instead make the responsibility for clearing the metadata cache live in user land. The former adds more complexity to maintain. The latter requires intervention from the bakery operator who presumably has direct access to the metadata storage and can call `rm` on the recipe metadata cache directory. That's certainly what I would do if I were executing a recipe locally--just clean out the metadata cache dir.

This discussion reminds me a lot of - https://github.com/pangeo-forge/roadmap/issues/29. **Where in our workflow do we allocate and manage temporary storage?** In https://github.com/pangeo-forge/roadmap/issues/29 we discussed this for the target storage and resolved it with https://github.com/pangeo-forge/roadmap/pull/34... But we have not really had that discussion for cache storage or metadata storage. Maybe this is something the recipe orchestrator could manage.

Alex, can I ask _why_ you need to recompute the metadata? This might help figure out the best way to support that need.

alxmrs:
> Alex, can I ask why you need to recompute the metadata? This might help figure out the best way to support that need.

I hit an unrelated error to the metadata cache (I'll probably bring this up in our Monday meeting anyway; see below). I noticed in the error trace that some metadata may be off. I had iterated a bunch on the Recipe parameters as well as the structure of the input data. Since I had cached the metadata previously, I wanted to rule out the possibility that the pipeline was failing due to stale metadata. 

When I recomputed the metadata (by changing the path for temp storage), I noticed that the error trace changed. To be more specific: In the first run, it reported inconsistent sequence lengths across my data in merge dimensions (Data was stored by month, chunked by hour. Across every merge dimension — i.e. groups of variables — the number of hours per month should be the same). In the second run, while the sequence lengths across concat dimensions differed (different # of total hours per month), they were all the same in merge dimensions.

PR #245, opened 2021-11-29
Opener refactor

### LINKED ISSUE #242: Make "inputs" concept more general in order to accommodate data access from APIs
Our discussion today with ECMWF about ERA5 (see https://github.com/pangeo-forge/staged-recipes/issues/92) surfaced the need for Pangeo Forge to be able to extract data from APIs.

### Current Situation in pangeo-forge-recipes

We assume that the recipe "inputs" are either
- Files that can be opened with fsspec
- OpenDAP endpoints

These inputs are generated by a [FilePattern](https://pangeo-forge.readthedocs.io/en/latest/recipe_user_guide/file_patterns.html) object. The file pattern must contain a function which translates "input keys" into a **string**. This string is then passed either to fsspec or directly to xarray.

The FilePattern interfaces directly with the recipe as follows:
```python
fp = FilePattern(...)
rec = Recipe(fp, ...)
```

The fact that the FilePattern returns a string is a limitation.

### Examples of data sources that don't fit with this model

- Data from the ECMWF [CDS API](https://cds.climate.copernicus.eu/api-how-to) -> The API ultimately gives you a netCDF file based on various query parameters
- MITgcm data opened with xmitgcm, e.g [ECCO data portal](https://medium.com/pangeo/petabytes-of-ocean-data-part-1-nasa-ecco-data-portal-81e3c5e077be) -> returns an Xarray Dataset directly

### One Solution: An additional "Opener" layer

Rather than passing the FilePattern directly to the Recipe and hoping that the Recipe knows what to do with it, we could imagine the following sort of pattern

```python
fp = FilePattern(...)
opener = XarrayFsspecOpener()
inputs = Inputs(fp, opener)
recipe = Recipe(inputs, ...)
```

Basically the "opener" would be guaranteed to return certain things that the recipe could use, for example:
- File-like objects that could be cached
- Xarray datasets that could be used in the recipe

The `Inputs` interface would be pretty light, something like

```python
@dataclass
class Inputs
    file_pattern: FilePattern
    opener: Opener

    def __getitem__(self, input_key):
        thing_to_open = self.fp[input_key]
        return self.opener.open(thing_to_open)
```

The FilePattern could then return whatever it wants, provided that the opener knows what to do with it. For example, FilePattern could return a dict of parameters to pass to an API, or any other Python object that is useful to the opener.

We could imagine providing some default openers
- XarrayFsspecOpener
- XarrayOpendapOpener
- PandasOpener
- etc.

... plus also allowing users to define their own custom openers to handle more specialized situations.

We could also consider making some of these mixins or using inheritance, such that we could do something like

```python
class FsspecOpener:
    def open_file(key):
        ...

class XarrayFsspecOpener(FsspecOpener):
    def open_dataset(key):
        with self.open_file(key) as f:
            yield xr.open_dataset(f)
```

etc.

 ### Pros and Cons

This would be a significant refactor. In the end, code would end up leaving XarrayZarrRecipe and moving to a new class (XarrayFsspecOpener). Overall I think this would make for better separation of concerns and more reuseability of code.

I am concerned that this would create yet another layer of complexity for the users. This could be mitigated by creating some convenience functions or adaptors that would reduce the amount of boilerplate that would need to be written.

rabernat:
Eventually fixes #242.

Goes on top of #238.

### DIFF
--- a/pangeo_forge_recipes/openers/base.py
+++ b/pangeo_forge_recipes/openers/base.py
@@ -1,0 +1,14 @@
+from abc import ABC, abstractmethod
+from typing import TypeVar, Generic, Generator
+
+T_in = TypeVar('T_in')
+T_out = TypeVar('T_out')
+
+
+class BaseOpener(ABC, Generic[T_in, T_out]):
+
+    @abstractmethod
+    def open(input: T_in) -> Generator[T_out, None, None]:
+        """Should be a Contextmanager which yields the output type
+        """
+        pass
--- a/pangeo_forge_recipes/openers/base.py
+++ b/pangeo_forge_recipes/openers/base.py
@@ -1,14 +1,15 @@
 from abc import ABC, abstractmethod
-from typing import TypeVar, Generic, Generator
+from typing import ContextManager, Generic, TypeVar
 
-T_in = TypeVar('T_in')
-T_out = TypeVar('T_out')
+T_in = TypeVar("T_in")
+T_out = TypeVar("T_out")
 
 
+# How to make an ABC with a contextmanager method
+# https://stackoverflow.com/a/69674704/3266235
 class BaseOpener(ABC, Generic[T_in, T_out]):
-
     @abstractmethod
-    def open(input: T_in) -> Generator[T_out, None, None]:
+    def open(self, input: T_in) -> ContextManager[T_out]:
         """Should be a Contextmanager which yields the output type
         """
         pass
--- a/pangeo_forge_recipes/openers/fsspec.py
+++ b/pangeo_forge_recipes/openers/fsspec.py
@@ -1,0 +1,152 @@
+import time
+from contextlib import contextmanager
+from dataclasses import dataclass, field
+from urllib.parse import parse_qs, urlencode, urlparse, urlunparse
+from typing import Optional, Generator
+import logging
+import os
+import tempfile
+
+import fsspec
+from fsspec.implementations.http import BlockSizeError
+
+from .base import BaseOpener
+from ..storage import CacheFSSpecTarget
+
+logger = logging.getLogger(__name__)
+
+def _add_query_string_secrets(fname: str, secrets: Optional[dict]) -> str:
+    if secrets is None:
+        return fname
+    parsed = urlparse(fname)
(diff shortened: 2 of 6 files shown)


rabernat:
As I mentioned at our latest meeting, I'm having some challenges around this refactor because our current "opener" mixes together file opening and caching. Caching is essentially a side effect which obscures the flow of data.

I am starting to wonder if we should continue to refactor things so that data only flows in one direction.

Imagine in we could create Recipes like this, in extreme pseudocode

```python
source = Source(file_pattern, opener, **options)
destination = ZarrDestination(storage_target, **options)

recipe = Recipe(source, destination)
```

This would create a one way flow of data from `source` ➡️  `destination`.

Now to add caching we could do the following

```python
cache = FileCacheDestination(cache_target)

recipe = Recipe(source, cache, destination)
```

This would create a one way flow of data from `source` ➡️  `cache` ➡️ `destination`.

This implies that any valid "destination" would also be a valid "source" for the next stage. In theory we could keep chaining steps together to build more complicated pipelines. For example.

```python
combine = CombineBlocks(time=10)

recipe = Recipes(source, cache, combine, destination)
```

Just starting to think through the implications of this model. A key question is: _What are the basic interface for one of the stages in a recipe? What methods / attributes are implemented by all of `source`, `cache`, `combine`, `destination`?_

Possible answers:
- Each has to have something equivalent to a `FilePattern` which allows the next stage to effectively iterate through a known number of steps. For the final step, it would just be a single item, the Zarr store itself.
- Each stage has to know what type of thing it will produce: a file / url, an Xarray dataset, etc.

Can each stage effectively ignore everything other than the most proximate previous stage?

TomNicholas:
Having spoken to @cisaacstern , I think that this refactor might be needed in order for @robertpincus to use [datatree](https://github.com/xarray-contrib/datatree) to open all the groups in [his NASA data](https://github.com/pangeo-forge/staged-recipes/issues/125#issuecomment-1106816417) in one function call.

cisaacstern:
This has been superseded by the beam refactor, so closing.

Cool to see how this work informed the opener transforms there!

PR #247, opened 2021-11-29
Remove BlockSizeError following fsspec 2021.11.1 release

### LINKED ISSUE #246: Missing BlockSizeError in fsspec 2021.11.1 breaks pangeo-forge-recipes import
The HTTP `BlockSizeError` was [removed in fsspec 2021.11.1](https://github.com/fsspec/filesystem_spec/commit/2de1b53646b23d989902017e916c5a57397895a8). This breaks pangeo-forge-recipes because we import that error here

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/6b44109c663eae2d494afa54c56d127dc33a6058/pangeo_forge_recipes/storage.py#L16

The short term workaround is to run pangeo-forge-recipes with `fsspec==2021.11.0`.

@martindurant, how would we replicate this error handling

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/6b44109c663eae2d494afa54c56d127dc33a6058/pangeo_forge_recipes/storage.py#L41-L47

with the latest fsspec release? Your message on the above-linked `Remove BlockSizeError` commit reads 

> Does *not* allow random access from not the start of the download

so I think we still need this type of error handling?

cisaacstern:
Closes #246

As noted in https://github.com/pangeo-forge/pangeo-forge-recipes/issues/246#issuecomment-981833344, BlockSizeError handling is no longer required following the release of fsspec 2021.11.1.

This PR removes it. No testing changes are proposed because the BlockSizeError is [not covered in the existing tests](
https://app.codecov.io/gh/pangeo-forge/pangeo-forge-recipes/blob/master/pangeo_forge_recipes/storage.py).

### DIFF
--- a/pangeo_forge_recipes/storage.py
+++ b/pangeo_forge_recipes/storage.py
@@ -13,7 +13,6 @@ from typing import Any, Iterator, Optional, Sequence, Union
 from urllib.parse import parse_qs, urlencode, urlparse, urlunparse
 
 import fsspec
-from fsspec.implementations.http import BlockSizeError
 
 logger = logging.getLogger(__name__)
 
@@ -38,13 +37,7 @@ def _copy_btw_filesystems(input_opener, output_opener, BLOCK_SIZE=10_000_000):
             interval = 5  # seconds
             bytes_read = log_count = 0
             while True:
-                try:
-                    data = source.read(BLOCK_SIZE)
-                except BlockSizeError as e:
-                    raise ValueError(
-                        "Server does not permit random access to this file via Range requests. "
-                        'Try re-instantiating recipe with `fsspec_open_kwargs={"block_size": 0}`'
-                    ) from e
+                data = source.read(BLOCK_SIZE)
                 if not data:
                     break
                 target.write(data)


rabernat:
Can you clarify what will happen now if we try to make a range request from a server that does not support range requests? Will it just download the whole file?

That is not the same behavior and could lead to big problems. (E.g. imagine we want 100 MB from a 10 TB file).

martindurant:
You will get 100MB, and the transfer will be terminated at the next socket block (2MB). There is no (foolproof) way to tell whether this 100MB is the range that was actually requested.

We could consider requiring that a range request comes back with the right range header, but this change was meant to fix for the case that X bytes are being read from the start of a file, but the server attempts to send it all.
I am open to suggestions on how to ensure both this fix and also correctly identify and cope with Range-considerate servers. Note that if the file size was never found in the first case, you should get a HTTPStreamFile streaming file, which doesn't know about blocks at all and only reads forwards.

cisaacstern:
> Can you clarify what will happen now if we try to make a range request from a server that does not support range requests? Will it just download the whole file?

The removed code is in `_copy_btw_filesystems`, which AFAICT is always used for caching entire files. 

@martindurant, do I understand correctly that with fsspec `2021.11.1`, even HTTP servers that do not accept random access, will still support [the simplified transfer logic proposed by this PR](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/247/files#diff-1b0c20425676e0f245bdd49a8db3355e0e45139497bf42748276b7fd8705e5b1L40-L50), without requiring the user to explicitly specify `{"block_size": 0}`?

>  (E.g. imagine we want 100 MB from a 10 TB file).

I don't think we ever do this with `_copy_btw_filesystems`? Or perhaps I'm forgetting something.

rabernat:
Looking more closely at the code, I think that this should be okay. We are not using this code path to explicitly select specific ranges but rather to stream the file. It sounds like an HTTPStreamFile will do the right thing.

So LGTM.

martindurant:
I believe all those points are right, but there is probably no downside to setting `{"block_size": 0}` in every case and forcing the "stream" code path.

I will be working an a generic fs-to-fs copy function in fsspec in the coming weeks and referring to these conversations, as well as deriving test cases. Hoping you can just use the upstream one, when it is ready.

cisaacstern:
> there is probably no downside to setting {"block_size": 0} in every case and forcing the "stream" code path

👍 I'll merge without this, as it seems not strictly necessary, but will keep it in mind in case we see any issues arise.

> I will be working an a generic fs-to-fs copy function in fsspec in the coming weeks 

Great news. My stale https://github.com/fsspec/filesystem_spec/pull/723 should probably be closed then. As noted in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/179#issue-964470590, my initial motivation for that was to resolve https://github.com/fsspec/filesystem_spec/issues/641, which I found _was_ indeed possible by (quoting from #179)

> [applying] the internal logic of `ftplib.FTP.retrbinary` to make this work, with the key point being that both reading and writing need to happen within the same (uninterrupted) connection context.

Perhaps then the "FTP server which does not support `abort`" may also be a worthwhile case to keep in mind for this work. A draft of my (albeit rough) solution [is preserved here](https://github.com/cisaacstern/pangeo-forge-recipes/blob/5d2002d99aa3555b4382d77beea4b7f99ef33708/pangeo_forge_recipes/storage.py#L31-L86).

PR #250, opened 2021-11-29
Add upstream dev matrix to tests

### LINKED ISSUE #248: Need to be testing against upstream dev versions
Recent issues like #246 could have been detected proactively if we were testing against upstream dev versions of major dependencies like:
- xarray
- dask
- fsspec

This is easy to configure as another ci environment in our test matrix.

cisaacstern:
Closes #248

Here's an idea for how this might work.

This PR uses a single `upstream-dev.yml` file to amend the existing CI environments to patch each of them with dev versions for the selected packages. (I'm pretty sure this update style works.) It doesn't cache these patched environments (which I think makes sense, as they'll be updating so frequently).

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -106,7 +106,11 @@ jobs:
           - python-version: 3.9
             pytest-mark: "executor_beam"
           - dependencies: "upstream-dev"
-            pytest-mark: ["no_executor", "executor_function", "executor_generator"]
+            pytest-mark: "no_executor"
+          - dependencies: "upstream-dev"
+            pytest-mark: "executor_function"
+          - dependencies: "upstream-dev"
+            pytest-mark: "executor_generator"
     steps:
       - uses: actions/checkout@v2
       - name: 🎯 Set cache number
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -51,7 +51,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.8, 3.9]
-        use-upstream-dev: [true, false]
+        use-upstream-dev: ['true', 'false']
     steps:
       - uses: actions/checkout@v2
       - name: 🔁 Setup Python
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -93,7 +93,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.8, 3.9]
-        use-upstream-dev: [true, false]
+        dependencies: ["releases-only", "upstream-dev"]
         pytest-mark: ["no_executor", "executor_function", "executor_generator",
                       "executor_dask", "executor_prefect", "executor_prefect_dask",
                       "executor_beam"]
@@ -122,8 +122,10 @@ jobs:
       - name: 🎯 Set path to include conda python
         run: echo "/usr/share/miniconda3/envs/pangeo-forge-recipes/bin" >> $GITHUB_PATH
       - name: 🧑‍💻 Maybe use upstream dev versions
-        if: matrix.use-upstream-dev == 'true'
-        run: mamba env update -n pangeo-forge-recipes -f ci/upstream-dev.yml
+        if: matrix.dependencies == 'upstream-dev'
+        run: |
+          mamba env update -n pangeo-forge-recipes -f ci/upstream-dev.yml \
+          && conda list
       - name: 🌈 Install pangeo-forge-recipes package
         shell: bash -l {0}
         run: |
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -106,8 +106,6 @@ jobs:
           - python-version: 3.9
             pytest-mark: "executor_beam"
           - dependencies: "upstream-dev"
-            pytest-mark: "no_executor"
-          - dependencies: "upstream-dev"
             pytest-mark: "executor_function"
           - dependencies: "upstream-dev"
             pytest-mark: "executor_generator"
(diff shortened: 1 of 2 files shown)


cisaacstern:
Hmm looks like this needs a little work, given that the Dask version installed in `prepare-env (3.8, true)` is the same as `prepare-env (3.8, false)`. Will think about this and get back shortly with an update.

rabernat:
I would recommend just creating a separate environment in `ci` and using pip to install the upstream packages. Like xgcm:

https://github.com/xgcm/xgcm/blob/master/ci/environment-upstream-dev.yml

cisaacstern:
> creating a separate environment in ci and using pip to install the upstream packages. Like xgcm

xgcm's CI environment files [aren't python-version specific](https://github.com/xgcm/xgcm/tree/master/ci) (in name or content). Which allows xgcm to build CI environments [with an explicit file name](https://github.com/xgcm/xgcm/blob/c164405ec56405e136bffd1a77bd47bfe505eeb6/.github/workflows/ci.yaml#L115), i.e.

```yaml
environment-file: ci/environment-upstream-dev.yml
```

Whereas pangeo-forge-recipes variablizes the environment file name using the python version matrix

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/53554795404c1358468aeafcbafd9d978cb2f045/.github/workflows/main.yaml#L74

So IIUC following xcgm on this would mean some greater degree of restructuring of our `main.yaml`.

For some reason the [conditional env update](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/250/files#diff-71cabc4177e41ea8f15a89eb65398fa8187739f68a2762706d84885cd8b2ddc3R86) I'm attempting here doesn't seem to be working, but if it were, I was thinking this would be a lighter-touch way of getting the additional testing without digging into our environment file naming convention.

But there may be an easier way to follow xgcm's example than I am realizing.

rabernat:
> But there may be an easier way to follow xgcm's example than I am realizing.

How about this

```yaml
      matrix:
        include:
          - python-version: 3.8
            env-file: ci/py3.8.yml
          - python-version: 3.9
            env-file: ci/py3.9.yml
          - python-version: 3.9
            env-file: ci/py3.9-dev.yml
```

Then refactor the rest of the workflow to use `{matrix.env-file}`.

cisaacstern:
It took a little trial-and-error, but the current solution does appear to work.

Note, e.g., the following package versions for the `prepare-env (3.8, releases-only)` job

- [dask 2021.11.2](https://github.com/pangeo-forge/pangeo-forge-recipes/runs/4375530917?check_suite_focus=true#step:10:95)
- [fsspec 2021.11.1](https://github.com/pangeo-forge/pangeo-forge-recipes/runs/4375530917?check_suite_focus=true#step:10:120)
- [xarray 0.20.1](https://github.com/pangeo-forge/pangeo-forge-recipes/runs/4375530917?check_suite_focus=true#step:10:333)

as compared to these versions in the `prepare-env (3.8, upstream-dev)` job

- [dask 2021.11.2+14.ga5aecac8](https://github.com/pangeo-forge/pangeo-forge-recipes/runs/4375530963?check_suite_focus=true#step:10:95)
- [fsspec 2021.11.1](https://github.com/pangeo-forge/pangeo-forge-recipes/runs/4375530963?check_suite_focus=true#step:10:119) (Seemingly not updated b/c no [commits](https://github.com/fsspec/filesystem_spec/commits/master) there since latest [release](https://pypi.org/project/fsspec/#history).)
- [xarray 0.20.2.dev24+gf0867284](https://github.com/pangeo-forge/pangeo-forge-recipes/runs/4375530963?check_suite_focus=true#step:10:332)

Of course the current approach does bump our job count from 18 to 33 jobs, though these are run in parallel so I don't think overall time is impacted too much.

Ultimately I did end up caching the upstream-dev environment as a way to pass it from `prepare-env` to `run-test`. The upstream-dev dependencies specified in `ci/upstream-dev.yml` are [updated from GitHub on each run](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/250/files#diff-71cabc4177e41ea8f15a89eb65398fa8187739f68a2762706d84885cd8b2ddc3R85-R87), but the rest of the upstream-dev environment is still only refreshed every three days (same as the releases-only envs).

This does depart from the xgcm style you suggested, @rabernat, but due to how baked-in `matrix.python-version` is to our current workflow, I think this is the simplest way to get this testing without a larger refactor of our (recently refactored) workflow. Totally open to a different approach, though, if there is some limitation to this style which I've overlooked.

rabernat:
This seems great Charles! 👍 

Let's also add Prefect to the upstream dev category.

My one question is whether we really want to run the upstream-dev on all executors on every PR. We now have 33 checks! 😱 Some options are:
- To just run the upstream-dev stuff as a cron job
- To add some more entries to `exclude` to reduce the number of tests run.

.github/workflows/main.yaml
```diff
@@ -93,13 +97,20 @@ jobs:
...
+        dependencies: ["releases-only", "upstream-dev"]
         pytest-mark: ["no_executor", "executor_function", "executor_generator",
                       "executor_dask", "executor_prefect", "executor_prefect_dask",
                       "executor_beam"]
         exclude:
           # No beam support in python 3.9
           - python-version: 3.9
             pytest-mark: "executor_beam"
+          - dependencies: "upstream-dev"
+            pytest-mark: "no_executor"
```
rabernat, line 109: I would remove this exclusion. Running the `no_executor` tests is pretty cheap and covers lots of core functionality not covered by the executor tests.

```suggestion
```

PR #252, opened 2021-12-03
Derived CMIP6 data recipe builder (WIP)

jbusecke:
@cisaacstern and I had a really productive hack today and I think we made some good progress towards using pangeo-forge to derive datasets from existing ARCO data. 

## Our Goal
This application of pangeo-forge represents a bit of a deviation from the core or initial mission of migrating legacy datasets into the cloud, but I believe it could really boost the adaptation of "cloud-first" workflows in many [science contexts](https://twitter.com/rabernat/status/1466390458402226185?s=20).  Other attempts have been made to achieve this functionality (https://github.com/pangeo-forge/pangeo-forge-recipes/issues/176, #205), but this represents our most successful effort to date.

We had the goal of producing a derived dataset from arbitrary CMIP6 and decided to start with a weighted mean of a surface variable to keep computation short, not get into too much trouble with dask chunking (averaging over lateral dimension in time chunks parallelizes nicely), but also produce something that would be actually useful in a science context.

## Successes

We added a new `builder` directory containing a module which could be extended to have several generalizable functions that build recipes but use logic that is specific to cmip6 and are inspired by the workflow developed in [cmip6_preprocessing](https://cmip6-preprocessing.readthedocs.io/en/latest/index.html#suggested-workflow)(this provided the advantage of an easy transition from interactive work to a recipe). 

We were able to programmatically build a set of recipes for different variables (sea surface temperature and salinity) and two examples models, execute these locally and plot the resulting data. 

The builder function relies on 'facets' which are used to query an [intake-esm catalog](https://pangeo-data.github.io/pangeo-cmip6-cloud/accessing_data.html#loading-an-esm-collection) that is also queried to get available weights (in this case surface area `areacello`) and find the best match to the data (see [here](https://cmip6-preprocessing.readthedocs.io/en/latest/postprocessing.html#Handling-grid-metrics-in-CMIP6) for details).

We made a little demo [notebook](https://notebooksharing.space/view/929a7075d4056cfc188ca9ff2b1f62b85e38806836b0aed8cbdf52bfa1465c37#displayOptions=) (Big Shoutout to @yuvipanda for the amazing notebooksharing.space 🔥)

### DIFF
--- a/pangeo_forge_recipes/builders/cmip.py
+++ b/pangeo_forge_recipes/builders/cmip.py
@@ -1,0 +1,67 @@
+from typing import List, Optional, Union
+
+from cmip6_preprocessing.postprocessing import match_metrics
+from cmip6_preprocessing.preprocessing import combined_preprocessing
+from cmip6_preprocessing.utils import google_cmip_col
+
+from ..patterns import pattern_from_file_sequence
+from ..recipes import XarrayZarrRecipe
+
+
+def cmip6_weighted_mean_recipe(
+    facets: dict,
+    weight_coord: str,
+    mean_dims: Union[List[str], str],
+    coarsen_time: Optional[int] = None,
+) -> XarrayZarrRecipe:
+
+    # Initialize CMIP6 catalog
+    col = google_cmip_col()
+
+    # Get path to variable data according to facets
+    # Note: might be superceded by opener refactor (custom opener which takes facets as kwargs)
+    cat = col.search(**facets)
+    path_list = cat.df["zstore"].tolist()
+    assert len(path_list) == 1
+    path = path_list[0]
+
+    # Load all available weights to enable imperfect matching, see:
+    # https://cmip6-preprocessing.readthedocs.io/en/latest/postprocessing.html#Handling-grid-metrics-in-CMIP6
+    cat_weight = col.search(
+        variable_id=weight_coord, source_id=facets["source_id"], grid_label=facets["grid_label"],
+    )
+    ddict_weight = cat_weight.to_dataset_dict(
+        zarr_kwargs={"consolidated": True, "use_cftime": True},
+        preprocess=combined_preprocessing,
+        aggregate=False,
+    )
+
+    def process_input(ds, fname):
+        ds = combined_preprocessing(ds)
+
+        ddict = dict(blank=ds)
+        ddict = match_metrics(ddict, ddict_weight, [weight_coord])
+
+        assert len(ddict) == 1
+        _, ds_out = ddict.popitem()
+
+        # TODO: write weight provenance to ds_out attributes here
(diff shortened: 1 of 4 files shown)


cisaacstern:
A few notes on implementation.

Proposed changes in #242 (see also https://github.com/pangeo-forge/pangeo-forge-recipes/issues/242#issuecomment-977323390) may make more complex derivations possible, but we were pleased to see that for  the weighted mean example demonstrated in above-linked notebook, that refactor was actually not strictly necessary. The two issues we did encounter with existing code were:

1. XarrayZarrRecipe assumes inputs are openable with:
https://github.com/pangeo-forge/pangeo-forge-recipes/blob/53554795404c1358468aeafcbafd9d978cb2f045/pangeo_forge_recipes/recipes/xarray_zarr.py#L292

    where `f` is typically an fsspec OpenFile, but can also be a string. AFAICT, the only way to get `f` to be passed as a string here is by setting `is_opendap=True` on the FilePattern instance. We ended up doing this (and commenting out some `is_opendap` related conditionals), because it seems that `xr.open_dataset` doesn't want to open Zarr stores from OpenFiles (even when passing `kw=dict(engine="zarr")`). Maybe I'm missing something here, but passing `f` as a string is our temporary workaround.

2. We are not caching the input for this recipe because, as previously mentioned, it already exists as a Zarr store. As it currently stands, there is no standalone stage for caching metadata (it happens along with `cache_input`). This is fine for our situation if we are not coarsening: we can just open the input dataset and pass `nitems_per_files=len(ds.time)` to the FilePattern. With coarsening, however, this becomes trickier. For this reason, we pulled out some of the internal metadata caching code and put it in our notebook. See https://github.com/pangeo-forge/pangeo-forge-recipes/pull/243#issuecomment-985087322 and @rabernat's response below that for further conversation on this issue.

yuvipanda:
So happy to see it be used in the wild, @jbusecke!

duncanwp:
Hi @jbusecke! This recipe looks exactly what I need for https://github.com/pangeo-forge/staged-recipes/issues/134 but seems to have stalled, is it likely to be merged or is there another approach I should follow?

cisaacstern:
👋 @duncanwp, thanks for checking in here. This issue was an early attempt that @jbusecke and I made to work on this issue of generalizing recipes derived from ESGF holdings. We have since moved on, and simply forgot to close this issue. The current state of these efforts is well-summarized by Julius in https://github.com/pangeo-data/pangeo-cmip6-cloud/issues/31#issuecomment-1196231858. We don't have an end-to-end solution deployed for this today, but have made a lot of headway since this issue, and welcome your collaboration on this. I'll close this issue now, as its gone stale, but please do follow up on the issue I've linked here, or any of the other issues that it links to. Look forward to working together on this!

duncanwp:
Perfect, thanks @cisaacstern! I'll dive in to that and see if / how I can help

PR #255, opened 2021-12-09
Improving the error message of `calculate_sequence_lens` in the `prepare_targets` step.

### LINKED ISSUE #253: Improve debuggability of "Inconsistent sequence lengths" error message.
I'm running the XarrayZarrRecipe on a large Grib dataset (Era 5). My FilePattern is structured as follows: 

```
    # ...
    def make_path(time: datetime.datetime, chunk: str) -> str:
        """Make path to Era5 data from timestamp and variable."""

        return (
            f"gs://some/gcs/bucket/"
            f"{time.year:04d}/{time.year:04d}{time.month:02d}_hres_{chunk}.grb2"
        )

    date_range = [
        ts.to_pydatetime()
        for ts in pd.date_range(start=parsed_args.start,
                                end=parsed_args.end,
                                freq="MS").to_list()
    ]
    date_dim = ConcatDim("time", date_range)
    chunks_dim = MergeDim("chunk", parsed_args.chunks)
    pattern = FilePattern(make_path, date_dim, chunks_dim)

    # ...
```

To summarize how the data is structured: Each file is contains a "chunk" of several related variables for a single month.

As I execute the Recipe, I eventually hit this error message in the `prepare_target` stage:

```
ValueError: Inconsistent sequence lengths found: f[[744 744 744 ... 744 744 744]
 [672 672 672 ... 672 672 672]
 [744 744 744 ... 744 744 744]
 ...
 [744 744 744 ... 744 744 744]
 [720 720 720 ... 720 720 720]
 [744 744 744 ... 744 744 744]]
```

After talking with @cisaacstern, the suspected root cause of the error is that some value is inconsistent from the rest of the row that it belongs. In other words, one of the files (or, associated metadata cache) has the wrong number of timesteps compared to all the other merge dimension chunks on that year x month. 

I mention this background context to motivate the need for a better error message here: To fix the Zarr conversion of my pipeline, I need to know: 
- If all the data is really there (for every row, is there one or more value that's different from all the other rows?)
- If it's not there, which item in the concat dimension is missing (what rows have problems?)
- What particular files from the file-pattern path are the culprit?
- Is this really a data error? Or, was the metadata cache produced erroneously? 

Ideally, the error message could be improved to help me answer some or all of these questions.

alxmrs:
I've re-implemented some of the logic of `calculate_sequence_lengths` to be more readable and to provide the user with a better error message. My intention is to provide enough information to the user that they can identify data errors during the `prepare_targets` step of the Xarray --> Zarr conversion.

The new error message now presents the user the problematic indices along the concat dimension, the erroneous values in question and their location. 

Fixes #253.

### DIFF
--- a/pangeo_forge_recipes/recipes/xarray_zarr.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr.py
@@ -393,7 +393,9 @@ def calculate_sequence_lens(
     if len(all_lens.shape) == 1:
         all_lens = np.expand_dims(all_lens, 1)
 
+    # check that all lens are the same along the concat dim
     unique_vec, unique_idx = np.unique(all_lens, axis=1, return_index=True)
+
     if len(unique_idx) > 1:
         correct_col = all_lens[:, unique_idx[0]]
         err_idx = np.where(np.expand_dims(correct_col, 1) != all_lens)
--- a/pangeo_forge_recipes/recipes/xarray_zarr.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr.py
@@ -369,6 +369,15 @@ def get_input_meta(metadata_cache: Optional[MetadataTarget], file_pattern: FileP
     return metadata_cache.getitems([_input_metadata_fname(k) for k in file_pattern])
 
 
+def _get_fname_from_error_pos(pos: Tuple[int, int], file_pattern: FilePattern) -> str:
+    kwargs = {}
+    for idx, p in enumerate(pos):
+        dim = file_pattern.combine_dims[idx]
+        kwargs.update({dim.name: dim.keys[p]})
+
+    return file_pattern.format_function(**kwargs)
+
+
 def calculate_sequence_lens(
     nitems_per_input: Optional[int],
     file_pattern: FilePattern,
@@ -398,10 +407,14 @@ def calculate_sequence_lens(
     if len(unique_idx) > 1:
         correct_col = all_lens[:, unique_idx[np.argmax(unique_counts)]]
         err_idx = np.where(np.expand_dims(correct_col, 1) != all_lens)
-        err_pos = list(zip(*err_idx[::-1]))
+        err_pos = list(zip(*err_idx))
+        # present a list of problematic files to the user, given the file pattern.
+        file_list = '\n'.join([f"- {_get_fname_from_error_pos(epos, file_pattern)!r}" for epos in err_pos])
         raise ValueError(
             f"Inconsistent sequence lengths between indices {unique_idx} of the concat dim."
             f"\nValue(s) {all_lens[err_idx]} at position(s) {err_pos} are different from the rest."
+            f"\nPlease check the following file(s) for data errors:"
+            f"\n{file_list}"
             f"\n{all_lens}"
         )
     return np.atleast_1d(unique_vec.squeeze()).tolist()
--- a/pangeo_forge_recipes/recipes/xarray_zarr.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr.py
@@ -16,6 +16,8 @@ from typing import Callable, Dict, Iterator, List, Optional, Sequence, Set, Tupl
 import dask
 import fsspec
 import numpy as np
+import scipy
+import scipy.stats
 import xarray as xr
 import zarr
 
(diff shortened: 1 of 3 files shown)


rabernat:
Thanks for starting this Alex! Sounds like a useful improvement.

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -16,6 +16,8 @@
 import dask
 import fsspec
 import numpy as np
+import scipy
+import scipy.stats
```
rabernat, line 20: Scipy is a pretty heavy dependency. I'm not sure it's worth bringing on board just to generate a log message.

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -390,14 +392,19 @@ def calculate_sequence_lens(
...
-    sequence_lens = all_lens[tuple(selector)]
-    if not (all_lens == sequence_lens).all():
-        raise ValueError(f"Inconsistent sequence lengths found: f{all_lens}")
-    return np.atleast_1d(sequence_lens.squeeze()).tolist()
+    if len(all_lens.shape) == 1:
+        all_lens = all_lens.reshape((all_lens.shape[0], 1))
+
+    unique_vec, unique_idx = np.unique(all_lens, axis=1, return_index=True)
+    if len(unique_idx) > 1:
+        err_idx = np.where(scipy.stats.mode(all_lens, axis=1)[0] != all_lens)
```
rabernat, line 400: Can we think of a clever way to do this without resorting to scipy.stats.mode? What does this line actually do?

alxmrs: I am trying to calculate the most frequent value in each row -- I am pretty sure I an re-create this with pure numpy.

alxmrs: I want to find the "typical" values in order to find the places where there are discrepancies.

martindurant: `np.bincount` (i.e., histogram with width 1)

alxmrs: Just realized: `unique` returns sorted unique vectors -- so, I probably don't need any other functions in the first place. Patching now...

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -390,14 +390,20 @@ def calculate_sequence_lens(
...
+    if len(all_lens.shape) == 1:
+        all_lens = np.expand_dims(all_lens, 1)
+
+    unique_vec, unique_idx = np.unique(all_lens, axis=1, return_index=True)
+    if len(unique_idx) > 1:
+        correct_col = all_lens[:, unique_idx[0]]
+        err_idx = np.where(np.expand_dims(correct_col, 1) != all_lens)
+        err_pos = list(zip(*err_idx[::-1]))
+        raise ValueError(
+            f"Inconsistent sequence lengths between indicies {unique_idx} of the concat dim."
```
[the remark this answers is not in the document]
alxmrs: Thanks!

alxmrs:
Question for reviewers: Right now, I output strings that look like 

```
Value(s) [ 7 10] at position(s) [(0, 2), (2, 3)] are different from the rest.
```
Instead of "different from the rest", should I also say what the most typical values are?

alxmrs:
I just tested this change end-to-end on a big dataset. As is, here is what the error message would look like: 
```
ValueError: Inconsistent sequence lengths between indices [3 0 1] of the concat dim.
Value(s) [744 744 744 744 744 744 744 744 744 744 744 744 744 744 744 744 720 720
 720 720 720 720 720 720 720 720 720 720 720 720 720 720 697 675] at position(s) [(0, 16), (1, 16), (2, 16), (4, 16), (5, 16), (6, 16), (7, 16), (8, 16), (9, 16), (10, 16), (11, 16), (12, 16), (13, 16), (14, 16), (15, 16), (16, 16), (0, 56), (1, 56), (2, 56), (4, 56), (5, 56), (6, 56), (7, 56), (8, 56), (9, 56), (10, 56), (11, 56), (12, 56), (13, 56), (14, 56), (15, 56), (16, 56), (0, 58), (0, 66)] are different from the rest.
[[744 744 744 ... 744 744 744]
 [672 672 672 ... 672 672 672]
 [744 744 744 ... 744 744 744]
 ...
 [744 744 744 ... 744 744 744]
 [720 720 720 ... 720 720 720]
 [744 744 744 ... 744 744 744]] 
```

In short, this patch needs more work.

For reference, this is for a conversion over many decades (split by month) where there are 17 variables.  It's apparent to me that on concat dim # 16 (which, I assume is the 16th month in the range?), value 3 is different from the rest. Ideally, only that value would be present. A similar pattern happens for the second set of 16 values (value 3 is different from the rest of the concat dim 56).  So, some corrections in my logic (and tests) are needed. 

Ideally, I would get a read out of file paths where there are errors, as this would help me take action to fix the source data errors.

alxmrs:
Ok, I found a fix to the problem stated above. My logic was mostly correct; I was hard-coding an index value. Now the error message reads: 
```
ValueError: Inconsistent sequence lengths between indices [3 0 1] of the concat dim.
Value(s) [673 703 697 675] at position(s) [(3, 16), (3, 56), (0, 58), (0, 66)] are different from the rest.
[[744 744 744 ... 744 744 744]
 [672 672 672 ... 672 672 672]
 [744 744 744 ... 744 744 744]
 ...
 [744 744 744 ... 744 744 744]
 [720 720 720 ... 720 720 720]
 [744 744 744 ... 744 744 744]]
```

Which is much better. 

Next, I've pushed a few changes to display the paths to the problematic files. I'll test this on the real dataset now to see how it looks.

rabernat:
@alxmrs could you merge / rebase this on master? I think that should fix the test failures.

rabernat (approved):
Sorry for taking so long to review Alex!

Improved error messages and better testing...what's not to like!?!  👏

Let's just try to get the tests suite all green and then I'll merge.

pangeo_forge_recipes/recipes/xarray_zarr.py
```diff
@@ -384,20 +393,35 @@ def calculate_sequence_lens(
...
+        file_list = "\n".join(
+            [f"- {_get_fname_from_error_pos(epos, file_pattern)!r}" for epos in err_pos]
+        )
+        raise ValueError(
+            f"Inconsistent sequence lengths between indices {unique_idx} of the concat dim."
+            f"\nValue(s) {all_lens[err_idx]} at position(s) {err_pos} are different from the rest."
+            f"\nPlease check the following file(s) for data errors:"
+            f"\n{file_list}"
+            f"\n{all_lens}"
+        )
```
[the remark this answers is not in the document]
alxmrs: Thanks! :)

alxmrs:
Is this ready to land?

Issue #256, opened 2021-12-18
Should we just adopt xarray-beam as our internal data model?


rabernat:
TL;DR: what the title says

### background

This team should feel very proud of the work we've done on this package over the past few months. We have shipped [dozens of real recipes](https://github.com/pangeo-forge/staged-recipes/pulls) (albeit without any permanent infrastructure to deploy them on) and learned lots of important lessons in the process. 

At the end of a long haul of work on this package, I'd like to zoom out and offer some reflections on our direction in terms of software architecture. I feel like we (specifically, _I_) have flailed around a bit in terms of the core internals of the Recipe model. Examples of big refactorings:
- #101 
- #167 
- #219
- #245 

In the meantime, we have worked on many orthogonal but also important aspects, mostly around I/O. We have also waited **months** 😡 for Columbia to release our free credits from Prefect Cloud to actually deploy our primary bakery, which has given us plenty of time to contemplate our Prefect dependency.

In https://github.com/pangeo-forge/pangeo-forge-recipes/pull/245#issuecomment-997070261, based on my experience trying to generalize the input layer of pangeo-forge-recipes, I am basically suggesting we move to a much more general / flexible model for recipes, where we simply chain stages together. Something like

```python
recipe = Recipe(source, transformation, destination)
```

or in beam-style:
```python
recipe = source | transformation | destination
``` 

This, coupled with a one-way flow of data between these stages, would basically inevitably look a lot like xarray-beam.

We have already danced around various integrations with beam, for example:
- https://github.com/google/xarray-beam/issues/32 - discussing the similarity between the xarray-beam's keys and our indexes
- https://github.com/google/xarray-beam/pull/31
- #169 
- #225 
- #240

I'd like to discuss the pros and cons of incorporating xarray-beam more tightly into pangeo-forge-recipes.

## What is xarray-beam?

Xarray-beam's [data model](https://xarray-beam.readthedocs.io/en/latest/data-model.html) defines a convention for "keys", which are more-or-less equivalent to our Indexes. These keys can be used in the context of general [Beam Pipelines](https://beam.apache.org/documentation/programming-guide/#creating-a-pipeline) which pass around xarray datasets. In contrast to pangeo-forge-recipes, there are not a lot of other custom objects. The [API](https://xarray-beam.readthedocs.io/en/latest/api.html) has just a few functions for reading, writing, rechunking, aggregating, etc.

[...]

PR #257, opened 2022-01-12
Switch Index to be a frozenset

rabernat:
Simplification suggested by Stephan in https://github.com/google/xarray-beam/issues/32#issuecomment-899900847.

Update:

I was amazed how simple it was to make this change. I just swapped our our `Index` class with
```python
Index = frozenset[DimIndex]
```
and mostly everything just worked, requiring no change to how indexes are created (e.g. `idx = Index((some, stuff))`). That was a nice surprise.

However, this PR has turned up some nasty mypy-related stuff, and also showed that that simple approach doesn't work in 3.8. In 3.8, we would need to do something like
```python
Index = FrozenSet[DimIndex]
```
However, if we do that, we get the error `TypeError: Type FrozenSet cannot be instantiated; use frozenset() instead` when we try to instantiate the Index.

I would like to say we will just drop python 3.8. However, afaik, beam does not yet work with 3.9. So it looks like we will need some workarounds.

Does anyone know the best way to solve this?

### DIFF
--- a/pangeo_forge_recipes/patterns.py
+++ b/pangeo_forge_recipes/patterns.py
@@ -1,14 +1,11 @@
 """
 Filename / URL patterns.
 """
-from __future__ import annotations
-
 import inspect
 from dataclasses import dataclass, field, replace
 from enum import Enum
 from itertools import product
 from typing import (
-    TYPE_CHECKING,
     Any,
     Callable,
     ClassVar,
@@ -94,13 +91,8 @@ class DimIndex:
         assert self.index < self.sequence_len
 
 
-if TYPE_CHECKING:
-    # If we just do this, we can't initialize an Index by just writing
-    # Index(list_of_dims)
-    Index = FrozenSet[DimIndex]
-else:
-    # But if we just do this, it won't pass mypy 😖
-    Index = frozenset[DimIndex]
+class Index(FrozenSet[DimIndex]):
+    pass
 
 
 CombineDim = Union[MergeDim, ConcatDim]
--- a/pangeo_forge_recipes/patterns.py
+++ b/pangeo_forge_recipes/patterns.py
@@ -1,17 +1,19 @@
 """
 Filename / URL patterns.
 """
+from __future__ import annotations
 
 import inspect
 from dataclasses import dataclass, field, replace
 from enum import Enum
 from itertools import product
 from typing import (
+    TYPE_CHECKING,
     Any,
     Callable,
     ClassVar,
     Dict,
-    Iterable,
+    FrozenSet,
     Iterator,
     List,
     Optional,
@@ -68,7 +70,7 @@ class MergeDim:
     operation: ClassVar[CombineOp] = CombineOp.MERGE
 
 
-@dataclass(frozen=True)
+@dataclass(frozen=True, order=True)
 class DimIndex:
     """Object used to index a single dimension of a FilePattern or Recipe Chunks.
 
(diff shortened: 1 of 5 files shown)


pangeo_forge_recipes/patterns.py
```diff
@@ -92,27 +94,13 @@ def __post_init__(self):
...
-
-    def __hash__(self):
-        return hash(frozenset(self))
+if TYPE_CHECKING:
+    # If we just do this, we can't initialize an Index by just writing
+    # Index(list_of_dims)
+    Index = FrozenSet[DimIndex]
+else:
+    # But if we just do this, it won't pass mypy 😖
+    Index = frozenset[DimIndex]
```
rabernat, line 103: Tests have shown that this does not work on python 3.8, even with `from __future__ import annotations`

rabernat:
Ok I think I found the solution:

```python
class Index(FrozenSet[DimIndex]):
    pass
```

rabernat:
The upstream-dev tests have surfaced a new issue

```
        # https://github.com/pangeo-forge/pangeo-forge-recipes/issues/151
        # Requires >4 GiB of memory to run.
        pd = pytest.importorskip("pandas")
        distributed = pytest.importorskip("distributed")
    
        dates = pd.date_range("2020-05-31T00:00:00", "2021-05-31T23:59:59", freq="30min")
        time_concat_dim = ConcatDim("time", dates, nitems_per_file=1)
        pattern = FilePattern(_make_filename_for_memory_usage_test, time_concat_dim,)
    
        recipe = XarrayZarrRecipe(
            pattern, xarray_open_kwargs={"group": "Grid", "decode_coords": "all"}, inputs_per_chunk=1,
        )
    
>       delayed = recipe.to_dask()

tests/recipe_tests/test_graph_memory.py:49: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pangeo_forge_recipes/recipes/base.py:28: in to_dask
    return DaskPipelineExecutor.compile(self._compiler())
pangeo_forge_recipes/executors/dask.py:75: in compile
    delayed = Delayed(prev_token[0], hlg)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = Delayed('finalize_target-2f31500925ec7e187d227c54ab925c75')
key = 'finalize_target-2f31500925ec7e187d227c54ab925c75'
dsk = HighLevelGraph with 5 layers.
<dask.highlevelgraph.HighLevelGraph object at 0x7f63f0f6e2b0>
 0. config
 1. cache_input
 2. prepare_target
 3. store_chunk
 4. finalize_target

length = None, layer = None

    def __init__(self, key, dsk, length=None, layer=None):
        self._key = key
        self._dask = dsk
        self._length = length
    
        # NOTE: Layer is used by `to_delayed` in other collections, but not in normal Delayed use
        self._layer = layer or key
        if isinstance(dsk, HighLevelGraph) and self._layer not in dsk.layers:
>           raise ValueError(
                f"Layer {self._layer} not in the HighLevelGraph's layers: {list(dsk.layers)}"
            )
E           ValueError: Layer finalize_target-2f31500925ec7e187d227c54ab925c75 not in the HighLevelGraph's layers: ['config', 'cache_input', 'prepare_target', 'store_chunk', 'finalize_target']
```

Apparently the way we are constructing high-level graphs is not compatible with the latest dask release.

I'm inclined to merge this, let the master tests fail, and open a new issue for it.

cisaacstern:
> I'm inclined to merge this, let the master tests fail, and open a new issue for it.

👍 upstream-dev tests ftw

cisaacstern:
Out of curiosity, is there a one-ish sentence way to encapsulate the advantage of a frozenset vs. a tuple here? I wasn't able to immediately grasp the motivation from the linked xarray-beam thread.

pangeo_forge_recipes/patterns.py
```diff
@@ -92,27 +91,8 @@ def __post_init__(self):
...
-        return tuple.__new__(Index, args)
-
-    def __str__(self):
-        return ",".join(str(dim) for dim in self)
-
-    def __eq__(self, other):
-        return (set(self) == set(other)) and (len(self) == len(other))
-
-    def __hash__(self):
-        return hash(frozenset(self))
```
rabernat, line 115: This class was already basically doing the same thing as a frozenset! So this change simplifies our Index type--a key data structure--to use a built-in type. Simpler and less code to maintain.

For context, this is a tiny part of the changes that will move us in the direction of #256.

PR #260, opened 2022-01-12
Fix dask hlg generation

### LINKED ISSUE #259: Dask executor is broken with dask upstream
In our Dask executor, we create a high-level graph as follows:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/6194c72343d82e39b09edcafbe7085b1d9464787/pangeo_forge_recipes/executors/dask.py#L74-L76

Our upstream test (dask commit `725110f9367931291a3e68c9d582544cdb032f77`) has revealed the following [error](https://github.com/pangeo-forge/pangeo-forge-recipes/runs/4793899318?check_suite_focus=true), triggered by the line

```python
delayed = Delayed(prev_token[0], hlg)
```

```
self = Delayed('finalize_target-8dfa49acbe8cb65c6a2f82ebca37350d')
key = 'finalize_target-8dfa49acbe8cb65c6a2f82ebca37350d'
dsk = HighLevelGraph with 5 layers.
<dask.highlevelgraph.HighLevelGraph object at 0x7f338c0b3280>
 0. config
 1. cache_input
 2. prepare_target
 3. store_chunk
 4. finalize_target

length = None, layer = None

    def __init__(self, key, dsk, length=None, layer=None):
        self._key = key
        self._dask = dsk
        self._length = length
    
        # NOTE: Layer is used by `to_delayed` in other collections, but not in normal Delayed use
        self._layer = layer or key
        if isinstance(dsk, HighLevelGraph) and self._layer not in dsk.layers:
>           raise ValueError(
                f"Layer {self._layer} not in the HighLevelGraph's layers: {list(dsk.layers)}"
            )
E           ValueError: Layer finalize_target-8dfa49acbe8cb65c6a2f82ebca37350d not in the HighLevelGraph's layers: ['config', 'cache_input', 'prepare_target', 'store_chunk', 'finalize_target']
```

It looks like Dask does not like how we are creating the Delayed object. Is this a dask regression? Or do we need to change our code.

@TomAugspurger - any insights here?

rabernat:
Fixes #259.

### DIFF
--- a/pangeo_forge_recipes/executors/dask.py
+++ b/pangeo_forge_recipes/executors/dask.py
@@ -51,7 +51,6 @@ class DaskPipelineExecutor(PipelineExecutor[Delayed]):
         prev_dependency = ()  # type: Union[Tuple[()], Tuple[str]]
         for stage in pipeline.stages:
             stage_graph = {}  # type: Dict[Union[str, Tuple[str, int]], Any]
-            # each block should
             if stage.mappable is None:
                 stage_key = append_token(stage.name, token)
                 func = wrap_standalone_task(stage.function)
--- a/pangeo_forge_recipes/executors/dask.py
+++ b/pangeo_forge_recipes/executors/dask.py
@@ -43,37 +43,35 @@ class DaskPipelineExecutor(PipelineExecutor[Delayed]):
 
         # start with just the config as a standalone layer
         # create a custom delayed object for the config
-        config_task = "config"
-        config_token = append_token(config_task, token)
-        layers[config_task] = {config_token: pipeline.config}
-        dependencies[config_task] = set()
+        config_key = append_token("config", token)
+        layers[config_key] = {config_key: pipeline.config}
+        dependencies[config_key] = set()
 
-        prev_token = ()  # type: Tuple[str, ...]
-        prev_task = config_task
+        prev_key = config_key
+        prev_dependency = ()  # type: Union[Tuple, Tuple[str]]
         for stage in pipeline.stages:
-            stage_token = append_token(stage.name, token)
             stage_graph = {}  # type: Dict[Union[str, Tuple[str, int]], Any]
+            # each block should
             if stage.mappable is None:
+                stage_key = append_token(stage.name, token)
                 func = wrap_standalone_task(stage.function)
-                stage_graph[stage_token] = (func, config_token) + prev_token
-                prev_token = (stage_token,)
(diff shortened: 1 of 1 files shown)


rabernat:
This worked for me locally (except for `test_memory_usage`) but did not work in the CI. 😞

rabernat:
Ok I can now reproduce this locally if I upgrade to dask==2021.12.0.

pangeo_forge_recipes/executors/dask.py
```diff
@@ -43,36 +43,35 @@ def compile(pipeline: Pipeline):
...
+        dependencies[config_key] = set()
 
-        prev_token = ()  # type: Tuple[str, ...]
-        prev_task = config_task
+        prev_key = config_key
+        prev_dependency = ()  # type: Union[Tuple[()], Tuple[str]]
         for stage in pipeline.stages:
-            stage_token = append_token(stage.name, token)
             stage_graph = {}  # type: Dict[Union[str, Tuple[str, int]], Any]
+            # each block should
```
TomAugspurger, line 54: Stray comment?

pangeo_forge_recipes/executors/dask.py
```diff
@@ -43,36 +43,35 @@ def compile(pipeline: Pipeline):
...
-        layers[config_task] = {config_token: pipeline.config}
-        dependencies[config_task] = set()
+        config_key = append_token("config", token)
+        layers[config_key] = {config_key: pipeline.config}
+        dependencies[config_key] = set()
 
-        prev_token = ()  # type: Tuple[str, ...]
-        prev_task = config_task
+        prev_key = config_key
+        prev_dependency = ()  # type: Union[Tuple[()], Tuple[str]]
```
TomAugspurger, line 51: Just for my own curiosity, just having `Tuple[str, ...]` didn't work? I would have thought that a length-zero tuple would be allowed.

TomAugspurger: Although, I guess this is slightly better, since it's explicitly either a size-0 empty tuple or a size-1 tuple (with a str).

TomAugspurger (approved):

pangeo_forge_recipes/executors/dask.py
rabernat: > Just for my own curiosity, just having `Tuple[str, ...]` didn't work?

Never tried it. For me mypy compliance is usually a matter of trial and error. 😬

PR #261, opened 2022-01-13
Use blockwise for Dask executor

gjoseph92:
See discussion around https://github.com/pangeo-forge/pangeo-forge-recipes/issues/259#issuecomment-1012518978. I think this works? Note that HLGs are changing a lot, so this will probably be broken even more frequently by upstream changes. And is even harder for other contributors to understand than it was before.

### DIFF
--- a/pangeo_forge_recipes/executors/dask.py
+++ b/pangeo_forge_recipes/executors/dask.py
@@ -1,8 +1,11 @@
+from __future__ import annotations
+
 from typing import Any, Dict, Set, Tuple, Union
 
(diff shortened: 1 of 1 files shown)


rabernat:
This is really great Gabe. All our tests pass and I'm inclined to merge as is. :pray:

My one worry is maintenance (I'm not sure I 100% understand what is happening), but your comments help mitigate that. I would definitely want @TomAugspurger's :+1: for this.

TomAugspurger (approved):

Issue #254, opened 2021-12-08
`finalize_target` is slow on large recipes with remote filesystem #987


TomAugspurger:
`finalize_target`, specifically https://github.com/pangeo-forge/pangeo-forge-recipes/blob/4f8c0522c1395aaf15a820cfd637a1f1e76729b2/pangeo_forge_recipes/recipes/xarray_zarr.py#L558, can be slow.

We call `set(group)` to figure out what which variables are available. I didn't realize it, but apparently that can trigger a full listing of the bucket / storage container, which is slow with many objects. (Zarr has a fastpath if the object implements `listdirs` https://github.com/zarr-developers/zarr-python/blob/f0677c2e051cad672e781c6bf63b5edfe57aaca4/zarr/storage.py#L163-L165), but fsspec mappers don't; The Azure portal does it somehow, so it might be feasible)

I wonder if we can determine the set of variables based on some other mechanism. For now, I'm going to manually run `finalize_target` with hard-coded variables for the dataset I'm working on (gpm-imerg).

cc @sharkinsspatial, I think that this is one of the reasons `finalize_target` was taking so long for you.

edit: Oh I just discoved the `walk_blobs` method in azure.storage.blob. Perahps fsspec could use `walk` to implement `listdirs`. I'll investigate and open an fsspec issue if it seems promising.

TomAugspurger:
From Ryan in https://github.com/pangeo-data/pangeo-cloud-federation/issues/987#issuecomment-989181682

> Thanks for reporting Tom.

> This seems like something that we should try to fix upstream, in fsspec and / or the fsspec + Zarr interface.

> In the meantime, if listing the group is really slow, we could at least only do it once. I think the current code iterates through the group keys twice.

Agreed on both counts.

martindurant:
> Zarr has a fastpath if the object implements listdirs https://github.com/zarr-developers/zarr-python/blob/f0677c2e051cad672e781c6bf63b5edfe57aaca4/zarr/storage.py#L163-L165), but fsspec mappers don't

Pretty sure the FSSpecStore does! I'll check...

martindurant:
https://github.com/zarr-developers/zarr-python/blob/master/zarr/storage.py#L1236

TomAugspurger:
[...]

PR #265, opened 2022-01-25
Avoid costly `set(group)` operation in finalize_target.

alxmrs:
As pointed out in #254, taking a set difference happens to be an expensive operation for large datasets. Specifically, taking `set(group)` lists all files in the bucket, which is currently implemented inefficiently.

In this patch, we find the set intersection with a filter operation. This should be faster since it only checks if a smaller subset of files exists.

### DIFF
--- a/pangeo_forge_recipes/recipes/xarray_zarr.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr.py
@@ -555,7 +555,7 @@ def finalize_target(*, config: XarrayZarrRecipe) -> None:
         # https://github.com/pangeo-forge/pangeo-forge-recipes/issues/214
         # intersect the dims from the array metadata with the Zarr group
         # to handle coordinateless dimensions.
-        dims = set(_gather_coordinate_dimensions(group)) & set(group)
+        dims = {dim for dim in _gather_coordinate_dimensions(group) if dim in group}
         for dim in dims:
             arr = group[dim]
             attrs = dict(arr.attrs)
--- a/pangeo_forge_recipes/recipes/xarray_zarr.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr.py
@@ -553,7 +553,7 @@ def finalize_target(*, config: XarrayZarrRecipe) -> None:
         target_mapper = config.target.get_mapper()
         group = zarr.open(target_mapper, mode="a")
         # https://github.com/pangeo-forge/pangeo-forge-recipes/issues/214
-        # intersect the dims from the array metadata with the Zarr group
+        # filter out the dims from the array metadata not in the Zarr group
         # to handle coordinateless dimensions.
         dims = (dim for dim in _gather_coordinate_dimensions(group) if dim in group)
         for dim in dims:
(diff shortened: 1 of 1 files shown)


rabernat:
Fantastic Alex! Will have a close look at this within a day or two.

TomAugspurger:
Trying this out now. I think it will help, but unfortunately the implementation of `_gather_dimension_coordinates` at https://github.com/pangeo-forge/pangeo-forge-recipes/pull/265/files#diff-e12c886cc124886c5cfa5313d760a36c39649af9da845077c663e6feab8487b5L541-L544 triggers the expensive `listdir` too :/ Here's the stacktrace when I interrupt it:

```pytb
KeyboardInterrupt                         Traceback (most recent call last)
Input In [24], in <module>
----> 1 dims = pangeo_forge_recipes.recipes.xarray_zarr._gather_coordinate_dimensions(group)

File /srv/conda/envs/notebook/lib/python3.8/site-packages/pangeo_forge_recipes/recipes/xarray_zarr.py:543, in _gather_coordinate_dimensions(group)
    541 def _gather_coordinate_dimensions(group: zarr.Group) -> List[str]:
    542     return list(
--> 543         set(itertools.chain(*(group[var].attrs.get(_ARRAY_DIMENSIONS, []) for var in group)))
    544     )

File /srv/conda/envs/notebook/lib/python3.8/site-packages/pangeo_forge_recipes/recipes/xarray_zarr.py:543, in <genexpr>(.0)
    541 def _gather_coordinate_dimensions(group: zarr.Group) -> List[str]:
    542     return list(
--> 543         set(itertools.chain(*(group[var].attrs.get(_ARRAY_DIMENSIONS, []) for var in group)))
    544     )

File /srv/conda/envs/notebook/lib/python3.8/site-packages/zarr/hierarchy.py:215, in Group.__iter__(self)
    196 def __iter__(self):
    197     """Return an iterator over group member names.
    198 
    199     Examples
   (...)
    213 
    214     """
--> 215     for key in sorted(listdir(self._store, self._path)):
    216         path = self._key_prefix + key
    217         if (contains_array(self._store, path) or
    218                 contains_group(self._store, path)):

File /srv/conda/envs/notebook/lib/python3.8/site-packages/zarr/storage.py:175, in listdir(store, path)
    172     return store.listdir(path)
    173 else:
    174     # slow version, iterate through all keys
--> 175     return _listdir_from_keys(store, path)

File /srv/conda/envs/notebook/lib/python3.8/site-packages/zarr/storage.py:157, in _listdir_from_keys(store, path)
    155 prefix = _path_to_prefix(path)
    156 children = set()
--> 157 for key in list(store.keys()):
    158     if key.startswith(prefix) and len(key) > len(prefix):
    159         suffix = key[len(prefix):]

File /srv/conda/envs/notebook/lib/python3.8/_collections_abc.py:702, in MappingView.__len__(self)
    701 def __len__(self):
--> 702     return len(self._mapping)

File /srv/conda/envs/notebook/lib/python3.8/site-packages/fsspec/mapping.py:160, in FSMap.__len__(self)
    159 def __len__(self):
--> 160     return len(self.fs.find(self.root))

File /srv/conda/envs/notebook/lib/python3.8/site-packages/adlfs/spec.py:965, in AzureBlobFileSystem.find(self, path, withdirs, prefix, **kwargs)
    964 def find(self, path, withdirs=False, prefix="", **kwargs):
--> 965     return sync(
    966         self.loop, self._find, path=path, withdirs=withdirs, prefix=prefix, **kwargs
    967     )
```

So the `for var in group` needs to be avoided, at least for mapping that don't implement `listdirs`.

alxmrs:
I was slow to reply but: Thanks for catching that, Tom. That explains my experience running this change. I was looking a bit into ways to avoid the `listdir` op you pointed out, however, such an approach would likely introduce unnecessary complexity vs the upstream fix.

Thanks, Ryan, for merging the change.

PR #284, opened 2022-02-14
add mypy_extensions to dependencies

### LINKED ISSUE #282: Installing `pangeo-forge-recipes`
## Issue
Answer the question: How should users set up an environment to use Pangeo Forge?

(It's very possible that this whole issue is not necessary and I am just confused, but I thought I would post it to attempt to clarify the current state)

### Long term plan: `conda`
In the long term it sounds like the preferred method is that users will be able to `conda install pangeo-forge-recipes`. At the current moment this is not yet implemented.

### Pip
There is a version of `pangeo-forge-recipes` that can be pip installed, but for me that verison gives an error when I try to use `XarrrayZarrRecipe`. If the current pip install is broken and we don't plan to support pip long term I wonder if we should remove `pangeo-forge-recipes` from pypi.

#### Error
`from pangeo_forge_recipes.recipes import XarrayZarrRecipe` yields:
```
[ImportError: cannot import name 'BlockSizeError' from 'fsspec.implementations.http' 
(/Users/rwegener/.pyenv/versions/miniconda3-4.7.12/envs/pf_env/lib/python3.9/site-
packages/fsspec/implementations/http.py)]()
```

### Local Development Install
As I understand it the only way to install `pangeo-forge-recipes` right now is to clone the repo and then do a `pip install -e`. 

## Action
For Ocean Sciences we will be using binder, so the installation question is not directly relevant. Still, I think it would be nice to have a clear recommended method for installation and for that to be directly accessible in the documentation.

rabernat:
Closes #282

I hope

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,6 +39,7 @@ install_requires =
     numcodecs >= 0.9.0
     fsspec[http] >= 2021.6.0
     fsspec-reference-maker >= 0.0.4
+    mypy_extensions
 
 [options.extras_require]
 dev =


cisaacstern (commented):
Looks good! But I'm not sure that it closes the linked issue about installation, given that @rwegener2 identifies the need to `brew install` system packages which pip cannot install. It would seem completing our open Conda Forge PR is the only (?) path forward to make the installation as low-friction as possible?

rabernat:
I don't see why. It's perfectly possible to pip install netcdf and any other dependencies.

cisaacstern:
> I don't see why. It's perfectly possible to pip install netcdf and any other dependencies.

Yes, this was my mistake. I interpreted https://github.com/pangeo-forge/pangeo-forge-recipes/issues/282#issuecomment-1039602593 to mean that somehow the install was broken without going outside pip for other dependencies. But I just:

- made a new Conda environment
- pip-installed `mypy_extensions` in it (which merging this PR render unnecessary)
- then pip-installed `"pangeo-forge-recipes==0.7.0"`

And I appear to end up with a working `pangeo-forge-recipes` installation. So IMO this is ready to merge.

Issue #276, opened 2022-02-04
Simplify storage configuration


rabernat:
Setting up storage on a recipe is currently pretty awkward and verbose, e.g. from [a tutorial](https://pangeo-forge.readthedocs.io/en/latest/tutorials/xarray_zarr/terraclimate.html)

```python
import tempfile
from fsspec.implementations.local import LocalFileSystem
from pangeo_forge_recipes.storage import FSSpecTarget, CacheFSSpecTarget, MetadataTarget

fs_local = LocalFileSystem()

target_dir = tempfile.TemporaryDirectory()
target = FSSpecTarget(fs_local, target_dir.name)

cache_dir = tempfile.TemporaryDirectory()
cache_target = CacheFSSpecTarget(fs_local, cache_dir.name)

meta_dir = tempfile.TemporaryDirectory()
meta_store = MetadataTarget(fs_local, meta_dir.name)

recipe.target = target
recipe.input_cache = cache_target
recipe.metadata_cache = meta_store
```

We should make this more compact and / or provide some reasonable defaults. We could even use a config file or environment variable to determine some parts automatically.

Issue #279, opened 2022-02-08
`copy_pruned` fails when `nkeep` < `inputs_per_chunk`


cisaacstern:
I needed to push https://github.com/pangeo-forge/registrar/pull/1/commits/fa6bc0d8960f4917d3b08501d3a22bfa4e072023 to get `copy_pruned` to work for the OISST recipe without raising

```python
---------------------------------------------------------------------------
Task 'store_chunk[0]': Exception encountered during task execution!
Traceback (most recent call last):
  File "/srv/conda/envs/notebook/lib/python3.8/site-packages/prefect/engine/task_runner.py", line 861, in get_task_run_state
    value = prefect.utilities.executors.run_task_with_timeout(
  File "/srv/conda/envs/notebook/lib/python3.8/site-packages/prefect/utilities/executors.py", line 323, in run_task_with_timeout
    return task.run(*args, **kwargs)  # type: ignore
  File "/Users/charlesstern/Dropbox/pangeo/registrar/registrar/flow.py", line 38, in wrapper
  File "/srv/conda/envs/notebook/lib/python3.8/site-packages/pangeo_forge_recipes/recipes/xarray_zarr.py", line 499, in store_chunk
    write_region, conflicts = region_and_conflicts_for_chunk(config, chunk_key)
  File "/srv/conda/envs/notebook/lib/python3.8/site-packages/pangeo_forge_recipes/recipes/xarray_zarr.py", line 232, in region_and_conflicts_for_chunk
    target_grid = ChunkGrid.from_uniform_grid(
  File "/srv/conda/envs/notebook/lib/python3.8/site-packages/pangeo_forge_recipes/chunk_grid.py", line 45, in from_uniform_grid
    assert (
AssertionError: invalid chunksize 20 and dimsize 2
```
I'm surprised that we didn't hit this earlier, but it seems like an actual bug (rather than user error). Basically, the un-pruned recipe defines `inputs_per_chunk=20` but the pruned recipe only has a total of 2 input files, and so fails when `region_and_conflicts_per_chunk` is called within `store_chunk`.

_Originally posted by @cisaacstern in https://github.com/pangeo-forge/registrar/issues/1#issuecomment-1033107536_

rabernat:
Looks similar to https://github.com/pangeo-forge/pangeo-forge-recipes/issues/200

PR #286, opened 2022-02-15
Add `.assign_temporary_storage` method via `TemporaryStorageMixin`

cisaacstern:
A lightweight (albeit possibly partial) solution for #276 which is hopefully simple enough to merge (and release) before OSM2022, to aid @rwegener2's ongoing work on https://github.com/pangeo-forge/pangeo-forge-recipes/pull/281.

To achieve the same result as represented by the code block in https://github.com/pangeo-forge/pangeo-forge-recipes/issues/276#issue-1124391211, users would now need only call:

```python
recipe.assign_temporary_storage()
```

I considered setting true defaults (as opposed using the method call), but ultimately thought requiring users to make this method call had some instructive value, insofar as it conveys the notion that storage targets for a given recipe are customizable.

@rabernat, by way of background, Rachel and I believe we need _some type_ of simplification for OSM in this area. I am by no means attached to this particular solution, it simply represents my best effort at something simple enough to merge (and then document!) on our short runway. Totally open to any alternative if you can think of a better one.

### DIFF
--- a/pangeo_forge_recipes/recipes/base.py
+++ b/pangeo_forge_recipes/recipes/base.py
@@ -1,11 +1,15 @@
 from __future__ import annotations
 
+import tempfile
 from abc import ABC
 from dataclasses import dataclass, replace
 from typing import Callable, ClassVar
 
+from fsspec.implementations.local import LocalFileSystem
+
 from ..executors.base import Pipeline
 from ..patterns import FilePattern, prune_pattern
+from ..storage import CacheFSSpecTarget, FSSpecTarget, MetadataTarget
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


rabernat:
I agree this is needed. But I have some different ideas about how to implement it.

I would like to first create a `StorageConfig` object that wrap target, cache, and metadata targets. So when we "assign storage" to a recipe, we just have one parameter to set. This will be an API change that will break many existing tutorials and examples.

Then we will make a default factory for a `TempStorageConfig`. This will effectively allow us to just remove the storage configuration from all the examples.

I would like to work on this (along with #279) early next week once I get out from under this week's deadlines. Would that timeline work for you? So for the purposes of planning the tutorial, you can definitely assume that this will be resolved soon.

cisaacstern:
Implementation timeline sounds good. Rachel and l are meeting to continue tutorial prep on Thursday, at which point it would be useful to have some (even placeholder) idea of what your API redesign will look like, so we can drop that into our evolving draft.

Sounds like rather than
```python
recipe.assign_temporary_storage()
```
you're envisioning a one-liner like

```python
recipe.storage_config = TempStorageConfig()
```
?

cisaacstern:
> Then we will make a default factory for a TempStorageConfig. This will effectively allow us to just remove the storage configuration from all the examples.

Oh... IIUC, this then becomes a true default on the recipe classes, insofar as users can just jump right into execution, without even a one-liner. That works! Closing this now.

rabernat:
It would probably be a class like this:

```python
@dataclass
class StorageConfig:
    target: FSSpecTarget
    cache: Optional[FSSpecTarget] = None
    metadata: Optional[MetadataTarget] = None
```

A function like this

```python
def temporary_storage_config():
    # create temporary directories
    return StorageConfig(...)
```

and a mixin like this

```python
@dataclass
class StorageMixin:
    storage_config: StorageConfig = field(default_factory=temporary_storage_config)
 ```

This would be mixed into both recipe classes (as yours is now). The default factory would automatically create a new temporary storage config on every recipe. The main API difference with what you have here would be that temporary storage would be the default, with no explicit `.assign_temporary_storage()` step .

If you would like to implement this yourself, please go right ahead!

Sorry for being so specific here. I have spent quite some time thinking this through and have a strong opinion on how it should be done.

Issue #272, opened 2022-02-04
Docs environment creation


rwegener2:
## The Issue
**Goal:** build development environment for the documentation

**Problem in a nutshell 🥜** - Building the documentation environment with `pip install -r ./docs/requirements.txt` results in an error that seems to be due to a versioning conflict between the dependencies

### The Error
When I try to install the requirements I get the error:
```
  error: subprocess-exited-with-error

  × python setup.py egg_info did not run successfully.
  │ exit code: 1
      Traceback (most recent call last):
        File "<string>", line 2, in <module>
        File "<pip-setuptools-caller>", line 14, in <module>
        File "/Users/rwegener/.pyenv/versions/miniconda3-4.7.12/envs/pf-docs-test/lib/python3.9/site-packages/setuptools/__init__.py", line 16, in <module>
          import setuptools.version
        File "/Users/rwegener/.pyenv/versions/miniconda3-4.7.12/envs/pf-docs-test/lib/python3.9/site-packages/setuptools/version.py", line 1, in <module>
          import pkg_resources
        File "/Users/rwegener/.pyenv/versions/miniconda3-4.7.12/envs/pf-docs-test/lib/python3.9/site-packages/pkg_resources/__init__.py", line 74, in <module>
          from pkg_resources.extern.jaraco.text import (
        File "/Users/rwegener/.pyenv/versions/miniconda3-4.7.12/envs/pf-docs-test/lib/python3.9/site-packages/pkg_resources/_vendor/jaraco/text/__init__.py", line 11, in <module>
          from pkg_resources.extern.jaraco.functools import compose, method_cache
        File "/Users/rwegener/.pyenv/versions/miniconda3-4.7.12/envs/pf-docs-test/lib/python3.9/site-packages/pkg_resources/_vendor/jaraco/functools.py", line 8, in <module>
          import pkg_resources.extern.more_itertools
        File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
        File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
        File "<frozen importlib._bootstrap>", line 666, in _load_unlocked
        File "<frozen importlib._bootstrap>", line 565, in module_from_spec
        File "/Users/rwegener/.pyenv/versions/miniconda3-4.7.12/envs/pf-docs-test/lib/python3.9/site-packages/pkg_resources/extern/__init__.py", line 52, in create_module
          return self.load_module(spec.name)
[...]

PR #287, opened 2022-02-15
Add `setup_logging` convenience function

cisaacstern:
The convenience function added in this PR is a version of the same function we've recommended to users in tutorials, with the addition of the `logger.handlers.clear()` block, which allows the function to be reused during a single interpreter session without duplicating logs.

The addition of this function is specifically motivated to help @rwegener2 with ongoing work on https://github.com/pangeo-forge/pangeo-forge-recipes/pull/281 in advance of OSM2022. Pre-packaging this function hopefully lowers the barrier to entry for new recipe developers by giving them the tools they need to get started with execution debugging out-of-the-box.

I considered various places we might add this function and ultimately settled on `.recipes.__init__` because users will already be importing from that path, so they can now just do:

```python
from pangeo_forge_recipes.recipes import XarrayZarrRecipe, setup_logging
```

Open to alternatives.

### DIFF
--- a/pangeo_forge_recipes/recipes/__init__.py
+++ b/pangeo_forge_recipes/recipes/__init__.py
@@ -2,3 +2,24 @@ from .reference_hdf_zarr import HDFReferenceRecipe
 from .xarray_zarr import XarrayZarrRecipe
 
 __all__ = ["XarrayZarrRecipe", "HDFReferenceRecipe"]
+
+
+def setup_logging(level: str = "INFO"):
+    """A convenience function that sets up logging for developing and debugging recipes in Jupyter,
+    iPython, or another interactive context.
+
+    :param level: One of (in decreasing level of detail) ``"DEBUG"``, ``"INFO"``, or ``"WARNING"``.
+      Defaults to ``"INFO"``.
+    """
+
+    import logging
+    import sys
+
+    formatter = logging.Formatter("%(name)s - %(levelname)s - %(message)s")
+    logger = logging.getLogger("pangeo_forge_recipes")
+    if logger.hasHandlers():
+        logger.handlers.clear()
+    logger.setLevel(getattr(logging, level))
+    sh = logging.StreamHandler(stream=sys.stdout)
+    sh.setFormatter(formatter)
+    logger.addHandler(sh)
--- a/pangeo_forge_recipes/recipes/__init__.py
+++ b/pangeo_forge_recipes/recipes/__init__.py
@@ -11,15 +11,15 @@ def setup_logging(level: str = "INFO"):
     :param level: One of (in decreasing level of detail) ``"DEBUG"``, ``"INFO"``, or ``"WARNING"``.
       Defaults to ``"INFO"``.
     """
-
     import logging
-    import sys
 
-    formatter = logging.Formatter("%(name)s - %(levelname)s - %(message)s")
+    from rich.logging import RichHandler
+
     logger = logging.getLogger("pangeo_forge_recipes")
     if logger.hasHandlers():
         logger.handlers.clear()
     logger.setLevel(getattr(logging, level))
-    sh = logging.StreamHandler(stream=sys.stdout)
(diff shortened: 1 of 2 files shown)


rabernat:
LGTM and is aligned with the advice [here](https://stackoverflow.com/questions/27016870/how-should-logging-be-used-in-a-python-package/27017068).

rabernat (approved):

rabernat:
One thought...we might try to import rich and use the [rich logger handler](https://rich.readthedocs.io/en/stable/logging.html) if possible...just because it's pretty and rich is cool.

cisaacstern:
That `RichHandler` looks great! (First in Terminal, then in JupyterLab; click for full resolution.)

Screen Shot 2022-02-15 at 3 54 11 PM
Screen Shot 2022-02-15 at 3 58 30 PM

pangeo_forge_recipes/recipes/__init__.py
```diff
@@ -2,3 +2,24 @@
...
+      Defaults to ``"INFO"``.
+    """
+    import logging
+
+    from rich.logging import RichHandler
+
+    logger = logging.getLogger("pangeo_forge_recipes")
+    if logger.hasHandlers():
+        logger.handlers.clear()
+    logger.setLevel(getattr(logging, level))
```
alxmrs, line 21: logging level is just an int, if you wanted to make the default level `logging.INFO`.

pangeo_forge_recipes/recipes/__init__.py
```diff
@@ -2,3 +2,24 @@
...
+def setup_logging(level: str = "INFO"):
+    """A convenience function that sets up logging for developing and debugging recipes in Jupyter,
+    iPython, or another interactive context.
+
+    :param level: One of (in decreasing level of detail) ``"DEBUG"``, ``"INFO"``, or ``"WARNING"``.
+      Defaults to ``"INFO"``.
+    """
+    import logging
+
+    from rich.logging import RichHandler
```
alxmrs, line 16: I'm hesitant of introducing rich as a dependency to the project. The console output is nice, but it increases the surface area for potential security exploits (still reeling from the Log4J issues over the holidays).

What do you think, @rabernat?

pangeo_forge_recipes/recipes/__init__.py
cisaacstern: Thanks for jumping in on this review, Alex.

AFAICT, by setting the default level as we currently do, via a default `level="INFO"` in the signature, the function is more self-documenting than if we were to set the default with the `int` equivalent. That is, the default is currently set in a manner that reflects how we'd like users to interact with it.

pangeo_forge_recipes/recipes/__init__.py
rabernat: I agree we should not make rich a hard dependency. Let's try to use rich (`try: import rich`) if it is installed and otherwise fall back on the builtin logger otherwise.

But I don't think that this has any intersection with the Log4J issue. Rich just provides a handler for python's builtin logging framework.

cisaacstern:
🤔 readthedocs build is broken in a way that looks potentially similar to the issues @rwegener2 identified in https://github.com/pangeo-forge/pangeo-forge-recipes/issues/272

pangeo_forge_recipes/recipes/__init__.py
cisaacstern: > I agree we should not make rich a hard dependency. Let's try to use rich (try: import rich) if it is installed and otherwise fall back on the builtin logger otherwise.

✔️  https://github.com/pangeo-forge/pangeo-forge-recipes/pull/287/commits/e25bb0def284d675bacc398d68499011e40cf0a2

rabernat:
The rtd build environment indicates that it is searching deep into the package history of distributed to resolve some dependency mismatch. I believe it can be fixed by more explicitly pinning package versions in docs/requirements.txt, or by switching to a conda-based environment.

PR #288, opened 2022-02-16
Replace storage attributes with `StorageConfig`

### LINKED ISSUE #276: Simplify storage configuration

cisaacstern:
Closes #276 based on design described by @rabernat in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/286#issuecomment-1040904877.

Hopefully helpful to @rwegener2's new user journey work in #281.


rabernat:
This is fantastic!

I see two paths we could take here.

- Fully flesh out this new feature in this PR by:
  - Updating the tutorials to use this new style of configuration
  - Writing a new documentation page about how to configure storage in different scenarios
- OR merge this as is (meaning that the documentation remains semi-broken) and leave those as TODOs for the near future

cisaacstern:
AFAICT, the [`FileNotFoundError`](https://github.com/pangeo-forge/pangeo-forge-recipes/runs/5211218855?check_suite_focus=true#step:9:197) causing the new `test_recipe_default_storage` to fail is caused by the fact that 

```python
import os
import tempfile
from fsspec.implementations.local import LocalFileSystem
from pangeo_forge_recipes.storage import FSSpecTarget

# Case 1 ------------------------------------------------------------------------------------------

def fsspectarget_factory():
    fs_local = LocalFileSystem()
    tempdir = tempfile.TemporaryDirectory()
    return FSSpecTarget(fs_local, tempdir.name)

target = fsspectarget_factory()
print(f"With tempdir defined inside factory function, {os.path.exists(target.root_path) = }")


# Case 2 ------------------------------------------------------------------------------------------

tempdir = tempfile.TemporaryDirectory()

def fsspectarget_factory():
    fs_local = LocalFileSystem()
    return FSSpecTarget(fs_local, tempdir.name)

target = fsspectarget_factory()
print(f"With tempdir defined outside factory function, {os.path.exists(target.root_path) = }")


# Case 3 ------------------------------------------------------------------------------------------

def tempdir_factory():
    tempdir = tempfile.TemporaryDirectory()
    return tempdir

tempdir = tempdir_factory()
print(f"With tempdir defined inside factory function (but not passed into FSSpecTarget), {os.path.exists(tempdir.name) = }")
```
```
With tempdir defined inside factory function, os.path.exists(target.root_path) = False
With tempdir defined outside factory function, os.path.exists(target.root_path) = True
With tempdir defined inside factory function (but not passed into FSSpecTarget), os.path.exists(tempdir.name) = True
```

So there appears to be something specific about passing a tempdir created within a function's local scope _into `FSSpecTarget`_, which results in the temporary directory being cleaned up...

cisaacstern:
@rabernat, setting aside for a moment the question of whether to document this now or later, wondering what you think of https://github.com/pangeo-forge/pangeo-forge-recipes/pull/288/commits/96c2edd21cf224e6ad7022a11b8224636e6eea9d. I'm aware that it's generally considered good practice to avoid `global`, but this was the only path I could find to addressing the scoping issues outlined in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/288#issuecomment-1041115070.

Interestingly, https://github.com/pangeo-forge/pangeo-forge-recipes/pull/286 didn't seem to exhibit this `TemporaryDirectory` scoping/cleanup problem (there were no tests on that PR, but I did run it locally and it seemed to work).

rabernat:
I'm confused by this and trying to investigate further myself. The tempdir should only be cleaned up when the variable is garbage collected. I would like to avoid globals. On the other hand tempfile itself does use globals - https://docs.python.org/3/library/tempfile.html.

rabernat:
Ok I think I understand what is going on here. The TemporyDirectory is getting garbage collected because we only save pass the `.name` attribute to fsspec. Consider this snippet, which I think reproduces the issue in a minimal way. I may have a solution.

```python
import tempfile
import os

def make_tmpdir(return_name_only=False):
    td = tempfile.TemporaryDirectory()
    assert os.path.exists(td.name)
    if return_name_only:
        return td.name
    else:
        return td 

d1 = make_tmpdir()
assert os.path.exists(d1.name)

d2 = make_tmpdir(return_name_only=True)
assert not os.path.exists(d2)
```

rabernat:
So now that the feature itself is done and tested, we need to decide what to do about the documentation.

cisaacstern:
Two missing steps on the feature side:

- Implement this for `HDFReferenceRecipe`
- Replace old storage attributes in `XarrayZarrRecipe` docstring

I can do both now.

@rwegener2, if you're able to make a quick pass through the docs, just to clean up any old syntax, that'd be great.

> Should the implementation now look like this, then?

Here's the best part! The user doesn't have to do anything. That factory function is passed into the recipe classes as a default factory ([official docs here](https://docs.python.org/3/library/dataclasses.html#dataclasses.field)) so as soon as you import a recipe into a notebook (or interpreter session, etc.), it automatically comes with temporary storage attached to it, and you can just jump directly into execution.

rabernat:
> it automatically comes with temporary storage attached to it, and you can just jump directly into execution.

The flipside of this is that, if the user wants to actually control the storage configuration more directly, they need some documentation about how to do this. Ideally we could document three different scenarios.
- Single machine local workflow - in this case, you just have to replace `storage_config.target` with your desired final destination
- Distributed execution on HPC - need to ensure that all 3 storage locations point to a shared filesystem
- Distributed execution in cloud - need to ensure that all 3 storage locations point to appropriate paths in object storage (S3, GCS, ABS, etc)

rwegener2:
> Here's the best part! The user doesn't have to do anything

That's pretty cool @cisaacstern !  I agree with @rabernat that we will need separate documentation describing how to specify your target location, but I think that's fine in a separate PR.

One question -- where does the output data get written in this new scheme?  I see that `recipe.target` is empty so I'm not sure where to look for the zarr.

cisaacstern:
> One question -- where does the output data get written in this new scheme? I see that recipe.target is empty so I'm not sure where to look for the zarr.

The default output data path is now `recipe.storage_config.target.root_path`

You can get an `fsspec` mapper to open it by just calling `rec.storage_config.target.get_mapper()`

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/29ba8e84b493630db882d298fb90d7fe06aa224b/tests/recipe_tests/test_XarrayZarrRecipe.py#L127

because the internal storage class defines it's own `get_mapper` method which uses `self.root_path` automatically

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/dd77bba689e1960ed7fb47a415d8b9505ccf1b56/pangeo_forge_recipes/storage.py#L97-L99

rabernat:
Yeah now that this is getting set totally automatically, it may be confusing to the user where to actually find their dataset. I wonder what's the best away to handle this...

rabernat:
I like that but I hesitate to introduce yet another new name "outpath". Why not just call the property `target`? I'm also concerned that the path itself is not that helpful without the accompanying filesystem. It would be better to have a fully qualified uri, e.g. `file://path` or `s3://path`.

rwegener2:
Is it possible to have the automatic target creation set the `recipe.target` `recipe.metadata_cache` and `recipe.input_cache` properties on the recipe in the process of setting itself up?  That would keep the effect of the automated process consistent with the effect of the manual process.

cisaacstern:
> It would be better to have a fully qualified uri, e.g. file://path or s3://path.

Good call, we can do that with something like

```python
@property
def target(self):
    return f"{self.storage_config.target.fs.protocol}://{self.storage_config.target.root_path}"
```

while we're at it, should we also throw in a shortcut to get the mapper, such as

```python
@property
def target_mapper(self):
    return self.target.storage_config.target.get_mapper()
```
?

rwegener2:
I like it. Should `metadata_cache` and `input_cache` also get the same functionality?  I don't see a strong practical need but consistency is nice for maintaining the mental model of the user.

cisaacstern:
> Should `metadata_cache` and `input_cache` also get the same functionality? I don't see a strong practical need but consistency is nice for maintaining the mental model of the user.

Very well articulated. I can see both sides, on the one hand consistency is good, and I can personally imagine having top-level accessors to these URIs as convenient for more advanced debugging. On the other hand, providing them sort of implies to the user that these are areas of the code they should be interested in, which is not entirely true.

Ryan, what's your preference? I'm happy either way.

rabernat:
Let's not expose the cache or metadata via a property.

cisaacstern:
@rwegener2 the following section of the tests

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/fe96c9d96b490f2aff5b811f9251ba474bac960f/tests/recipe_tests/test_XarrayZarrRecipe.py#L129-L135

shows the three ways one can now open the target, of which the third should be what we recommend to users. That is, after executing a `recipe`, the easiest way for the user to open the resulting dataset should be

```python
import xarray as xr

ds = xr.open_zarr(recipe.target_mapper)
```

LMK if you run into any other questions with the documentation! As you can presumably see, your discerning eye is resulting in major improvements to the user experience in real time! 🎉

rwegener2:
Alright I just pushed a commit that has updated tutorials for all the tutorials except the opendap tutorial.  In the case of an OpenDAP dataset using the default targets running `recipe.to_function()()` yields `ValueError: Can't cache opendap inputs`. 

My original thought was to just leave the OpenDAP tutorial alone and not use the new technique, but doing that seems to clashing with the default targets.  When I tried to add back in the old method (below) I got the same `ValueError` as when I omitted the manual file set up.
```
import tempfile
from fsspec.implementations.local import LocalFileSystem
from pangeo_forge_recipes.storage import FSSpecTarget, CacheFSSpecTarget, MetadataTarget

fs_local = LocalFileSystem()

target_dir = tempfile.TemporaryDirectory()
target = FSSpecTarget(fs_local, target_dir.name)

meta_dir = tempfile.TemporaryDirectory()
meta_store = MetadataTarget(fs_local, meta_dir.name)

recipe.target = target
recipe.metadata_cache = meta_store
recipe
```
Thoughts?

cisaacstern:
Great catch, @rwegener2! https://github.com/pangeo-forge/pangeo-forge-recipes/pull/288/commits/9f43aa55823fd5ec36294b08e2f6f69fa89d4e29 should fix this. For the OpenDAP tutorial, you should now be able to just delete the **Assign Targets** section and use our new default targets. LMK if not.

rabernat:
> recipe.target = target
> recipe.metadata_cache = meta_store

This syntax should no longer work after this PR. You would need to do 
```python
recipe.storage_config.target = target
recipe.storage_config.metadata = meta_store
```

> In the case of an OpenDAP dataset using the default targets running `recipe.to_function()()` yields `ValueError: Can't cache opendap inputs`.

This may be revealing a new bug. If you're getting this at execution time (as opposed to initialization time), you're hitting this code path

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/ce9d8b971159b850a739187bc2eaba773b14c52f/pangeo_forge_recipes/recipes/xarray_zarr.py#L143-L146

However, the `cache_inputs` flag should be getting checked at initialization time, here

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/ce9d8b971159b850a739187bc2eaba773b14c52f/pangeo_forge_recipes/recipes/xarray_zarr.py#L746-L750

As you can see from the block above, `cache_inputs` should be automatically getting set to `False` whenever we have opendap inputs.

We need to do some debugging to understand why this logic is failing.

rabernat:
I have resolved the opendap notebook issues in 73ace5d793a9dbfb5951236eed7c46400aead125.

cisaacstern:
@rwegener2 tells me the docs looks good now, and @rabernat previous gave a 👍 on the feature, so going to merge this now 🎉

PR #289, opened 2022-02-16
Try pinning more dependencies to fix rtd build

rabernat:


rabernat:
No systematic method here, just hopeful tweaking. Looks like rtd is working again.

At some point we need to revisit our dependency situation.

PR #293, opened 2022-02-18
Intro tutorial part2

rwegener2:
## What has been built?
This PR address the second part of Issue #280. It builds on PR #281 by creating Part 2 of an updated Introduction Tutorial sequence. Part 2 focuses on Running a recipe locally.

## How was it done?
* Created a new folder `introduction_tutorial` with the 1st part of the new tutorial
* updated`index.md` for the intro tutorials (1 more part to come)

## How can it be tested?
Clicking "Details" for the docs test that gets run by CI.

### chunksize and prunning
In Part 1 of the recipe we told the user to use `XarrayZarrRecipe(pattern, inputs_per_chunk=10)`. Once the recipe gets pruned and run, though, it errors with `AssertionError: invalid chunksize 10 and dimsize 2`. I can change `inputs_per_chunk` to 2 to stop this error, but I wanted to flag it in case others had better ideas. If the only use of pruned recipes is to run test chunks, it might be advantageous one day to allow pruned to skip over that error.  On a shorter term basis, though, if anyone has different ideas from switching to `inputs_per_chunk=2` I'm all about it.

### consolidated metadata warning
When running the recipe function (`recipe_pruned.to_function()()`) a `RuntimeWarning: Failed to open Zarr store with consolidated metadata, falling back to try reading non-consolidated metadata.` warning appears. It's not the end of the world but if anyone knows what I could change to stop the warning that would be great.

### logging output
The logging output is quite long and my current opinion is that it disrupts the flow of the tutorial without much gain. To deal with this I commented out the # setup_logging() line. This looks awkward in it's own way, though, so I'm not sure which way to go here.

### engine error on run
(Adding post docs-build)
running the recipe function (`recipe_pruned.to_function()()`) errors with `ValueError: unrecognized engine h5netcdf must be one of: ['netcdf4', 'store', 'zarr']`.  You can see it [here](https://pangeo-forge--293.org.readthedocs.build/en/293/introduction_tutorial/intro_tutorial_part2.html#run-the-recipe). I'm surprised this happens, since we run so many other recipes with the docs those libraries I would have thought have to be installed.

cc @cisaacstern

PR #296, opened 2022-02-21
Move netcdf4 and h5netcdf to required install

rwegener2:
This PR addresses an issue in the depenencies surfaced by the `ValueError: unrecognized engine h5netcdf must be one of: ['netcdf4', 'store', 'zarr']`  surfaced in #293  (see "engine error on run"  section of the PR description).


rabernat (approved):
Seems fine to me. At some point, we need to get our dependency story straight (minimum vs. optional dependencies). But for now, to make things "just work" out of the box, I am ok with this.

PR #308, opened 2022-03-02
Add ogp headers

rabernat:
This should hopefully make our site have nice twitter / discourse / etc previews using [Open Graph Protocol](https://ogp.me/).


rabernat:
I don't think this is working but I'm going to merge it anyway and go from there.

PR #310, opened 2022-03-03
Add links to sandbox in installation and tutorials

rwegener2:
## What was done
- updated the text on the Installation page to emphasize the sandbox
- Added a link to the installation page and to the sandbox at the top of each tutorial (incomplete)

## Notes

~@cisaacstern would you be able to tell me how to get the sandbox link from the button in [the sandbox repo](https://github.com/pangeo-forge/sandbox)? The existing binder link is incorrect.~ ~Update: I figure you're not awake yet but still scratch that. I'm rearranging some things.~ Update on the update: New commit, but I'm still not sure about generating a link. Perhaps we can talk about it at our meeting later.

PR #313, opened 2022-03-03
Add feedback from Paige

### LINKED ISSUE #312: Integrate feedback to intro tutorials
This fantastic feedback came from Paige. A PR is coming soon to make these changes.
## PART 1
**First page “Introduction Tutorial”**
- [x] it says that you expect folks to have “Pangeo-forge-recipes” installed. Can you add a link?
- [x] More clarity on this page if I am running the tutorial on my local computer, on the Cloud, on a Binder, etc. 
Maybe also specify where you expect folks to run the tutorial from vs. where they should actually work when using Pangeo Forge for real (e.g. use Binder for the demo, but when using Pangeo Forge normally run on your own computer) 
_Status_: (<- add "code won't be saved from sandbox" in the sandbox link PR. Add section discussing local dev vs. sandbox)

**Where should I write this code?**
- [x] Might be nice to include a bit more on why I would choose one over the other. The wording is a bit unclear/vague - is it best to just start with a text editor? Is there a reason Jupyter notebook would be better?

**Explore the structure**
- [x] “By putting the URL into a web browser…”
This is confusing - should I be copy/pasting the above link? The wording just seems to specifically indicate that I shouldn’t click the link, but rather copy/past. Maybe just say “By clicking on the above link, we can explore…”
I like the line “important takeaways are…” :)

**About the FilePattern object**
- [x] Under bullet point 2, the s of “keys” is not pink. There are a couple other places where this is inconsistent as well.

**Define a Format Function**
- [ ] In code block that defines `make_url(time)`
The %d is in pink - is this intentional?
_Status_: This is odd, but I'm not sure how to fix it.  Leaving as is for now.

## PART 2
**Create the Recipe object**
- [x] Typo in last line: 
should be “In more complex recipes additional arguments may get used, but…”
- [x] How do they know that setting `inputs_per_chunk` =10 results in chunks of about size 100MB? Maybe mention how they can check the size of their data, especially if they haven’t downloaded it yet

**Set up logging**
- [x] Could add another sentence of why logging is useful (it’s mentioned that it can be useful, but why? What is being logged?)

**Prune the recipe**
- [x] Typo in first line: “setup” should be “set up”
- [ ] How does `recip.copy_pruned()` decide how long the pruned time dimension is? Is it always 2 time steps? Is `copy_pruned()` kind of equivalent to running e.g. xarray_dataset.isel(time=slice(0,2))?
- [x] Is the `copy_pruned()` function only useful for demos like this? When would I use this in creating a real recipe?

**Run the recipe**
- [x] Typo in second-to-last sentence: “A” should be lower case (within the parentheses)
- [x] Typo in last sentence: “it can be run…”

**Code Summary**
- [x] Typo: should say “code written in part 2”

## General comments
- [x] I think it might be worth reminding the user what the overall goals/outcomes are for the different tutorial sections. Near the end of Part 2, there is a line saying that the time has come to run the recipe, but I realized that I wasn’t sure/forgot what the recipe did and why that was exciting. 
- [x] It seems like there are 2 main aspects of the Pangeo-Forge process: converting data to zarr (or some other ARCO format) and then uploading to the Cloud. In my eyes, it looks like parts 1 and 2 are for converting data to zarr and then part 3 is for adding to cloud. Maybe make that more clear in the first page of the Intro tutorial? (e.g. when outlining the 3 parts) I also know that you have thought about this WAY more than me, so if I’m wrong or what I’ve said doesn’t make sense, just ignore! :)
- [x] Adding an extra sentence about what a FilePattern does on the first page of the Tutorial (the overview page) might also be helpful, or at least at the top of the page for Part 1.
- [x] The text on the overview page of the tutorial appears to essentially be copied over at the top of the pages in parts 1 and 2. This is just fine, but I think having some sort of designation (e.g. a horizontal line) to let the reader know that that is text we have already read would be nice. I found myself starting to read it each time, realizing that I had already read it, but then having to scan around to figure out where the next text for that part of the tutorial starts.

rwegener2:
Closes #312 

changes are described in the ticket.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -13,10 +13,8 @@ If you have not added the conda forge channel or set it as your default channel 
 
 If those steps are complete you can install the library with `conda install pangeo-forge-recipes`.
 
-## Remote access
+## Pangeo Forge Sandbox
 
-Pangeo Forge also has a [code sandbox](https://github.com/pangeo-forge/sandbox) which has `pangeo-forge-recipes` installed and template documents for creatig a recipe. Click the button below to launch the sandbox using Binder:
-
-[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/pangeo-forge/sandbox/binder)
+Pangeo Forge also has a [code sandbox](https://github.com/pangeo-forge/sandbox) accessible online! The sandbox is an environment with `pangeo-forge-recipes` installed and template documents for creating a recipe. Links to both a Template workspace and a blank workspace with `pangeo-forge-recipes` installed are available in the [sandbox README](https://github.com/pangeo-forge/sandbox/blob/main/README.md).
 
 You'll also notice that the tutorials have Binder links. The binder links can be used to explore `pangeo-forge-recipes` without setting up a local environment.
--- a/docs/introduction_tutorial/index.md
+++ b/docs/introduction_tutorial/index.md
@@ -1,16 +1,25 @@
 # Introduction Tutorial
 
-Welcome back to the Pangeo Forge introduction tutorial!
+Welcome to the Pangeo Forge introduction tutorial!
 
+### The Structure
 This tutorial is split into three parts:
 
 1. Defining a `FilePattern`
 2. Defining a recipe and running it locally
 3. Setting up a recipe to run in the cloud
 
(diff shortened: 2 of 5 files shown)


rwegener2:
This PR now supersedes #310

PR #323, opened 2022-03-09
use fsspec simplecache

rabernat:
This PR would get rid of our custom local file copying logic in favor of fsspec's [simplecache mecahnism](https://filesystem-spec.readthedocs.io/en/latest/features.html#remote-write-caching).

After filing https://github.com/intake/intake-xarray/issues/116 I realized how easy it should be. @martindurant has been saying this for a while. 🙃

### DIFF
--- a/pangeo_forge_recipes/storage.py
+++ b/pangeo_forge_recipes/storage.py
@@ -263,14 +263,9 @@ def file_opener(
         logger.info(f"Opening '{fname}' directly.")
         opener = _get_opener(fname, secrets, **open_kwargs)
     if copy_to_local:
-        _, suffix = os.path.splitext(fname)
-        ntf = tempfile.NamedTemporaryFile(suffix=suffix)
-        tmp_name = ntf.name
-        logger.info(f"Copying '{fname}' to local file '{tmp_name}'")
-        target_opener = open(tmp_name, mode="wb")
-        _copy_btw_filesystems(opener, target_opener)
-        yield tmp_name
-        ntf.close()  # cleans up the temporary file
+        logger.info(f"Copying '{fname}' to local file")
+        with _get_opener("simplecache::" + fname, secrets, **open_kwargs) as fp:
+            yield fp.name
     else:
         logger.debug(f"file_opener entering first context for {opener}")
         with opener as fp:


rabernat:
Comprehensive test suite FTW! The test failure is due to HTTP credentials not being passed through via fsspec_open_kwargs.

Here is a minimal reproducer.

```python
import aiohttp
import fsspec 

url = "http://httpbin.org/basic-auth/foo/bar"
auth = aiohttp.BasicAuth(login='foo', password='bar')

with fsspec.open(url, auth=auth) as fp:
    data = fp.read()  # works
    
with fsspec.open("simplecache::" + url, auth=auth) as fp:
    data = fp.read()  # ClientResponseError
```

@martindurant - would you consider this an fsspec issue? Or is there a way to pass the credentials through?

martindurant:
Should be 
```python
with fsspec.open("simplecache::" + url, http=dict(auth=auth)) as fp:
    data = fp.read()  # ClientResponseError
```
because the URL now has two components, and fsspec needs to know which of these to send the kwargs to.

rabernat:
Great, thanks for chiming in!

The problem is that in Pangeo Forge (in contrast to the toy example I shared), we don't necessary know that it's an http link. We could try to parse that out, but that feels fragile. Is there a different way of invoking the simplecache?

martindurant:
Ah, I see. The system isn't designed for passing arbitrary arguments to "the second filesystem", but I'll have a think about what can be done.

fsspec uses the following to parse the URL pieces:
```python
x = re.compile(".*[^a-z]+.*")
bits = (
    [p if "://" in p or x.match(p) else p + "://" for p in path.split("::")]
    if "::" in path
    else [path]
)
```

martindurant:
I suppose you could do the following:
```python
In [16]: fs, _ = fsspec.core.url_to_fs(url, auth=auth)

In [17]: fs2 = fsspec.filesystem("simplecache", fs=fs)

In [18]: with fs2.open(url) as f:
    ...:     print(f.read())
    ...:
b'{\n  "authenticated": true, \n  "user": "foo"\n}\n'
```

rabernat:
What about this?

```python
fs, _, paths = fsspec.get_fs_token_paths(url, storage_options={'auth': auth})
cache_fs = fsspec.implementations.cached.CachingFileSystem(fs=fs)

with cache_fs.open(paths[0]) as fp:
    data = fp.read()
```

martindurant:
I'm pretty sure that amounts to the same thing.

PR #322, opened 2022-03-09
Implement `FilePattern.file_type`

cisaacstern:
Closes #320

Wanted to prioritize this because it will resolve:

- https://github.com/pangeo-forge/pangeo-forge-recipes/issues/315
- https://github.com/pangeo-forge/staged-recipes/issues/120

Defaulting `FilePattern.file_type` to `FileType.netcdf4` means that this PR is backwards compatible with recent releases and documentation. We might add a block quote note about this in the docs somewhere to the effect of:

> **Note**: We assume your inputs are NetCDF4 files. If they are NetCDF3, simply specify `FilePattern(..., file_type="netcdf3")`

The only other thing I think I'd like to do is consider wrapping the `xr.open_dataset` call in XarrayZarrRecipe in a try/except, so that we can raise a more descriptive error in situations like those that motivated this PR.

PR #324, opened 2022-03-10
Install and import from `kerchunk` (not fsspec-reference-maker)

cisaacstern:
I believe the fact that [upstream-dev tests are failing, despite an upstream fix in kerchunk](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/322#issuecomment-1064530544) may be related to the fact that we never migrated from installing and importing from the the fsspec-reference-maker namespace.

As a result, when we install the upstream dev requirements, we get a [kerchunk installation](https://github.com/pangeo-forge/pangeo-forge-recipes/runs/5503039808?check_suite_focus=true#step:10:173) _in addition to_ an [fsspec-reference-maker installation](https://github.com/pangeo-forge/pangeo-forge-recipes/runs/5503039808?check_suite_focus=true#step:10:123). Obviously, this is problematic.

I hope migrating to the `kerchunk` namespace for installation and import may solve this.


cisaacstern:
Status update:

- As is, this PR appears to solve the kerchunk upstream dev problems for the `3.8` build
- We're now hitting an separate error related to `nbmake` in the `3.9` build, which I've documented with a reproducer in
   - https://github.com/treebeardtech/nbmake/issues/63

Re: kerchunk, even though there is a Conda Forge feedstock available, I've kept it as a `pip` dependency, because installing from `conda-forge` alongside `pynio` (another one of our dependencies) causes problems.

The minimal reproducer is to build and activate an environment from

```yaml
# kerchunk-plus-pynio.yml

name: kerchunk-plus-pynio
channels:
  - conda-forge
dependencies:
  - python=3.8
  - kerchunk>=0.0.4
  - pynio
```
with 

```
$ conda env create --file kerchunk-plus-pynio.yml
```
Then call 
```
(kerchunk-plus-pynio) $ python3 -c "import Nio"
```
which replicates [this line](https://github.com/pydata/xarray/blob/eea76733770be03e78a0834803291659136bca31/xarray/backends/pynio_.py#L18) from the xarray pynio backend, and raises
```python-traceback
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/miniconda3/envs/kerchunk-plus-pynio-conda/lib/python3.8/site-packages/PyNIO/Nio.py", line 83, in <module>
    from _nio import *
ImportError: dlopen(/miniconda3/envs/kerchunk-plus-pynio-conda/lib/python3.8/site-packages/PyNIO/_nio.cpython-38-darwin.so, 0x0002): Library not loaded: @rpath/libtbb.dylib
  Referenced from: /miniconda3/envs/kerchunk-plus-pynio-conda/lib/libtiledb.dylib
  Reason: tried: '/miniconda3/envs/kerchunk-plus-pynio-conda/lib/libtbb.dylib' (no such file), '/miniconda3/envs/kerchunk-plus-pynio-conda/lib/libtbb.dylib' (no such file), '/miniconda3/envs/kerchunk-plus-pynio-conda/lib/libtbb.dylib' (no such file), '/miniconda3/envs/kerchunk-plus-pynio-conda/lib/libtbb.dylib' (no such file), '/miniconda3/envs/kerchunk-plus-pynio-conda/lib/libtbb.dylib' (no such file), '/miniconda3/envs/kerchunk-plus-pynio-conda/lib/libtbb.dylib' (no such file), '/miniconda3/envs/kerchunk-plus-pynio-conda/lib/libtbb.dylib' (no such file), '/miniconda3/envs/kerchunk-plus-pynio-conda/lib/python3.8/site-packages/PyNIO/../../../libtbb.dylib' (no such file), '/miniconda3/envs/kerchunk-plus-pynio-conda/lib/libtbb.dylib' (no such file), '/miniconda3/envs/kerchunk-plus-pynio-conda/lib/libtbb.dylib' (no such file), '/miniconda3/envs/kerchunk-plus-pynio-conda/lib/libtbb.dylib' (no such file), '/miniconda3/envs/kerchunk-plus-pynio-conda/lib/python3.8/site-packages/PyNIO/../../../libtbb.dylib' (no such file), '/miniconda3/envs/kerchunk-plus-pynio-conda/bin/../lib/libtbb.dylib' (no such file), '/miniconda3/envs/kerchunk-plus-pynio-conda/bin/../lib/libtbb.dylib' (no such file), '/usr/local/lib/libtbb.dylib' (no such file), '/usr/lib/libtbb.dylib' (no such file)
```
Is this worth raising on https://github.com/conda-forge/kerchunk-feedstock? I ask because the xarray docs indicate that pynio [no longer actively maintained](https://xarray.pydata.org/en/stable/user-guide/io.html#formats-supported-by-pynio) so maybe the solution is for us to just drop the `pynio` dependency?

martindurant:
Since the original fsspec-reference-maker surely has no users, I could remove it from PyPI, if that helps; but it's more involved fro conda-forge. I can also remove the alias package from kerchunk - but you won't notice that unless I make a new release (which would bring combine2 with it, since I should merge that).

rabernat:
For a fix for the nbmake issue, see https://github.com/pydata/xarray/pull/6350.

martindurant:
kerchunk's deps are extremely minimal, so I don't see how that's causing a problem.
```yaml
requirements:
  host:
    - pip
    - python >=3.7
  run:
    - fsspec
    - python >=3.7
    - ujson
```

cisaacstern:
Since `ujson` and `pynio` both have C components and the traceback in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/324#issuecomment-1064693202 indicates is a missing `.dylib` ... maybe it's `ujson`?

martindurant:
Don't know, maybe? It says the lib was missing, though, not some symbol clash. For reference, libtbb is some multiprocessing abstraction.

Could try explicitly listing ujson with some specific version, but that's shooting in the dark.

cisaacstern:
> For a fix for the nbmake issue, see https://github.com/pydata/xarray/pull/6350

Thanks Ryan.  https://github.com/pangeo-forge/pangeo-forge-recipes/pull/324/commits/4fa6feaef7dfa6e0f89e31d6f06c068dbe5abd0d also fixes this, because it bumps to `nbclient>=0.5.13`, which includes the https://github.com/jupyter/nbclient/pull/209 bugfix.

cisaacstern:
Martin, is the [removal of `MultiZarrToZarr.xarray_open_kwargs`](https://github.com/fsspec/kerchunk/compare/0.0.5..main#diff-4b40de22f27d5b5e8abb9b7d2cf1a9f874f3bdfa3bd0f21731a8a88cc68a3267L47) in `main` intended as a breaking change? We're now [failing here](https://github.com/pangeo-forge/pangeo-forge-recipes/runs/5514012784?check_suite_focus=true#step:9:1073) due to that

martindurant:
Yes, this is unavoidable. I am looking at this repo now for the fix. The signature is now:
```
    def __init__(self, path, coo_map=None, concat_dims=None, coo_dtypes=None,
                 identical_dims=None,
                 target_options=None, remote_protocol=None, remote_options=None,
                 inline_threshold=500, preprocess=None, postprocess=None):
```
where probably the only content of `xarray_open_kwargs` was `concat_dims`, which now becomes simply `concat_dims` by itself.

cisaacstern:
Re: `kerchunk`/`pynio` incompatibility, this is actually a `ujson`/`pynio` incompatibility as this environment breaks with the same traceback reported in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/324#issuecomment-1064693202:

```yaml
name: ujson-pynio
channels:
  - conda-forge
dependencies:
  - python=3.8
  - pynio
  - ujson
```
I'm a bit out of my depth in debugging this, but my hunch is that pynio being [unmaintained](https://github.com/conda-forge/pynio-feedstock/issues/100) is the problem, and maybe we want to just drop the pynio dependency?

martindurant:
Certainly, ujson is common enough that it could easily creep into the environment via some other dependency in another package later, so if we don't need pynio...

cisaacstern:
The `pynio` question is being tracked in #325 so the only remaining question here is how to handle the upstream `kerchunk` API change. (Note that for some reason the `3.8, upstream-dev` builds just passed, but that is a fluke of caching or somesuch thing. I re-ran these tests locally against `kerchunk@main` and of course they also fail due to the API change.)

cisaacstern:
Per offline conversation with @rabernat, I'll merge this as-is, with the knowledge that after the next `kerchunk` release, we'll need to pin `kerchunk` to `>={next release}` and change our `MultiZarrToZarr` call signature.

PR #326, opened 2022-03-11
Kerchunuk combine2

martindurant:
Following https://github.com/fsspec/kerchunk/pull/122


martindurant:
This will, of course, only pass with the upstream tests.

Note [this line](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/326/files#diff-56b442c85de53820352942579d40930246061aa3dc59d6ebd96569d1f254d747R50) - combining always uses all the available files, because they are in the file pattern, but the intent is to check against just the first one. Is there a way to select only the first file?

cisaacstern:
If the goal is to select only the first input file (in the concat dimension) prior to executing the recipe, then I think 

```python
recipe = HDFReferenceRecipe(file_pattern)
recipe = recipe.copy_pruned(nkeep=1)
```
should work?

cisaacstern:
If the goal is to select only the first input file (in the concat dimension) prior to executing the recipe, then I think 

```python
recipe = HDFReferenceRecipe(file_pattern)
recipe = recipe.copy_pruned(nkeep=1)
```
should work?

martindurant:
That'll do nicely, thanks @cisaacstern

pangeo_forge_recipes/recipes/reference_hdf_zarr.py
```diff
@@ -133,12 +130,19 @@ class HDFReferenceRecipe(BaseRecipe, StorageMixin, FilePatternMixin):
...
-        Only used if `file_pattern` has more than one file.
-    :param xarray_concat_args: kwargs passed to xarray.concat
-        Only used if `file_pattern` has more than one file.
+    :param identical_dims: coordiate-like variables that are assumed to be the
+        same in every input, and so do not need any concatenation
+    :param coo_map: set of "selectors" defining how to fetch the dimension coordinates
+        of any given chunk for each of the concat dimes. By default, this is the variable
+        in the dataset with the same name as the given concat dim, except for "var",
+        where the default is the name of each input variable. See the doc of
+        MultiZarrToZarr for more details on possibilities.
```
rabernat, line 139: Can we get intersphinx going on kerchunk so we can cross-reference its documentation here?

pangeo_forge_recipes/recipes/reference_hdf_zarr.py
```diff
@@ -149,15 +153,18 @@ class HDFReferenceRecipe(BaseRecipe, StorageMixin, FilePatternMixin):
     #  store_chunk, finalize_target. The strategy was used for NWM's 370k files.
 
     # TODO: as written, this recipe is specific to HDF5 files. kerchunk
-    #  also supports TIFF and grib2 (and more coming)
+    #  also supports TIFF, FITS, netCDF3 and grib2 (and more coming)
```
rabernat, line 156: Could we use the new `file_type` attribute for validation here? See #322

rabernat (commented):
Thanks for getting this started Martin! 🚀

pangeo_forge_recipes/recipes/reference_hdf_zarr.py
martindurant: What do I need to do?

pangeo_forge_recipes/recipes/reference_hdf_zarr.py
martindurant: Sure; but of course it would be nice to extend this recipe type to cope with all the possible kerchunk scanners.

cisaacstern (commented):
Martin, just two quick notes on the file type checks, the new enum object is `patterns.FileType`.

cisaacstern:
LGTM

PR #327, opened 2022-03-11
CI env: drop pynio, move kerchunk to conda deps

### LINKED ISSUE #325: Drop `pynio` dependency?
As discussed in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/324#issuecomment-1065339106, `pynio` is unmaintained and cannot be installed from Conda Forge without causing conflicts with other dependencies. (The present conflict is with `ujson` but others could arise.)

> A workaround for this conflict is, for now, to install `kerchunk` (which brings in `ujson`) from `pip`.

Do we rely on xarray's PyNIO backend for anything? The xarray docs note that it [conflicts with recent `netcdf4` versions](https://xarray.pydata.org/en/stable/user-guide/io.html#formats-supported-by-pynio).

If we don't use PyNIO, perhaps it's best to remove it, which would allow installing `kerchunk` from Conda Forge (and prevent other such conflicts in the future).

cisaacstern:
Closes #325

We can add `pynio` back into `3.8` env once https://github.com/conda-forge/pynio-feedstock/issues/114 is resolved, and back into `3.9` env once that issue plus https://github.com/conda-forge/pynio-feedstock/issues/100 is resolved.

### DIFF
--- a/ci/py3.8.yml
+++ b/ci/py3.8.yml
@@ -20,6 +20,7 @@ dependencies:
   - intake
   - intake-xarray
   - lxml
+  - kerchunk>=0.0.4
   - netcdf4
   - numcodecs
   - numpy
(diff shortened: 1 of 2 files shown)


cisaacstern:
Note failing upstream dev tests are due to kerchunk API change.

We can leave these broken until the next kerchunk release, at which point https://github.com/pangeo-forge/pangeo-forge-recipes/pull/326 will solve the problem.

PR #332, opened 2022-03-28
Slight tweaks to the main doc page

rabernat:


rabernat:
I'm confused why the RTD build is failing. 😞

cisaacstern:
Hmm it's failing on a Jinja import. I think `jinja2` is just a default dependency of sphinx/myst? But maybe we can pin it to an explicit version in docs/requirements.txt?

FWIW, as of the latest refactor, we're using Jinja substitution explicitly

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/95fb18defba9692b5c13c4815929ef035125075c/docs/conf.py#L56-L74

To get the [`{{ pangeo_forge_bot_header }}`](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/master/docs/pangeo_forge_cloud/pr_checks_reference.md#presence) to look like a "real" GitHub comment, [e.g. here](https://pangeo-forge.readthedocs.io/en/latest/pangeo_forge_cloud/pr_checks_reference.html#presence).

But (1) I think sphinx/myst would import `jinja2` either way?; and (2) obviously the build worked last week when that was merged.

rabernat:
Looks like this is https://github.com/sphinx-doc/sphinx/issues/10291

cisaacstern:
> We should also update main index page to reflect the new documentation organization, but that can wait for another PR.

I assume you mean updating [How the docs are organized](https://pangeo-forge.readthedocs.io/en/latest/#how-the-documentation-is-organized) section to reflect the Recipes/Cloud distinction?

If so, I'm open to it, though I'll note in it's current form I did think of that section as a deliberately alternate/thematic view into the organization, essentially the [Divio Docs System](https://documentation.divio.com) view, whereas the [High Level Overview](https://pangeo-forge.readthedocs.io/en/latest/what_is_pangeo_forge.html#high-level-overview) and the Table of Contents/Navigation present the structural Recipes/Cloud division.

So we could either stick with that distinction, but make it more obvious that that's the goal, or make the organization section explicitly a structural view rather than a thematic one. Either way, just wanted to clarify the initial idea.

PR #336, opened 2022-03-31
Revive pynio in ci envs

cisaacstern:
Following refresh of pynio feedstock (xref https://github.com/conda-forge/pynio-feedstock/issues/114#issuecomment-1082368476), I believe we can add this dependency back into the CI envs. Let's see if all the tests pass.

### DIFF
--- a/ci/py3.8.yml
+++ b/ci/py3.8.yml
@@ -28,10 +28,7 @@ dependencies:
   - pip
   - prefect
   - pydap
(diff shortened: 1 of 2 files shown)


cisaacstern:
The xarray docs note that the xarray pynio backend is deprecated, so closing this

https://docs.xarray.dev/en/stable/user-guide/io.html#formats-supported-by-pynio

PR #337, opened 2022-04-12
Experiment: How HDFReferenceRecipe could look as a Beam Pipeline.

alxmrs:
This is a prototype for using Apache Beam for the internal data model of Pangeo Forge Recipes. Here, I demo how HDFReferenceRecipe could be structured into modular components via composite Beam transforms.

xref: #256


martindurant:
Honestly, I find this pretty hard to read :|
Is it reasonable to set some time aside in the regular meeting to describe what's involved here?

pangeo_forge_recipes/recipes/reference_hdf_zarr.py
```diff
@@ -179,3 +209,6 @@ def _validate_file_pattern(self):
 
     def iter_inputs(self) -> Iterable[Hashable]:
         yield from self.file_pattern
+
+    def to_beam(self):
+        return OpenPattern(self.file_pattern) | FileNames() | ScanFiles(self) | WriteZarrReference(self)
```
rabernat, line 214: @martindurant - this is the part that is easy to read! 🤩

rabernat:
Beam programming takes some getting used to, but once it clicks, it seems super elegant to me.

> Is it reasonable to set some time aside in the regular meeting to describe what's involved here?

Alex, Charles and I met today. We can discuss more at the next coordination meeting. We need to evaluate the pros and cons.

cisaacstern:
Thanks for this experiment! Closing since we now have all this (in some form) in `0.10.0` 😄

PR #343, opened 2022-04-20
Use direct string interp in tutorial 1

leifdenby:
Use `...{time:%Y%m}...".format(time=time)` in place of creating an intermediate string with YYYYmm.


rabernat:
Thanks so much for this @leifdenby!

I realized that we are not actually running this tutorial in CI, so we don't have an automated way to check that this actually works. Have you tried it and verified that it does work?

leifdenby:
> I realized that we are not actually running this tutorial in CI, so we don't have an automated way to check that this actually works. Have you tried it and verified that it does work?

Yes, I re-ran the whole notebook afresh and all cells validated. That's why the execution count number changed, because previously the first cell had execution count as `2`.

I'm happy to wait to have this merged though if you're planning on setting up testing of the tutorial notebooks too.

docs/introduction_tutorial/intro_tutorial_part1.ipynb
```diff
@@ -102,9 +102,9 @@
     "### Create a URL Template String\n",
```
rwegener2, line 102: <p>I would suggest we keep these as they were before. The idea here is to consider conceptually how we are modifying the strings, not get bogged down on having to read the strftime sytnax yet.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pangeo-forge/pangeo-forge-recipes/pull/343/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

rwegener2, line 102: <p>I like the way you made URL_FORMAT a variable!  I think it makes what is happening in the return value clearer.</p><p>The one change I would make would be to define the URL_FORMAT variable inside the function. Otherwise a user could change URL_FORMAT outside the function and change what happens inside the function.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/pangeo-forge/pangeo-forge-recipes/pull/343/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

rwegener2:
Thanks for making this PR @leifdenby! I left a few small comments, but I think in general it looks good. Let me know if anything is unclear!

docs/introduction_tutorial/intro_tutorial_part1.ipynb
leifdenby: <p>I don't quite understand what you mean by "keep these". Do you mean revert the text back to the original here? I was changing it because I am suggesting to use direct string interpolation here (rather than have placeholders that expect variables with names <code>{yyyymm}</code> for example). I also think having a variable called <code>yyyymm</code> in the string is a bit confusing because then the reader might think that the way to achieve the <code>%Y%m</code> format is to call your variable <code>yyyymm</code>.</p><p> </p><p>Maybe I could just change the text to </p><p> </p><blockquote>where <code>{time:%Y%m}</code> and <code>{time:%Y%m%d}</code> will change for each file by changing the value of <code>time</code>. Of the three ...</blockquote>

leifdenby: <p>True, the user could change the value. I only made it global because there is a convention in python I think to put constant variables in all-caps in the global namespace. It makes it more explicit what things are constants and what are not in code, and avoids reallocating the string every time it is used. Also, other modules can then import this URL_FORMAT if needed somewhere else. But I'm happy to move the string inside the function if you prefer :)</p>

rwegener2: Okay, I see your logic there. 👍🏻

docs/introduction_tutorial/intro_tutorial_part1.ipynb
```diff
@@ -102,9 +102,9 @@
...
     "We can generalize the URL to say that OISST datasets are accessed using a URL of the format:\n",
     "\n",
-    "`https://www.ncei.noaa.gov/data/sea-surface-temperature-optimum-interpolation/v2.1/access/avhrr/{year_month}/oisst-avhrr-v02r01.{year_month_day}.nc`\n",
+    "`https://www.ncei.noaa.gov/data/sea-surface-temperature-optimum-interpolation/v2.1/access/avhrr/{time:%Y%m}/oisst-avhrr-v02r01.{time:%Y%m%d}.nc`\n",
     "\n",
-    "where `{year_month}` and `{year_month_day}` change for each file. Of the three dimensions of this dataset - latitude, longitude and time - the individual files are split up by time.\n",
+    "where `{time:%Y%m}` and `{time:%Y%m%d}` change for each file (we're using direct string interpolation of `datetime` objects here, anything that [strftime](https://strftime.org/) supports can be put in). Of the three dimensions of this dataset - latitude, longitude and time - the individual files are split up by time.\n",
```
rwegener2, line 107:
```suggestion
    "where `{time:%Y%m}` (representing a year and a month) and `{time:%Y%m%d}` (representing a year, month, and day) change for each file. (We're using direct string interpolation of `datetime` objects here, anything that [strftime](https://strftime.org/) supports can be put in). Of the three dimensions of this dataset - latitude, longitude and time - the individual files are split up by time.\n",
```

docs/introduction_tutorial/intro_tutorial_part1.ipynb
rwegener2: To me, the main learning goal at this early stage in the tutorial is not about Python, it conceptual. I think we want to make sure the learner understands that the fundamental premise of pangeo forge recipes is that we are accessing predictably named urls that are organized according to the structure of the dataset. We aren't teaching code at this point in the tutorial, so my original comment was intended to say that we start using direct string interpolation when we actually start coding in the Building a FilePatttern section.

But that aside, I can see how it would be better to be consistent in how we write the template string. I like the clarifying text that you put in and I added a few more quick words that I hope will keep people on a conceptual level at this point in the tutorial. So this was all a looong way of saying I think it looks good!

Thanks again for contributing this!

rwegener2 (approved):
I made that suggestion, which you can take or leave. Thanks again!

Issue #342, opened 2022-04-20
kerchunk/hdfreference tutorial is broken


cisaacstern:
For details, see:

> https://github.com/pangeo-forge/pangeo-forge-recipes/runs/6086899678?check_suite_focus=true

It appears that this tutorial was not updated to reflect changes to the `HDFReferenceRecipe` signature made in #326.

@martindurant, how would we adapt [this tutorial](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/master/docs/pangeo_forge_recipes/tutorials/hdf_reference/reference_cmip6.ipynb) to reflect the new signature?

PR #344, opened 2022-04-20
Update for new kerchunk

martindurant:
Fixes https://github.com/pangeo-forge/pangeo-forge-recipes/issues/342

Note that I **think** the times come out correctly, so the section "Problem with Time Encoding" can be removed. Please can someone check?


cisaacstern:
Thanks, Martin! I'll run the check now.

cisaacstern:
Tests pass to this LGTM! Thanks again for the quick fix

cisaacstern:
> Note that I think the times come out correctly, so the section "Problem with Time Encoding" can be removed. Please can someone check?

Oh, I realized I didn't take a specific look at this. Let's follow up in another Issue/PR if needed.

PR #350, opened 2022-04-28
Add prefect dask wrapper executor

### LINKED ISSUE #347: Create new executor to run Dask inside Prefect
Our large-scale deployments have revealed that prefect can't handle flows with > 1000 tasks. Prefect also charges per task. However, it is difficult to abandon prefect completely in the short term because we rely on it for orchestration.

As a workaround, at our last meeting we discussed creating a new executor that simply wraps the dask executor in a single prefect task. This would hopefully give us prefect's convenient orchestration plus the plain dask executor's better scalability.

rabernat:
Closes #347

To try this out:

```python
flow = recipe.to_prefect(wrap_dask=True)
```

### DIFF
--- a/pangeo_forge_recipes/executors/prefect.py
+++ b/pangeo_forge_recipes/executors/prefect.py
@@ -3,6 +3,7 @@ from typing import Any, Dict, List, Optional
 from prefect import Flow, task, unmapped
 
 from .base import Pipeline, PipelineExecutor, StageAnnotations, StageAnnotationType
+from .dask import DaskPipelineExecutor
 
 
 def annotations_to_task_kwargs(annotations: Optional[StageAnnotations]) -> Dict[str, Any]:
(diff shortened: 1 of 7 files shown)


pangeo_forge_recipes/executors/prefect.py
```diff
@@ -43,3 +44,22 @@ def compile(pipeline: Pipeline):
     @staticmethod
     def execute(flow: Flow):
         return flow.run()
+
+
+@task
+def run_pipeline_with_dask(pipeline):
+    delayed = DaskPipelineExecutor.compile(pipeline)
+    # how do we make sure this uses the right dask cluster?
+    delayed.compute()
```
martindurant, line 53: Can we get the cluster client in scope here? Dask will default to the "current" distributed client, the most recently created one.

martindurant: `distributed.client._global_clients` is a weak dict containing all the clients that have been instantiated. We could ensure that there is exactly one that is still active. I'm no *too* worried about it, though, since there is only one cluster, and the recipes themselves should not be creating any.

rabernat:
Hmmm looks like the tests turned up and error that I did not get locally.

rabernat:
The tests that failed is: tests/recipe_tests/test_XarrayZarrRecipe.py::test_recipe[execute_recipe_with_prefect-netCDFtoZarr_subset_recipe-netcdf_local_file_pattern_sequential-execute_recipe_prefect_wrapper-0-netcdf_local_paths_sequential_2d]

I did not get this the first time around, but now it turns out I can **intermittently** reproduce it.

rabernat:
Disturbingly, I restarted the tests and now everything worked. This confirms we are dealing with an **intermittent** h5py-related error. If I had to guess, I'd say that it's something about threads. My hope is that running things through a distributed scheduler helps.

PR #356, opened 2022-05-04
Fix "Create a recipe" link in README (current link is broken)

asross:
I think it should go [here](https://pangeo-forge.readthedocs.io/en/latest/pangeo_forge_recipes/recipe_user_guide/index.html)!


rabernat:
Good catch!

Issue #37, opened 2021-01-22
Support incremental appending


rabernat:
Currently, when a recipe is run, it will always cache all of the inputs and write all of the chunks. However, it would be nice to have an option where, if the target already exists, it only write NEW chunks. This raises some design questions.

- Currently, the target is never read until we start to execute the recipe (not until the `prepare_target` stage). However, for this to work, the `iter_inputs()` and `iter_chunks()` methods needs to know _which_ inputs and chunks to process. In order to build the pipeline for execution, this information needs to already be inside the recipe object. So this implies that we need open the target in `__post_init__`. Could this cause problems?
- How do we align the recipe with the target? For the standard [NetCDFZarrSequential](https://pangeo-forge.readthedocs.io/en/latest/tutorials/netcdf_zarr_sequential.html) recipe, it may be as simple as looking at the length of the sequence dimension: if the target has 100 items but the recipe has 120, we assume the last 20 need to be appended. But are there edge cases to worry about?

This intersects a bit with the "versioning" question in #3. 

If we agree on the answers to the questions above, I think we can move ahead with implementing incremental updates to the `NetCDFZarrSequentialRecipe` class.

rabernat:
Appending is significantly more complicated for the case discussed in #50: variable items per input. In this case, we don't know the size of the target dataset from the outset, so we can't use simple heuristics like the one I proposed above to figure out the append region.

Maybe recipes can implement their own methods for examining the target and determining which input chunks are needed? For that, it seems like the recipe would have to know more about the inputs than just a list of paths. For instance, it might have to understand time. What if `input_urls` were a dictionary rather than list, and the keys held some semantic meaning that could be used to compare to the target?

davidbrochart:
Maybe instead of looking at what has been produced, we could look at what has been consumed? For instance, after running a recipe, we could store the list of input files that were processed, so that when we get a new list next time the recipe is run, we look at the already processed input files and restore a "resuming" state. That could be done by having a "dry run", that would run the recipe without actually producing anything. We would still have the list of total processed input files, which might be needed when we "finalize" the target.
I don't know where we could store the list of processed input files, probably alongside the target, that seems the more natural. On the source side, tying a recipe to a target doesn't seem right.

rabernat:
This is a good idea David. Perhaps we could store the list of input files directly in the target dataset metadata itself (`attrs`). This would be useful for incremental appending but also for general provenance tracking.

davidbrochart:
[...]

PR #359, opened 2022-05-06
Persist execution context in storage target

cisaacstern:
This is a WIP to address @rabernat's suggestion in https://github.com/pangeo-forge/pangeo-forge-recipes/issues/37#issuecomment-1118899827 that we persist execution context information in the storage target, which is the first part of a solution for #37 and https://github.com/pangeo-forge/user-stories/issues/5.

Here's a rough idea of how this might look:

- `BaseRecipe.get_execution_context` compiles a dict of metadata we want to capture
- Each recipe class individually elects when/where this will be persisted to the target. The stage in which this occurs will be specific to each recipe class (`finalize_target` for XarrayZarrRecipe, a different stage for HDFReferenceRecipe, etc.). As will the exact location in the target (`.zmetadata` for  XarrayZarrRecipe, somewhere else for other storage formats, etc.).


pangeo_forge_recipes/recipes/base.py
```diff
@@ -50,6 +52,14 @@ def to_beam(self):
     def sha256(self):
         return dataclass_sha256(self, ignore_keys=self._hash_exclude_)
 
+    def get_execution_context(self):
+        return dict(
+            # See https://stackoverflow.com/a/2073599 re: version
+            version=pkg_resources.require("registrar")[0].version,
```
rabernat, line 58: Why track the registrar version? Ideally the target contents don't depend on the registrar. I was thinking we could track the pangeo_forge_recipes package version.

cisaacstern: Good catch! I copy-and-pasted this from the registrar, and forgot to update that.

Also this comment reminds me to add this PR to my TODO list of the week. 😄

cisaacstern: fixed in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/359/commits/3ca6066b4f9ad6b19747f70a0c5ae11b779589da

cisaacstern:
The latest test failures are from `test_XarrayZarrRecipe.py::test_process`, and stem from the fact that in #349 we punted on how to hash preprocessor functions. In this PR, we're calling `recipe.get_execution_context()` (which includes a `recipe.sha256()` call) on every execution, so the tests which include preprocessors fail. The options would seem to be:

1. Find a way to deterministically hash Python functions. (I've previously wondered if the function objects `__code__` attribute might offer some path forward here...)
2. Don't set a `"pangeo-forge:recipe_hash"` if the instance contains a preprocessor.
3. Do set a `"pangeo-forge:recipe_hash"` for recipes with a preprocessor, but exclude the preprocessor from the calculation. We might think of this as thematically similar to the `unsafe_hash` option for built-in dataclasses.

Preprocessors are a very useful feature, as (will be) hashes, so my preference would be to try to find a way to achieve option 1 if possible. I will take a brief pause on this PR to look into this, but if this seems infeasible, as a backup plan we could move forward with option 2 (not setting hashes for recipes with preprocessors). It seems to me that option 3 could lead to some very confusing situations which we'd prefer to avoid.

tests/recipe_tests/test_XarrayZarrRecipe.py
```diff
@@ -120,7 +131,7 @@ def test_recipe(recipe_fixture, execute_recipe):
     rec = RecipeClass(file_pattern, **kwargs)
     execute_recipe(rec)
     ds_actual = xr.open_zarr(target.get_mapper()).load()
-    xr.testing.assert_identical(ds_actual, ds_expected)
+    xr.testing.assert_identical(drop_execution_context_attrs(ds_actual), ds_expected)
```
rabernat, line 134: I'm not a fan of having to include this statement in every test.

Instead, let's write a wrapper around `xr.testing.assert_identical` like this

```python
def assert_identical(ds1, ds2):
    xr.testing.assert_identical(
        drop_execution_context_attrs(ds1),
        drop_execution_context_attrs(ds2)
    )
```

and call it from the individual tests.

docs/pangeo_forge_recipes/recipe_user_guide/execution.md
```diff
@@ -91,3 +91,26 @@ with beam.Pipeline() as p:
...
+    "pangeo-forge:recipe_hash": "{recipe hash as returned by `recipe.sha256()`}"
+    "pangeo-forge:inputs_hash": "{file pattern hash as returned by `recipe.file_pattern.sha256()`}"
+}
+```
+
+Each recipe class defines where to store this metadata:
+
+- `XarrayZarrRecipe`: Added to Zarr group attributes, and therefore also available via the
+`xarray.Dataset.attrs` when opening Zarr stores with xarray.
+- `HDFReferenceRecipe`:
```
rabernat, line 113: It should be straightforward to add the metadata to the kerchunk dataset as well. But we can leave that for a future PR.

rabernat (commented):
This is looking great! Just one nit pick about the testing syntax.

docs/pangeo_forge_recipes/recipe_user_guide/execution.md
cisaacstern: 👍 [62183f3](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/359/commits/62183f35badc963d36b0622954d03d76865a494f)

tests/recipe_tests/test_XarrayZarrRecipe.py
cisaacstern: 👍 b781e30

cisaacstern:
Thanks for the review, Ryan. `inspect.getsource` is very cool, hadn't seen it before! I think that's the last thing we need here, will add that now.

rabernat (commented):
Sorry I never merged this! I just see one small issue, then LGTM.

tests/recipe_tests/test_XarrayZarrRecipe.py
```diff
@@ -21,6 +21,26 @@
 from pangeo_forge_recipes.storage import MetadataTarget, StorageConfig
 
 
+def drop_execution_context_attrs(ds: xr.Dataset) -> xr.Dataset:
+    """Drop pangeo-forge execution context attrs from a dataset."""
+
+    ds_copy = ds.copy(deep=True)
```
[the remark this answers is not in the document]
cisaacstern: I'm not sure! It's "just" the tests so happy to use regular copy and see how it goes.

TBH, I've now run into enough weird behavior of copied dictionaries that are _not_ deep copies (due to shared references with the original, I guess), that I just reflexively make all dictionary copies deep now. 😅

cisaacstern, line 27:
```suggestion
    ds_copy = ds.copy()
```

pangeo_forge_recipes/recipes/base.py
KoscheyAV1: []()

PR #369, opened 2022-05-25
Beam refactor

### LINKED ISSUE #256: Should we just adopt xarray-beam as our internal data model?

rabernat:
Closes #256 

Just a very first start, based on discussion today.

### DIFF
--- a/pangeo_forge_recipes/recipes/xarray_zarr_beam.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr_beam.py
@@ -1,0 +1,43 @@
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass
+from typing import Any, Optional, Tuple
+
+import apache_beam as beam
+
+from ..patterns import Index
+from ..storage import CacheFSSpecTarget, OpenFileType, get_opener
+
+logger = logging.getLogger(__name__)
+
+
+IndexedElement = Tuple[Index, Any]
+# can we use a generic, e.g Indexed[xr.Dataset]?
+
+
+# This has side effects if using a cache
+@dataclass
+class OpenWithFSSpec(beam.PTransform):
+    """Write keyed chunks to a Zarr store in parallel."""
+
+    cache: Optional[CacheFSSpecTarget] = None
+    secrets: Optional[dict] = None
+    open_kwargs: Optional[dict] = None
+
+    def _open_with_fsspec(self, element: IndexedElement) -> Tuple[Index, OpenFileType]:
+        key, fname = element
+        open_kwargs = self.open_kwargs or {}
+        if self.cache is not None:
+            logger.info(f"Opening '{fname}' from cache")
+            # this has side effects
+            self.cache.cache_file(fname, self.secrets, **open_kwargs)
+            open_file = self.cache.open_file(fname, mode="rb")
+        else:
+            logger.info(f"Opening '{fname}' directly.")
+            open_file = get_opener(fname, self.secrets, **open_kwargs)
+        open_file
+        return key, open_file
+
+    def expand(self, pcoll):
+        return pcoll | "Open with fsspec" >> beam.Map(self._open_with_fsspec)
--- a/pangeo_forge_recipes/recipes/xarray_zarr_beam.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr_beam.py
@@ -19,7 +19,7 @@ IndexedElement = Tuple[Index, Any]
 # This has side effects if using a cache
 @dataclass
 class OpenWithFSSpec(beam.PTransform):
-    """Write keyed chunks to a Zarr store in parallel."""
+    """Open indexed items from a FilePattern with FSSpec, optionally caching along the way."""
 
     cache: Optional[CacheFSSpecTarget] = None
     secrets: Optional[dict] = None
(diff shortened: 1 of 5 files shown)


pangeo_forge_recipes/recipes/xarray_zarr_beam.py
```diff
@@ -0,0 +1,43 @@
...
+import apache_beam as beam
+
+from ..patterns import Index
+from ..storage import CacheFSSpecTarget, OpenFileType, get_opener
+
+logger = logging.getLogger(__name__)
+
+
+IndexedElement = Tuple[Index, Any]
+# can we use a generic, e.g Indexed[xr.Dataset]?
```
alxmrs, line 16: Yes, you can!  :) 

https://beam.apache.org/documentation/sdks/python-type-safety/#defining-generic-types

alxmrs (commented):
Looking good so far :)

rabernat:
I have pushed the `beam-refactor` branch to this remote. We are going keep iterating on it there via future PRs. Will merge to main at some far future date when it is all actually working. Going to close this for now.

PR #370, opened 2022-05-27
try opening with contextmanager

rabernat:
This is one possible workaround for https://github.com/fsspec/filesystem_spec/issues/579

This fails with the following error

```
apache_beam/coders/coder_impl.py:268: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

x = <xarray.Dataset>
Dimensions:  (time: 1, lat: 18, lon: 36)
Coordinates:
  * time     (time) datetime64[ns] 2010-01-01
 ...
    bar      (time, lat, lon) int64 ...
    foo      (time, lat, lon) float64 ...
Attributes:
    conventions:  CF 1.6

>   lambda x: dumps(x, protocol), pickle.loads)

../../anaconda/envs/pangeo-forge-recipes/lib/python3.8/site-packages/apache_beam/coders/coders.py:802: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <File-like object HTTPFileSystem, http://127.0.0.1:53014/000.nc>

    def __reduce__(self):
        return (
            reopen,
            (
                self.fs,
                self.url,
                self.mode,
                self.blocksize,
>               self.cache.name,
                self.size,
            ),
        )
E       AttributeError: 'NoneType' object has no attribute 'name'

../../anaconda/envs/pangeo-forge-recipes/lib/python3.8/site-packages/fsspec/implementations/http.py:637: AttributeError

During handling of the above exception, another exception occurred:

pcoll_xarray_datasets = <_ChainedPTransform(PTransform) label=[Create|OpenWithFSSpec|OpenWithXarray] at 0x7ff11a6e23d0>

    def test_OpenWithXarray(pcoll_xarray_datasets):
        with TestPipeline() as p:
[...]

PR #371, opened 2022-06-01
Add start / stop fields to DimIndex

cisaacstern:
Closes #367 when complete

PR #373, opened 2022-06-01
Try to actually load the dataset

rabernat:
This adds and important step that was left out of #370: actually loading the dataset. 

Unfortunately this reveals more new errors related to serializability of fsspec objects.


cisaacstern:
> left out of https://github.com/pangeo-forge/pangeo-forge-recipes/pull/371

for those following along at home, I think the reference here is to #370?

PR #377, opened 2022-06-14
Schema aggregation

### LINKED ISSUE #52: Concatenate across multiple dimensions
Currently we only support sequential inputs, along a single dimension, commonly time.

```
foo.time0.nc
foo.time1.nc
...
foo.timeN.nc
```

However, many potential recipes instead multiple axes in the files names, e.g. time and ensemble member:
```
foo.member0.time0.nc
foo.member0.time1.nc
...
foo.member1.time0.nc
foo.member1.time1.nc
...
foo.memberM.timeN.nc
```

This raises many of the same issues as #39.

rabernat:
This is a big PR which implements what I consider to be the hardest thing we have to do: combine schemas across multiple dimensions into a single dataset schema. Although it is not [yet] tested, I think the logic implemented here will resolve #52.

One potentially disruptive aspect of this PR is that it redefines the core Index object emitted by FilePattern and used throughout the package from

```python
class Index(FrozenSet[DimIndex]):
```

to
```python
class Index(Dict[DimKey, DimVal])
```

However, I consider these indexes as internal, not part of public API, so we can change them.


pangeo_forge_recipes/aggregation.py
```diff
@@ -0,0 +1,154 @@
...
+    return {"attrs": None, "coords": None, "data_vars": None, "dims": None, "chunks": None}
+
+
+@dataclass
+class XarrayCombineAccumulator:
+    schema: dict = field(default_factory=empty_xarray_schema)
+    concat_dim: Optional[str] = None
+
+    def add_input(self, s: dict, position: int) -> None:
+        s = s.copy()  # avoid modifying input
```
alxmrs, line 23: Is a deepcopy prudent here?

pangeo_forge_recipes/aggregation.py
```diff
@@ -0,0 +1,154 @@
...
+    if v1 is None:
+        return v2
+    all_vars = set(v1) | set(v2)
+    new_vars = {}
+    for vname in all_vars:
+        if vname not in v1:
+            new_vars[vname] = v2[vname]
+        elif vname not in v2:
+            new_vars[vname] = v1[vname]
+        else:
```
alxmrs, line 130: When will the case be reached? As I understand it, `all_vars` is either in v1 or v2.

pangeo_forge_recipes/combiners.py
```diff
@@ -0,0 +1,76 @@
...
+        accumulator.add_input(schema, position)
+        return accumulator
+
+    def merge_accumulators(
+        self, accumulators: Sequence[XarrayCombineAccumulator]
+    ) -> XarrayCombineAccumulator:
+        if len(accumulators) == 1:
+            return accumulators[0]
+        # mypy did not like sum(accumulators)
+        return reduce(lambda a, b: a + b, accumulators)
```
alxmrs, line 73: Nit: you could use `operator.add`. I expect this to be a bit better, since it won't carry as much data in a closure.

tests/test_combiners.py
```diff
@@ -0,0 +1,172 @@
...
+
+    with pipeline as p:
+        input = p | pcoll
+        output = input | beam.CombineGlobally(
+            CombineXarraySchemas(DimKey(name=concat_dim, operation=CombineOp.CONCAT))
+        )
+        assert_that(output, has_correct_schema(expected_schema))
+
+
+def test_NestDim(schema_pcoll_concat_merge, pipeline):
```
alxmrs, line 107: I'd like to better understand the use cases related to nested dimensions. Can you explain how these might be found in the wild?

alxmrs (commented):
Mostly questions in this review, and a few nits.

pangeo_forge_recipes/aggregation.py
rabernat: Good question. The only reason I make a copy in the first place is to add the `chunks` key. I need to `copy` for the test suite, where schema fixtures are shared between different invocations of the test. In the wild, I imagine that the input `s` will almost always be discarded after calling `add_input`. The dicts can potentially be very large, so there would be a performance cost to `deepcopy`.

alxmrs: That makes sense to me. If you don't depend too much on the values (esp, nested values), then a copy sounds sufficient.

pangeo_forge_recipes/aggregation.py
alxmrs: Thanks!

tests/test_combiners.py
alxmrs: Thanks, this helps me understand what's going on better. I think nesting is a fine terminology, though I'll let you know if I think of something better. I still am a bit fuzzy on all of the details, but it occurs to me: How about we have the transforms related to nesting be private? Since users will probably only make use of the results of some top level `GroupByKey` or combiner, we don't want to risk messing things up with internal-only transforms.

Issue #376, opened 2022-06-06
List of beam PTransforms to implement to recreate XarrayZarrRecipe


rabernat:
When the Beam refactor is complete (#256), the Pangeo Forge public API will look quite different. Broadly, we will export three main things:

1. High-level recipe "builders" that work identically or very similarly to our existing Recipe implementation. The precise API and internal structure for these is not yet determined. Will they be Beam PCollections? Dataclasses? We should experiment with some different design patterns?
2. Beam PTransforms that can be composed together to create recipes. Following the [Beam Ptransform style guide](https://beam.apache.org/contribute/ptransform-style-guide/), these PTransforms should be lightweight wrappers around generic python functions, which themselves will be exposed as...
3. General-purpose plain functions

A sketch of how 2 and 3 might look is already underway in #375.

Let's use this issue brainstorm all of the PTransforms we imagine we will need to implement to get feature complete with XarrayZarrRecipe

- [ ] `FilePatternSource` - Right now we are just using `beam.Create(pattern.items()` to start our pipelines. Might be better to follow the docs for [Custom I/O connectors](https://beam.apache.org/documentation/io/developing-io-overview/)
- [x] `OpenWithFSSpec` - Turn whatever comes out of the FilePattern into an FSSpec OpenFile object. Implemented in #375
- [x] `OpenWithXarray` - Turn an OpenFile or URL PCollection into an Xarray dataset PCollection (without loading into memory if possible). Implemented in #375.
- [x] `InferXarraySchema` - Takes a collection of Xarray datasets and figures out a schema for the target dataset. Implemented in #377.
- [x] `PrepareZarrTarget` - Takes that schema and use it to initialize the Zarr target. This is a singleton operation (single item PCollection). Can we make that explicit in beam? Implemented in #379
- [ ] `RechunkForTarget` - Takes the Xarray Dataset PCollection, plus a specification of the target chunks, and returns a new PCollection that is evenly aligned with the target chunks
- [ ] `ToZarr` - Here we could possibly just use xarray-beam `ChunksToZarr`. In progress in #391

alxmrs:
Re: `FilePatternSource`: You may be able to follow some examples of I/O connectors from the Geobeam project: https://github.com/GoogleCloudPlatform/dataflow-geobeam/blob/main/geobeam/io.py#L356

alxmrs:
[...]

PR #391, opened 2022-07-27
Zarr fragment writers

rabernat:
This implements one of the last few key pieces of #376: writing dataset fragments to Zarr in parallel. Still lots TODO here, namely:
- [ ] Better testing around chunks. We are no longer using locks for writing, so we have to take care to explicitly align writes with chunks. This PR does not implement rechunking.
- [ ] Unit tests for `IndexItems` (and maybe move its core functionality to a different module)
- [ ] Improve docstrings for transforms
- [ ] Incorporate feedback on the API. Do we like the names of the functions, PTransforms, and their arguments? All of this is free to change.
- [ ] Parametrize the 💩 out of `test_end_to_end.py::test_xarray_zarr`

I don't know why there are so many commits in this PR. Something about how I merged my branch with upstream. The actual diff is not that huge.


tests/test_end_to_end.py
```diff
@@ -0,0 +1,44 @@
...
+        # TODO determine this dynamically
+        combine_dims = [DimKey("time", operation=CombineOp.CONCAT)]
+        schemas = datasets | DatasetToSchema()
+        schema = schemas | DetermineSchema(combine_dims=combine_dims)
+        indexed_datasets = datasets | IndexItems(schema=schema)
+        target = schema | PrepareZarrTarget(target_url=tmp_target_url)
+        _ = indexed_datasets | StoreToZarr(target_store=target)
+
+    ds = xr.open_dataset(tmp_target_url, engine="zarr").load()
+    xr.testing.assert_equal(ds, daily_xarray_dataset)
```
rabernat, line 44: This block is really the highlight of this PR. It is the closest we have come yet to recreating the original monolithic XarrayZarrRecipe in modular Beam style.

But there is still a lot of room to refine this. Perhaps some of the intermediate steps can be merged into a single PTransform. I'm not sure the user needs to be passing schemas around.

Really interesting in getting feedback on the API and flow here - @cisaacstern and @alxmrs.

tests/test_end_to_end.py
```diff
@@ -0,0 +1,44 @@
...
+    with pipeline as p:
+        inputs = p | beam.Create(pattern.items())
+        datasets = inputs | OpenWithXarray(file_type=pattern.file_type)
+        # TODO determine this dynamically
+        combine_dims = [DimKey("time", operation=CombineOp.CONCAT)]
+        schemas = datasets | DatasetToSchema()
+        schema = schemas | DetermineSchema(combine_dims=combine_dims)
+        indexed_datasets = datasets | IndexItems(schema=schema)
+        target = schema | PrepareZarrTarget(target_url=tmp_target_url)
+        _ = indexed_datasets | StoreToZarr(target_store=target)
```
alxmrs, line 41: Here's a translation (really, for myself) to more idiomatic Beam:

```suggestion
       combine_dims = [DimKey("time", operation=CombineOp.CONCAT)]
       datasets = (
           p
           | beam.Create(pattern.items())
           | OpenWithXarray(file_type=pattern.file_type)
        )
        schema = (
            datasets 
            | DatasetToSchema() 
            | DetermineSchema(combine_dims=combine_dims)
        )
        target = schema | PrepareZarrTarget(target_url=tmp_target_url)
        _ = (
            datasets 
            | IndexItems(schema=schema)
            | StoreToZarr(target_store=target)
         )
```

My high level feedback is that this relies seems a bit too imperative. This is my bias, but I could see the pipeline fitting closer to the [XArray Beam abstractions](https://github.com/google/xarray-beam/blob/4dced4ead5a9511ab18027780f3a98f964584067/examples/era5_climatology.py#L70), something like:

```
(
    p
    | beam.PatternToChunks(pattern)
    # other processing...
    | beam.ChunksToZarr()
```

Though, I could see how each of these steps would be composed out of the components you have here. Exposing the right level of granularity is hard.

alxmrs: From your comment below: 
> Perhaps some of the intermediate steps can be merged into a single PTransform? I'm not sure the user needs to be passing schemas around.

I totally agree with this sentiment!

tests/test_end_to_end.py
alxmrs: I'll review this PR in more depth later today or tomorrow. It's exciting to see all this progress!

vietnguyengit:
Hi @rabernat , maybe my question is not too related to this PR, but keywords `zarr` and `pangeo-forge` bring me here, I wonder when I build a recipe to convert multiple NetCDF files (around 100k+) to Zarr to S3 using `pangeo-forge-recipe` on the private environment (that I won't use `pangeo-forge` cloud, the recipe also build to use with Prefect 2.0, will `pangeo-forge-recipe` and `DaskTaskRunner` of Prefect perform the tasks (e.g. `xarray.dataset.to_zarr()`) in parallel wisely? Currently, `DaskTaskRunner` of Prefect won't be able to pick up the tasks of `xarray` functions automatically, tasks need to be wrapped in a function with `@task` decorator to be added into Dask graph, eg.

```python
@task
def xr_to_zarr(dataset):
  dataset.to_zarr(...)
```
but doing this way, the whole block `xr_to_zarr` will be performed on a single thread. When without Prefect, calling `dataset.to_zarr(...)` , Dask will be able to fill up the threads with tasks, something looks like this:

image

Much appreciated your time reading this. Please advice. Thank you.

rabernat:
Hi @vietnguyengit! Thanks for your interest in Pangeo Forge. We would love to help you, but please open a [new issue](https://github.com/pangeo-forge/pangeo-forge-recipes/issues) with your question from above. As you said, your question is not related to this pull request.

martindurant:
I am looking into it. However, the following snippet is informative:
```
import pickle, io
b = io.BytesIO(b"data")
s = io.TextIOWrapper(b)
pickle.dumps(s)  # TypeError
```
i.e., even simple file-like types are often not pickleable. I have been playing with weakrefs to augment OpenFile.open's logic, but I cannot get around something fundamental like the above, without writing an explicitly pickleable wrapper class such as LocalFileOpener currently is. This way we could, with a little effort, more rigorously guarantee that the output of OpenFile.open() (or inside `with`) can be pickled.

Conversely, we could backtrack and declare that file-like objects should not, in general, be pickled and we don't guarantee that they can; this means only pickling *outside* contexts. Indeed, it is somewhat rare for context manager yielded objects to be pickleable, since they explicitly contain state and at least some closure.

rabernat:
I understand that pickling these objects is complicated (remember https://github.com/fsspec/filesystem_spec/issues/579#issuecomment-1140377495 🙃 ). The problem here is API stability. I spent literally many full days trying to figure out a solution for serializing fsspec open file objects inside Xarray datasets that worked with beam. I [found one](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/beam-refactor/pangeo_forge_recipes/openers.py). It's coded up and tested in Pangeo Forge. The fsspec 2022.7.1 release made some change that broke whatever fragile solution I had found, which triggered the tests to fail. (Unit testing ftw!) So there are two options:
- The solution I found was actually unsupported behavior. It was an accident that it worked. I have to go back to the drawing board.
- There was a regression in fsspec in the 2022.7.1 release which needs to be fixed in fsspec.

Do you have a sense of which of the two options above is more appropriate here?

martindurant:
I'll have a look to see what changed, and whether it can easily be overcome. I'd say that indeed, this was not explicitly supported. I'll also look into my suggestion, above, of how it could be supported and see whether that's tractable or not.

rabernat:
Upstream dev passes thanks to https://github.com/fsspec/filesystem_spec/pull/1024. Gonna move on.

cisaacstern:
🚀

martindurant:
Sorry, I'm going to have to revert that PR. It breaks too many assumptions in other places :|

In short: an OpenFile should either 
- be used in a context
- have `.close()` called *on the OpenFile* to ensure all contained file-like objects are cleaned up. 

Since all AbstractBufferedFile descendants (s3fs, gcsfs, abfs) are pickleable, and now so are LocalFileOpeners, you should be OK to pickle inside the context, if you want, either the OpenFile or the file-like it generates.

Issue #392, opened 2022-07-27
Raster data support


norlandrhagen:
Hi there,

Wondering if `.TIFF` files were supported in the `XarrayZarrRecipe`. 

In this blurb from the docs, it seems to be possible: 
```The inputs can be in any [file format Xarray can read](http://xarray.pydata.org/en/latest/user-guide/io.html), including NetCDF, OPeNDAP, GRIB, Zarr, and, via [rasterio](https://rasterio.readthedocs.io/), GeoTIFF and other geospatial raster formats. T```

When trying to make use of rioxarray for opening up the file, I set `{'engine': 'rasterio'}`, but got this error.

```ValueError: pangeo-forge-recipes will automatically set the xarray backend for files of type 'netcdf4' to '{'engine': 'h5netcdf'}', which is different from the value you have passed via `xarray_open_kwargs`. If this input file is actually of type 'netcdf4', please remove `{'engine': 'rasterio'}` from `xarray_open_kwargs`. If this input file is not of type 'netcdf4', please update this recipe by passing a different value to `FilePattern.file_type`.```

I also tried to specify the `file_type` in the [`FilePattern`](https://pangeo-forge.readthedocs.io/en/latest/pangeo_forge_recipes/api_reference.html?highlight=api#file-patterns), but couldn't find a accepted file format. 

It seems like `tiff/geotiff` isn't an option in [`patterns.py#L111`](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/d1875c3f483188ed72e34177c58bc896b947aa76/pangeo_forge_recipes/patterns.py#L111)

Anyhow, just wondering if there are any example recipes using rioxarray or if anyone has any advice on how to proceed.

cc @cisaacstern 

Thanks!

cisaacstern:
This recipe uses raster data: https://github.com/pangeo-forge/staged-recipes/pull/133

But it was written before the `file_type` kwarg was added to FilePattern.

To make this work now, we'd need a PR adding `tiff` to the options for FilePattern. Feel free to have a go at that if you'd like!

While that's in progress, downgrading to `pangeo-forge-recipes==0.8.2` will allow you to run the recipe with `{'engine': 'rasterio'}` without hitting the ValueError.

norlandrhagen:
Sounds good! I'll open up a PR adding `tiff`.

cisaacstern:
Awesome, feel free to ping me with any questions about that!

[...]

PR #393, opened 2022-08-01
Feature/add tiff file opt

norlandrhagen:


rabernat:
Thanks for working on this Raphael! Let us know how we can help.

norlandrhagen:
Thanks @rabernat! Yeah, this tiny update worked on a recipe of tiff files. Any other tests to add or additions to the docs you would suggest?

cisaacstern:
@norlandrhagen, thanks for working on this! In order for this `FileType` to translate to the selection of the correct xarray backend at runtime, you'll also need to add an entry to

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/d1875c3f483188ed72e34177c58bc896b947aa76/pangeo_forge_recipes/recipes/xarray_zarr.py#L39-L42

probably 

```python
FileType.tiff: dict(engine="rasterio")
```

?

In terms of testing, all members of `FileType` are automatically tested for, basically, existence here:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/d1875c3f483188ed72e34177c58bc896b947aa76/tests/test_patterns.py#L188-L189

A more ambitious testing path would be to generate some synthetic GeoTIFF data, as we do in the test suite for NetCDF data, and then try running an XarrayZarrRecipe with it. IMO that's rather ambitious and perhaps beyond the scope of what's necessary for this PR.

norlandrhagen:
Thanks for the feedback! I incorporated your edits in the `OPENER_MAP`

cisaacstern:
I think the test failures are because Prefect 2.0 was released since the last time we ran these tests.

To get them to pass, we'll need to pin Prefect to the last pre-2.0 release in the ci environments:

https://github.com/pangeo-forge/pangeo-forge-recipes/tree/master/ci

norlandrhagen:
Sounds good. I can open up a separate PR for that. It looks like they slightly changed the way the `DaskExecutor` is imported, so might have to make a few code tweaks.

cisaacstern:
We will soon be dropping Prefect entirely, so I would suggest going back to the latest version of Prefect which does not require any tweaks to `pangeo-forge-recipes` code, however old that is. Thanks Raphael!

norlandrhagen:
Ah icic, I misread your comment!

cisaacstern:
With #398 merged, merging from upstream should resolve test failures here.

cisaacstern:
@norlandrhagen, do you have pre-commit installed locally, and if so is your local pre-commit passing?

> If you don't have it installed locally, from your repo root, you should be able to just `pip install pre-commit` and then call `pre-commit install` followed by `pre-commit run --all-files`.

I'm not sure what this pre-commit error we're seeing is, and just wondering if it's GitHub Actions-specific, or not.

norlandrhagen:
Hmm, pre-commit seems to be passing?

image

cisaacstern:
Hmm yeah that's odd. I'll pull this branch and see if I can fix this.

norlandrhagen:
Thanks!

cisaacstern:
@norlandrhagen, I think what happened is that, when pre-commit runs automatically at commit time, it only checks files that were changed in that commit (this might be configurable). I made the fix by manually calling `pre-commit run --all-files`, which ensures that all files are re-formatted, and this seems to have resolved the issue.

If we could just do one last check on this PR, I think we'll be glad we did. Specifically, this PR aims to close https://github.com/pangeo-forge/pangeo-forge-recipes/issues/392, in which you mentioned getting some errors when trying to run a recipe with TIFF inputs. Could I ask that you just manually verify that installing `pangeo-forge-recipes` from

```
pip install "git+http://github.com/norlandrhagen/pangeo-forge-recipes.git@feature/add_tiff_file_opt"
```

resolves the issue you reported in the #392? Writing true tests for this I think is beyond the scope of what we'll want to get into now, but if you could just manually check that this branch solves your issue, it could give us a bit more confidence that we won't have to circle back with another PR to fix #392.

norlandrhagen:
@cisaacstern just installed this branch and tested a simple recipe that opened tiffs. I think this PR is ready to merge if that works for you.

cisaacstern:
Awesome, sounds good to me. Thanks for this contribution, @norlandrhagen! And for your patience with the issues that came up along the way, such as #398. I'll merge now.

norlandrhagen:
Glad to to help out a bit!

PR #394, opened 2022-08-02
Small fix in prepare_target: catch `PathNotFoundError`.

alxmrs:
Prepare target checks if a Zarr exists by attempting to open a Zarr and then creating one when an error occurs. I've found that in addition to the `zarr.errors.GroupNotFoundError`, the `zarr.errors.PathNotFoundError` needs to be caught in order to not break the xarray_zarr.py pipeline.

### DIFF
--- a/pangeo_forge_recipes/recipes/xarray_zarr.py
+++ b/pangeo_forge_recipes/recipes/xarray_zarr.py
@@ -507,7 +507,7 @@ def prepare_target(*, config: XarrayZarrRecipe) -> None:
             # TODO: check that target_chunks id compatibile with the
             # existing chunks
             pass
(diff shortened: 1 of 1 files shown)


rabernat:
Looks like a prefect release has broken all of our test suite.

```
ImportError while loading conftest '/home/runner/work/pangeo-forge-recipes/pangeo-forge-recipes/tests/conftest.py'.
tests/conftest.py:29: in <module>
    from prefect.executors import DaskExecutor
E   ModuleNotFoundError: No module named 'prefect.executors'
```

I think the solution is probably to pin an older version of prefect in our CI. Since we are planning on removing prefect support anyway, I don't see any reason not to do this as a short term fix to get the tests passing.

rabernat:
Prefect test issue  is being fixed in #398. In the meantime, LGTM.

PR #398, opened 2022-08-12
pinned prefect version to 1.3.0 to avoid 2.0 major changes

norlandrhagen:
@cisaacstern Hopefully this version addresses https://github.com/pangeo-forge/pangeo-forge-recipes/pull/393

### DIFF
--- a/ci/py3.8.yml
+++ b/ci/py3.8.yml
@@ -26,7 +26,7 @@ dependencies:
   - numpy
   - pandas
   - pip
-  - prefect
+  - prefect>=1.3.0
   - pydap
   # bring back eventually once pynio conda-forge package does not conflict
   # with ujson, which is a depencency of kerchunk's conda-forge feedstock.
--- a/ci/py3.8.yml
+++ b/ci/py3.8.yml
@@ -26,7 +26,7 @@ dependencies:
   - numpy
   - pandas
   - pip
-  - prefect>=1.3.0
+  - prefect==1.3.0
   - pydap
   # bring back eventually once pynio conda-forge package does not conflict
   # with ujson, which is a depencency of kerchunk's conda-forge feedstock.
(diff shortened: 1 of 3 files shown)


ci/py3.8.yml
```diff
@@ -26,7 +26,7 @@ dependencies:
   - numpy
   - pandas
   - pip
-  - prefect
+  - prefect>=1.3.0
```
cisaacstern, line 29: I think `>=1.3.0` will bring in the latest Prefect, so we probably just want `==`?

ci/py3.9.yml
```diff
@@ -28,7 +28,7 @@ dependencies:
   - numpy
   - pandas
   - pip
-  - prefect
+  - prefect>=1.3.0
```
cisaacstern, line 31: Same question as above.

cisaacstern (commented):
Thanks Raphael! I think we want `==`, not `>=`?

norlandrhagen:
:sweat_smile: Oof, I think it's too late for me to make edits haha. Good catch.

cisaacstern:
This is great progress. Now it's just the upstream dev tests that are failing (for the same reason).

Deleting this line should resolve that

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/d1875c3f483188ed72e34177c58bc896b947aa76/ci/upstream-dev.yml#L8

We no longer need to be testing against upstream dev versions of Prefect, since we plan to drop it as a dependency soon.

cisaacstern:
I _think_ what's going on here is that the change to `upstream-dev.yml` for some reason did not reset the cached envs.

I just manually deleted (I think all of) the cached upstream-dev envs as described here https://stackoverflow.com/a/64819132, so I'll re-run the checks now.

cisaacstern:
Awesome, this passes, so will merge now. Thanks, @norlandrhagen!

PR #405, opened 2022-08-29
Refactor index types

rabernat:
This is an internal refactor of the core `Index` datatype. Like xarray-beam `Key` we use something that looks like a dictionary to keep track of where a fragment lies within a dataset.

This PR does a few things:
- Moves `Index` to a standalone `types` module
- Refactors `Index` to use a key value pair of `Dimension` and `Position`
- Tries to eliminate redundancy in `Position`, moving closer to xarray-beam's "offests" concept.

We may be able to eventually just use the same type of index as xarray-beam. In the meantime, this simplifies things and moves us in that direction.

My main complaint is that this feels a very java-like.

cc @shoyer if you want to take a look.


pangeo_forge_recipes/patterns.py
```diff
@@ -239,31 +172,34 @@ def concat_sequence_lens(self) -> Dict[str, Optional[int]]:
...
-                dim
-                for dim in self.combine_dims
-                if dim.name == dimkey.name and dim.operation == dimkey.operation
+                combine_dim
+                for combine_dim in self.combine_dims
+                if combine_dim.dimension == dimension
             ]
             if len(dims) != 1:
-                raise KeyError(r"Could not valid combine_dim for indexer {idx_key}")
+                raise KeyError(f"Could not valid combine_dim for dimension {dimension}")
```
alxmrs, line 189: My personal preference is `{dimension!r}`.

pangeo_forge_recipes/writers.py
```diff
@@ -10,12 +10,14 @@
...
-        concat_dim_val = index.find_concat_dim(dim)
-        if concat_dim_val:
+        concat_dimension = index.find_concat_dim(dim)
+        if concat_dimension:
             # we are concatenating over this dimension
-            assert concat_dim_val.start is not None
-            assert concat_dim_val.stop == concat_dim_val.start + dimsize
-            region_slice.append(slice(concat_dim_val.start, concat_dim_val.stop))
+            position = index[concat_dimension]
+            assert position.indexed
```
alxmrs, line 17: Would be useful to have an assert message, maybe with some message to help end-users debug.

alxmrs (approved):
Two small nits. Looks great!

Issue #222, opened 2021-10-15
Transfer inputs using Globus


rabernat:
### Background

Pangeo Forge recipes operate on "inputs", string URLs that point to individual source files. These inputs are generated by the recipe [FilePattern](https://pangeo-forge.readthedocs.io/en/latest/recipe_user_guide/file_patterns.html) object.

Currently, Pangeo Forge relies 100% on [fsspec](https://filesystem-spec.readthedocs.io/en/latest/) to download these files. This is great because fsspec has a ton of [implementations](https://filesystem-spec.readthedocs.io/en/latest/api.html#built-in-implementations) for just about every file transfer / storage protocol in the world.

However, fsspec does not support [Globus](https://www.globus.org/)...

> Globus provides a secure, unified interface to your research data. Use Globus to 'fire and forget' high-performance data transfers between systems within and across organizations.

...and it probably never will. Globus is a system for transferring whole files and nothing else, so wrapping it in a filesystem-like abstraction probably just doesn't make sense. (@martindurant correct me if I'm wrong.) However, so much scientific data is accessible over Globus. Our impact will be much greater if we can interoperate with Globus somehow.

However, using Globus will likely require us to refactor our approach to file transfers; Globus operates at a higher level, so much of the logic we have for opening, streaming, and saving files will be unnecessary (also discussed in #95).

### Questions about Globus

To move forward, we need to answer some basic questions about Globus. I'm going to tag @rpwagner, who has been very helpful in this area in the past and see if he bites.

- Is there such thing as a "global namespace" for the entire Globus federation? Can I refer to a specific file at a specific endpoint via a string url, e.g `globus://my-globus-endpoint.columbia.edu/my/specific_file.nc`?
- If we wanted to use Globus to transfer data from user-specified HPC systems to our cloud buckets via automated batch jobs, how would we handle authentication? Does Globus support "service accounts," long-lived tokens, etc,...any way of storing credentials persistently for months / years without requiring interactive login? (Obviously such credentials would have to be managed securely, but that's the case for any credentials.)
- Would we need a paid Globus subscription to transfer data to buckets in S3 / GCS?
- Does Globus work with OSN?
- What does the python code look like to create and wait on a file transfer using Globus? 
- Is it possible to rename the file as it is transferred to the target endpoint?

[...]

PR #408, opened 2022-09-01
Document how to access data with Globus

rabernat:
This PR documents the workaround developed in https://github.com/pangeo-forge/pangeo-forge-recipes/issues/222#issuecomment-1080830165 to access data via Globus over HTTPS.

I don't consider this an actual solution to #222 because we cannot leverage the full Globus transfer API.


yuvipanda:
This seems great, thanks @rabernat

rabernat:
lol I just forgot about this PR. Thanks Yuvi!

PR #425, opened 2022-10-21
Rechunk before writing

rabernat:
This PR attempts to automatically rechunk a pcollection when storing it to Zarr in order to align chunks properly.


rabernat:
@alxmrs I could use some Beam advice here.

Consider this Rechunk Ptransform

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/345c17275eee7edb8c158d94c0f578d3a97b1764/pangeo_forge_recipes/transforms.py#L246-L262

Here's the problem: rechunking is very expensive. As currently coded up, rechunking will always happen. However, in many cases, it's an unnecessary step--specifically, when the pcollection fragments are already aligned with the target chunks.

My question is this: what's the proper way to have conditional logic in my Beam PTransform. Basically I want to examine `schema` and `target_chunks` and determine if rechunking is actually necessary. If not, I just want to pass the PCollection through unchanged. The problem is that, while `target_chunks` is known at pipeline construction time, `schema` is not; it has to be determined by examining the inputs.

I'm a bit stuck on the best way to implement this. Any advice you have would be much appreciated. 🙏

alxmrs:
Hey Ryan! Let's discuss the issue here a bit more in depth. My top-of-mind answer – I think you could borrow this pattern from the `beam.Create()` transform: 

https://github.com/apache/beam/blob/2d151c306ee0ccb2397c350f54d6333af88651d5/sdks/python/apache_beam/transforms/core.py#L3470

Create decides to to a reshuffle (an expensive operation) only if the tuple length is past a certain threshold. You may not even need to do this in a composite PTransform -- I think an `if` in the `expand` could work. 

Though – is the crux of the issue that you need to change behavior depending on the contents of the PCollections? How should you decide when to rechunk?

> while target_chunks is known at pipeline construction time, schema is not; it has to be determined by examining the inputs.

Ah, that does seem to be the case. For here, I recommend taking a leaf out of XArray-Beam's playbook, and using the `beam.AsSingleton()` operator: 

https://github.com/google/xarray-beam/blob/310c865741ebd9ca2f027fe78f6000756f7e4696/xarray_beam/_src/zarr.py#L309

In general, I recommend reading the whole implementation of the transform linked here. To me, it seems like both approaches try to address the same problem. However, your schema calculating ability may have an advantage over XArray-Beam, because more information can be made available via the `FilePattern` interface.

@shoyer probably has more context on the tradeoffs of this approach, given he has a parallel solution to this problem.

rabernat:
This PR has never been merged, but it really needs to go into beam refactor. Otherwise things may not work properly.

I'll try to put some time into this soon!

Issue #427, opened 2022-10-25
Generating a recipe's hash fails when using `process_chunk` and `process_input` functions


andersy005:
Recently, I've encountered a beam issue that has not been easy to reproduce. This seems to affect recipes that use the `process_input` and `process_chunk` functions.

- https://github.com/pangeo-forge/staged-recipes/pull/145
- https://github.com/pangeo-forge/AGDC-feedstock/pull/3
- https://github.com/pangeo-forge/LMRv2p1_MCruns_ensemble_gridded-feedstock/pull/2
- https://github.com/pangeo-forge/staged-recipes/pull/133

Everything works perfectly fine when using apache beam locally. However, for reasons I don't understand,  pangeo-forge-recipes is unable to hash these two inputs when using dataflow, resulting in the error below. 

 **There are plenty of recipes/feedstocks that are blocked by this issue, and any help diagnosing the issue would be greatly appreciated**.  Cc @alxmrs, @rabernat, @derekocallaghan, @yuvipanda 

Apologies for the lack of a minimal reproducible example. I haven't been able to come up with one :( 

```python
  File "/srv/conda/envs/notebook/lib/python3.9/inspect.py", line 827, in findsource
    raise OSError('source code not available')
OSError: source code not available [while running 'Start|cache_input|Reshuffle_000|prepare_target|Reshuffle_001|store_chunk|Reshuffle_002|finalize_target|Reshuffle_003/prepare_target-ptransform-56']
,

  a6170692e70616e67656f2d66-10201802-cyzc-harness-sbr9
      Root cause: Traceback (most recent call last):
  File "apache_beam/runners/common.py", line 1417, in apache_beam.runners.common.DoFnRunner.process
  File "apache_beam/runners/common.py", line 837, in apache_beam.runners.common.PerWindowInvoker.invoke_process
  File "apache_beam/runners/common.py", line 983, in apache_beam.runners.common.PerWindowInvoker._invoke_process_per_window
  File "/usr/local/lib/python3.9/dist-packages/apache_beam/transforms/core.py", line 1877, in <lambda>
  File "/srv/conda/envs/notebook/lib/python3.9/site-packages/pangeo_forge_recipes/executors/beam.py", line 14, in _no_arg_stage
    fun(config=config)
  File "/srv/conda/envs/notebook/lib/python3.9/site-packages/pangeo_forge_recipes/recipes/xarray_zarr.py", line 587, in prepare_target
[...]

PR #428, opened 2022-10-26
Use bytecode for hashes in *some* cases

yuvipanda:
This is guaranteed to be *not* stable across python versions (and probably other implementation details), but without this stuff just crashes

Fixes https://github.com/pangeo-forge/pangeo-forge-recipes/issues/427


yuvipanda:
Superseeded by #429

PR #390, opened 2022-07-25
(WIP/DO NOT SUBMIT) Re-factor `HDFReferenceRecipe` to expose kerchunk types

darothen:
Working branch for discussion in #387, refer to there for now.

Issue #387, opened 2022-07-19
add GRIB2ReferenceRecipe 


rsignell-usgs:
We have `HDFReferenceRecipe`, but kerchunk also handles GRIB2.    Perhaps `GRIB2ReferenceRecipe`?

darothen:
This seems pretty useful; I've been playing around with this during the evening, and it looks like fsspec/kerchunk#198 would unblock writing recipes that read from remotely hosted GRIB files. Directly translating the example HRRR concatenation into a recipe here would be a really cool addition. Happy to play around with this and propose a refactor for `HDFReferenceRecipe` once that upstream PR in kerchunk lands.

darothen:
A quick update - I was able to hack out a minimal working example of using the existing `HDFReferenceRecipe` class to read GRIB2 files. The demo code is in #390, but note that it would help if you build a very recent version of kerchunk - at least after the significant GRIB utility refactor in fsspec/kerchunk#198, and for the demo code below, fsspec/kerchunk#204.

In a nut-shell, it was pretty straightforward to modify the existing codepaths for scanning GRIB files. A major departure from the HDF formulation is that each GRIB message gets its own reference, so we need a more elegant way to combine this so that a single GRIB file maps to a single Zarr file. I opted just to flatten all the messages, assuming metadata about coords remains constant, but there are other places this could be fixed, namely in the `MultiZarrToZarr` utility. I think a simple dataclass which helps mediate between the reference spec produced by `fsspec` would help here, since there will probably be a variety of reference lists that need to be handled.

The other big change would be the actual interface for `HDFReferenceRecipe` - or whatever it would become. Given the small inconsistencies between the NetCDF4 reader and the GRIB reader here, a really good case could be made for having separate reference classes, but I'll explore how much overlap there is and see if there is a sensible re-factoring that would allow easy extension for each specialized filetype.

For a quick demo, here's a [simple gist](https://gist.github.com/darothen/5380e223ae5bc894006a5b6ed5a27cbb) which subsets from a HRRR forecast and concatenates the output. This workflow is **vastly superior** to most ones that I've seen for extracting useful data from the online HRRR archive - it's fast, easy to modify, and is very concise thanks to the great mechanics inside pangeo-forge.

darothen:
[...]

PR #438, opened 2022-11-04
Grib Reference Recipe

norlandrhagen:
PR building on https://github.com/pangeo-forge/pangeo-forge-recipes/pull/390.  Credit to [darothen](https://github.com/darothen).
Based on discussion from https://github.com/pangeo-forge/pangeo-forge-recipes/issues/387. 

Minor update from #390, fixes `s3` reading issue.

### DIFF
--- a/pangeo_forge_recipes/recipes/__init__.py
+++ b/pangeo_forge_recipes/recipes/__init__.py
@@ -1,8 +1,8 @@
 from .base import BaseRecipe
-from .reference_hdf_zarr import HDFReferenceRecipe
+from .reference_hdf_zarr import GribReferenceRecipe, HDFReferenceRecipe
 from .xarray_zarr import XarrayZarrRecipe
 
-__all__ = ["BaseRecipe", "XarrayZarrRecipe", "HDFReferenceRecipe"]
+__all__ = ["BaseRecipe", "XarrayZarrRecipe", "HDFReferenceRecipe", "GribReferenceRecipe"]
 
 
 def setup_logging(level: str = "INFO"):
--- a/pangeo_forge_recipes/recipes/reference_hdf_zarr.py
+++ b/pangeo_forge_recipes/recipes/reference_hdf_zarr.py
@@ -39,21 +39,19 @@ def scan_file(chunk_key: ChunkKey, config: HDFReferenceRecipe):
 
 def scan_grib(chunk_key: ChunkKey, config: GribReferenceRecipe):
     assert config.storage_config.metadata is not None, "metadata_cache is required"
+
     fname = config.file_pattern[chunk_key]
     ref_fname = os.path.basename(fname + ".json")
 
     with file_opener(fname, **config.storage_options) as fp:
+
         protocol = getattr(getattr(fp, "fs", None), "protocol", None)  # make mypy happy
         if protocol is None:
             raise ValueError("Couldn't determine protocol")
         target_url = unstrip_protocol(fname, protocol)
-        reference = create_grib2_reference(
-            fp,
-            fname,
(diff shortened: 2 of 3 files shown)


norlandrhagen:
TODO:

- [ ] Add `GribReferenceRecipe` testing.
- [ ] Add documentation for `GribReferenceRecipe` example.

ci/py3.8.yml
```diff
@@ -7,11 +7,12 @@ dependencies:
   - apache-beam
   - black
   - boto3
-  - cfgrib<0.9.9.0
+  - cfgrib>=0.9.10.2 # Needed for grib2 reference recipe to work
```
andersy005, line 10: Does this mean that we need to update the requirements in `setup.cfg` to include `cfgrib` and `eccodes`?

norlandrhagen: @andersy005, just updated.

andersy005:
@norlandrhagen, it's my understanding that the tests are failing due to this line 

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/fe5d1f7241e724fde76f86ce582fb0b830af3036/tests/recipe_tests/test_XarrayZarrRecipe.py#L172

setting ` open_input_with_kerchunk=False` seems to be one solution but i don't quite understand why this stopped working in this PR. Perhaps that @cisaacstern has some insights since he is the last person to have touched this part of the codebase :)

cisaacstern:
> setting `open_input_with_kerchunk=False` seems to be one solution

IIUC based on the docstring

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/fe5d1f7241e724fde76f86ce582fb0b830af3036/tests/recipe_tests/test_XarrayZarrRecipe.py#L169

This is explicitly testing `open_input_with_kerchunk=True` so setting that to `False` is not the solution here.

> but i don't quite understand why this stopped working in this PR

I don't have an immediate answer but questions that come to mind:

1. What was the last PR where this test passed?
2. Has the version of kerchunk used in CI changed since then?
3. Has anything that touches `open_input_with_kerchunk` changed since then?

andersy005:
It's likely this PR isn't the culprit because 

- https://github.com/pangeo-forge/pangeo-forge-recipes/pull/440 is  failing also


> Has anything that touches open_input_with_kerchunk changed since then?

I doubt it. It appears the [latest merge into main](https://github.com/pangeo-forge/pangeo-forge-recipes/commit/fe5d1f7241e724fde76f86ce582fb0b830af3036) succeeded: https://github.com/pangeo-forge/pangeo-forge-recipes/actions/runs/3415831179/jobs/5685406660

cisaacstern:
Per @norlandrhagen and my discussion this afternoon:

- `beam-refactor` remains on the near horizon, so any work that goes into this should be **(a)** as narrowly-scoped as possible; and **(b)** done with the awareness that it will need to be re-done before too long.
- The basics of what's required to make this work are simple enough that I believe it's worth continuing to pursue now, given that Raphael is interested (and it has relevance to his work), and he/we are open to using it as a learning experience to entrain him further into `pangeo-forge-recipes` development, and inform what will be required for a similar feature for the `beam-refactor`.
- Re: design, a mergable version of this should follow @rabernat's suggestion in https://github.com/pangeo-forge/pangeo-forge-recipes/issues/387#issuecomment-1189449925, to use a single `ReferenceRecipe` object, which can infer file type from the user-passed `file_pattern.file_type` (as does `XarrayZarrRecipe`).
- Re: testing, Raphael is going to run the tests against `master` locally, and assuming they work, may simply start a new PR from there.

Thanks Raphael for your engagement on this!

PR #459, opened 2022-12-28
Move FilePattern functions into methods

yuvipanda:
There were 3 useful functions that were free-floating, and turn that into methods of
the `FilePattern` class instead. These methods always operated on `FilePattern` objects,
and took a single param (akin to `self`) plus other args.

It also makes some possible manipulations (ast or otherwise) easier.

This PR also renames some of the `blockchain` methods to refer instead to the
[merkle tree](https://en.wikipedia.org/wiki/Merkle_tree), which is the underlying
data structure used by plenty of things that try to do this kinda structured diff
detection (such as git). I think overall when searching for useful properties of
a datasctructure, 'blockchain' is probably not really useful anymore while
merkle trees are far more useful.

### DIFF
--- a/docs/pangeo_forge_recipes/development/release_notes.md
+++ b/docs/pangeo_forge_recipes/development/release_notes.md
@@ -10,7 +10,7 @@ Collectively, this provides for generation of deterministic hashes for both reci
 pattern instances. Checking these hashes against those from a prior version of the recipe can be
 used to determine whether or not a particular recipe instance in a Python module (which may
 contain arbitrary numbers of recipe instances) has changed since the last time the instances in
-that module were executed. The file pattern hashes are based on a blockchain built cumulatively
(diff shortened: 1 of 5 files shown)


yuvipanda:
Given that we are breaking compatibility with this release, I think it is useful to consolidate everything! We should either have functions that take an object as first param everywhere, or use methods everywhere.

rabernat (approved):
Thanks a lot for this Yuvi! I agree it makes sense to move these methods to the FilePattern class.

I do not have strong opinions on the naming of anything and am happy to use Merkle Tree instead of blockchain.

Please fix pre-commit, then LGTM.

cisaacstern (approved):
Thanks for the thoughtful work, Yuvi!

Aside from pre-commit LGTM!

yuvipanda:
Thanks @rabernat and @cisaacstern!

PR #460, opened 2022-12-28
Allow target to be str or FSSpecTarget

yuvipanda:
The latter allows them to be complex FSSpec objects that can encode information about S3 / GCS, and how they can be authenticated with.

### DIFF
--- a/pangeo_forge_recipes/transforms.py
+++ b/pangeo_forge_recipes/transforms.py
@@ -79,17 +79,20 @@ def _add_keys(func):
 class OpenURLWithFSSpec(beam.PTransform):
     """Open indexed string-based URLs with fsspec.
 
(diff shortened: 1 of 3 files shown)


rabernat:
Thanks for this! I understand the motivation for this change. Before moving forward, I think it's worth discussing the design here a little bit.

The reason that I switched to allowing only str targets is because I figured that it would make it much easier to supply the target URL via a beam runtime parameter. So my first question is, _do you think an fsspec object can be passed as a runtime parameter?_

As an alternative to supporting fsspec objects, have we considered just using the default boto / gcs methods for specifying credentials (environment variables, config files, etc.)? I believe this should work well for everything but `endpoint_url`.

yuvipanda:
I think after this @rabernat saw pangeo-forge slack messages that had a lot more context! Curious what you think now, @rabernat?

PR #471, opened 2023-01-12
Original variable encodings are retained

### LINKED ISSUE #465: Original `time` encoding units not used in Zarr generated by `beam-refactor` branch
While porting the EOOffshore CCMP recipe to use the `beam-refactor` branch, I noticed that the [output Zarr contains non-unique `time` values](https://github.com/pangeo-forge/pangeo-forge-recipes/issues/453#issuecomment-1373744070).

It looks like the original dataset `time` `encoding` `units` aren't retained in `pangeo_forge_recipes.writers._store_data()`:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/eea8e4452a15dc5aeed86df9eefe2130faaf1036/pangeo_forge_recipes/writers.py#L30

At this point the original `time` encoding is as expected (**i.e. `hours...`**): 

```python
{'chunksizes': (524288,), 'fletcher32': False, 'shuffle': False, 'source': '<fsspec.implementations.local.LocalFileOpener object at 0x7f88c259aac0>', 'original_shape': (4,), 'dtype': dtype('float64'), 'units': 'hours since 1987-01-01 00:00:00'}
```

The `units` are subsequently replaced after this statement, where the use of **`days...`** is resulting in the non-unique values:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/eea8e4452a15dc5aeed86df9eefe2130faaf1036/pangeo_forge_recipes/writers.py#L31

```python
{'chunksizes': (524288,), 'fletcher32': False, 'shuffle': False, 'source': '<fsspec.implementations.local.LocalFileOpener object at 0x7f88c259aac0>', 'original_shape': (4,), 'dtype': dtype('float64'), 'units': 'days since 1970-01-01 00:00:00', '_ARRAY_DIMENSIONS': ['time'], '_CoordinateAxisType': 'Time', '_Fillvalue': -9999.0, 'axis': 'T', 'calendar': 'proleptic_gregorian', 'delta_t': '0000-00-00 06:00:00', 'long_name': 'Time of analysis', 'standard_name': 'time'}
```

I don't have a fix for this yet, but it looks like the issue is in `pangeo_forge_recipes.aggregation._to_variable()` based on the existing comments:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/eea8e4452a15dc5aeed86df9eefe2130faaf1036/pangeo_forge_recipes/aggregation.py#L198-L203

derekocallaghan:
* Fixes #465
* Original variable encoding `source` and `_FillValue` are excluded
* No test failures with these changes, although some warnings occur. I suspect the latter may be related to my local environment:

```bash
$ pytest tests
======================================================================================================================== test session starts =========================================================================================================================
platform linux -- Python 3.9.13, pytest-7.2.0, pluggy-1.0.0
rootdir: .../pangeo-forge-recipes, configfile: setup.cfg
plugins: anyio-3.6.2, lazy-fixture-0.6.3
collected 318 items / 1 skipped                                                                                                                                                                                                                                      

tests/test_aggregation.py .....                                                                                                                                                                                                                                [  1%]
tests/test_chunk_grid.py ........                                                                                                                                                                                                                              [  4%]
tests/test_combiners.py ..                                                                                                                                                                                                                                     [  4%]
tests/test_end_to_end.py ...                                                                                                                                                                                                                                   [  5%]
tests/test_transforms.py ................                                                                                                                                                                                                                      [ 10%]
tests/test_combiners.py ..                                                                                                                                                                                                                                     [ 11%]
tests/test_end_to_end.py ...                                                                                                                                                                                                                                   [ 12%]
tests/test_transforms.py ................                                                                                                                                                                                                                      [ 17%]
tests/test_combiners.py ..                                                                                                                                                                                                                                     [ 17%]
tests/test_end_to_end.py ...                                                                                                                                                                                                                                   [ 18%]
tests/test_transforms.py ................                                                                                                                                                                                                                      [ 23%]
tests/test_combiners.py ......                                                                                                                                                                                                                                 [ 25%]
tests/test_locking.py ......                                                                                                                                                                                                                                   [ 27%]
tests/test_openers.py ........................                                                                                                                                                                                                                 [ 35%]
tests/test_transforms.py ................                                                                                                                                                                                                                      [ 40%]
tests/test_openers.py ..........................................................................                                                                                                                                                               [ 63%]
tests/test_patterns.py ........ss.............ss..............                                                                                                                                                                                                 [ 75%]
tests/test_pipelines.py ........                                                                                                                                                                                                                               [ 78%]
tests/test_rechunking.py .....................................                                                                                                                                                                                                 [ 89%]
tests/test_storage.py .......                                                                                                                                                                                                                                  [ 92%]
tests/test_transforms.py ......................                                                                                                                                                                                                                [ 99%]
tests/test_utils.py ..                                                                                                                                                                                                                                         [ 99%]
tests/test_writers.py .                                                                                                                                                                                                                                        [100%]

========================================================================================================================== warnings summary ==========================================================================================================================

<EXCLUDED WARNINGS>

====================================================================================================== 314 passed, 5 skipped, 28 warnings in 190.78s (0:03:10) =======================================================================================================
```

### DIFF
--- a/pangeo_forge_recipes/aggregation.py
+++ b/pangeo_forge_recipes/aggregation.py
@@ -210,16 +210,8 @@ def _to_variable(template, target_chunks):
     chunks = tuple(target_chunks[dim] for dim in dims)
     # we pick zeros as the safest value to initialize empty data with
     # will only be used for dimension coordinates
-    # WARNING: there are lots of edge cases aroudn time!
-    # Xarray will pick a time encoding for the dataset (e.g. "days since days since 1970-01-01")
-    # and this may not be compatible with the actual values in the time coordinate
-    # (which we don't know yet)
-    # TODO: previous comment regarding encoding should no longer
-    # be relevant now that variable encoding will be used if available
     data = dsa.zeros(shape=shape, chunks=chunks, dtype=dtype)
     # TODO: add more encoding
-    # TODO: is the previous comment still relevant now that
-    # variable encoding will be used if available?
     encoding = template.get("encoding", {})
     encoding["chunks"] = chunks
     return xr.Variable(dims=dims, data=data, attrs=template["attrs"], encoding=encoding)
--- a/pangeo_forge_recipes/aggregation.py
+++ b/pangeo_forge_recipes/aggregation.py
@@ -23,11 +23,17 @@ class XarraySchema(TypedDict):
 
 
 def dataset_to_schema(ds: xr.Dataset) -> XarraySchema:
-    """Convert the output of `dataset.to_dict(data=False)` to a schema
+    """Convert the output of `dataset.to_dict(data=False, encoding=True)` to a schema
     (Basically justs adds chunks, which is not part of the Xarray ouput).
     """
 
-    d = ds.to_dict(data=False)
+    # Remove redundant encoding options
+    for v in ds.variables:
+        for option in ['_FillValue', 'source']:
+            # TODO: should be okay to remove _FillValue?
+            if option in ds[v].encoding:
+                del ds[v].encoding[option]
+    d = ds.to_dict(data=False, encoding=True)
     return XarraySchema(
         attrs=d.get("attrs"),
         coords=d.get("coords"),
@@ -164,6 +170,8 @@ def _combine_vars(v1, v2, concat_dim, allow_both=False):
                 raise DatasetCombineError(f"Can't merge datasets with the same variable {vname}")
             attrs = _combine_attrs(v1[vname]["attrs"], v2[vname]["attrs"])
             dtype = _combine_dtype(v1[vname]["dtype"], v2[vname]["dtype"])
+            # Can combine encoding using the same approach as attrs
+            encoding = _combine_attrs(v1[vname]["encoding"], v2[vname]["encoding"])
             (d1, s1), (d2, s2) = (
                 (v1[vname]["dims"], v1[vname]["shape"]),
                 (v2[vname]["dims"], v2[vname]["shape"]),
@@ -182,7 +190,13 @@ def _combine_vars(v1, v2, concat_dim, allow_both=False):
                     )
                 else:
                     shape.append(l1)
-            new_vars[vname] = {"dims": dims, "attrs": attrs, "dtype": dtype, "shape": tuple(shape)}
+            new_vars[vname] = {"dims": dims,
+                               "attrs": attrs,
+                               "dtype": dtype,
+                               "shape": tuple(shape),
+                               "encoding": encoding
+                               }
+
     return new_vars
 
 
@@ -199,9 +213,14 @@ def _to_variable(template, target_chunks):
     # Xarray will pick a time encoding for the dataset (e.g. "days since days since 1970-01-01")
     # and this may not be compatible with the actual values in the time coordinate
     # (which we don't know yet)
+    # TODO: previous comment regarding encoding should no longer
+    # be relevant now that variable encoding will be used if available
     data = dsa.zeros(shape=shape, chunks=chunks, dtype=dtype)
     # TODO: add more encoding
-    encoding = {"chunks": chunks}
+    # TODO: is the previous comment still relevant now that
+    # variable encoding will be used if available?
+    encoding = template.get("encoding", {})
+    encoding["chunks"] = chunks
     return xr.Variable(dims=dims, data=data, attrs=template["attrs"], encoding=encoding)
 
 
--- a/pangeo_forge_recipes/aggregation.py
+++ b/pangeo_forge_recipes/aggregation.py
@@ -29,7 +29,7 @@ def dataset_to_schema(ds: xr.Dataset) -> XarraySchema:
 
     # Remove redundant encoding options
     for v in ds.variables:
-        for option in ['_FillValue', 'source']:
+        for option in ["_FillValue", "source"]:
             # TODO: should be okay to remove _FillValue?
             if option in ds[v].encoding:
                 del ds[v].encoding[option]
@@ -190,12 +190,13 @@ def _combine_vars(v1, v2, concat_dim, allow_both=False):
                     )
                 else:
                     shape.append(l1)
-            new_vars[vname] = {"dims": dims,
-                               "attrs": attrs,
-                               "dtype": dtype,
-                               "shape": tuple(shape),
-                               "encoding": encoding
-                               }
+            new_vars[vname] = {
+                "dims": dims,
+                "attrs": attrs,
+                "dtype": dtype,
+                "shape": tuple(shape),
+                "encoding": encoding,
+            }
 
     return new_vars
 
--- a/tests/test_aggregation.py
+++ b/tests/test_aggregation.py
@@ -32,7 +32,7 @@ def test_schema_to_template_ds(specified_chunks):
             expected_chunksize = _expected_chunks(size, specified_chunks.get(dim, None))
             assert chunksize == expected_chunksize
     # Confirm original time units have been preserved
-    assert ds.time.encoding["units"] == dst.time.encoding["units"]
+    assert ds.time.encoding.get("units") == dst.time.encoding.get("units")
     schema2 = dataset_to_schema(dst)
     assert schema == schema2
 
--- a/tests/test_aggregation.py
+++ b/tests/test_aggregation.py
@@ -32,7 +32,7 @@ def test_schema_to_template_ds(specified_chunks):
             expected_chunksize = _expected_chunks(size, specified_chunks.get(dim, None))
             assert chunksize == expected_chunksize
     # Confirm original time units have been preserved
-    assert ds.time.encoding['units'] == dst.time.encoding['units']
+    assert ds.time.encoding["units"] == dst.time.encoding["units"]
     schema2 = dataset_to_schema(dst)
     assert schema == schema2
 
--- a/tests/test_aggregation.py
+++ b/tests/test_aggregation.py
@@ -31,6 +31,8 @@ def test_schema_to_template_ds(specified_chunks):
             chunksize = var.chunksizes[dim]
             expected_chunksize = _expected_chunks(size, specified_chunks.get(dim, None))
             assert chunksize == expected_chunksize
+    # Confirm original time units have been preserved
+    assert ds.time.encoding['units'] == dst.time.encoding['units']
     schema2 = dataset_to_schema(dst)
     assert schema == schema2
 
--- a/tests/test_writers.py
+++ b/tests/test_writers.py
@@ -140,5 +140,4 @@ def test_store_dataset_fragment(temp_store):
     xr.testing.assert_identical(ds, ds_target)
     # assert_identical() doesn't check encoding
     # Checking the original time encoding units should be sufficient
-    # Zarr retains the original "days since %Y:%m%d" and removes " %H:%M:%S"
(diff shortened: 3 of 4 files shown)


pangeo_forge_recipes/aggregation.py
```diff
@@ -23,11 +23,17 @@ class XarraySchema(TypedDict):
...
 def dataset_to_schema(ds: xr.Dataset) -> XarraySchema:
-    """Convert the output of `dataset.to_dict(data=False)` to a schema
+    """Convert the output of `dataset.to_dict(data=False, encoding=True)` to a schema
     (Basically justs adds chunks, which is not part of the Xarray ouput).
     """
 
-    d = ds.to_dict(data=False)
+    # Remove redundant encoding options
+    for v in ds.variables:
+        for option in ["_FillValue", "source"]:
```
rabernat, line 32: Could you explain the rationale for special casing these two options?

pangeo_forge_recipes/aggregation.py
```diff
@@ -23,11 +23,17 @@ class XarraySchema(TypedDict):
...
+            # TODO: should be okay to remove _FillValue?
+            if option in ds[v].encoding:
+                del ds[v].encoding[option]
+    d = ds.to_dict(data=False, encoding=True)
```
rabernat, line 36: I think that when I first started working on this, this option didn't even exist yet! See https://github.com/pydata/xarray/issues/6634

Nice when things come together. 😄

pangeo_forge_recipes/aggregation.py
```diff
@@ -164,6 +170,8 @@ def _combine_vars(v1, v2, concat_dim, allow_both=False):
                 raise DatasetCombineError(f"Can't merge datasets with the same variable {vname}")
             attrs = _combine_attrs(v1[vname]["attrs"], v2[vname]["attrs"])
             dtype = _combine_dtype(v1[vname]["dtype"], v2[vname]["dtype"])
+            # Can combine encoding using the same approach as attrs
+            encoding = _combine_attrs(v1[vname]["encoding"], v2[vname]["encoding"])
```
rabernat, line 174: Brilliant!

pangeo_forge_recipes/aggregation.py
```diff
@@ -199,9 +214,14 @@ def _to_variable(template, target_chunks):
     # Xarray will pick a time encoding for the dataset (e.g. "days since days since 1970-01-01")
     # and this may not be compatible with the actual values in the time coordinate
     # (which we don't know yet)
+    # TODO: previous comment regarding encoding should no longer
+    # be relevant now that variable encoding will be used if available
```
rabernat, line 218: Delete the irrelevant comment please. 🙏

tests/test_aggregation.py
```diff
@@ -31,6 +31,8 @@ def test_schema_to_template_ds(specified_chunks):
             chunksize = var.chunksizes[dim]
             expected_chunksize = _expected_chunks(size, specified_chunks.get(dim, None))
             assert chunksize == expected_chunksize
+    # Confirm original time units have been preserved
+    assert ds.time.encoding["units"] == dst.time.encoding["units"]
```
rabernat, line 35: Would this test fail without the changes in `aggregation.py`?

tests/test_writers.py
```diff
@@ -138,3 +138,7 @@ def test_store_dataset_fragment(temp_store):
     ds_target = xr.open_dataset(temp_store, engine="zarr").load()
 
     xr.testing.assert_identical(ds, ds_target)
+    # assert_identical() doesn't check encoding
+    # Checking the original time encoding units should be sufficient
+    # Zarr retains the original "days since %Y:%m%d" and removes " %H:%M:%S"
+    assert " ".join(ds.time.encoding["units"].split(" ")[0:-1]) == ds_target.time.encoding["units"]
```
rabernat, line 144: Zarr has nothing to do with it. Xarray and cftime are managing this logic. I think it drops of the time if the time is 0:0:0.

Wouldn't it be easier to fix this at the source (i.e. strip the time in line 39 of `data_generation.py`)? That way you can just do

```python
assert ds.time.encoding == ds_target.time.encoding
```

rabernat (commented):
This looks great. A few thoughts below.

pangeo_forge_recipes/aggregation.py
derekocallaghan: I excluded these as they were causing certain test failures where expected schemas were compared with actual. E.g. any combiner tests using `has_correct_schema()`:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/beam-refactor/tests/test_combiners.py#L98-L102

```python
def has_correct_schema(expected_schema):
    def _check_results(actual):
        assert len(actual) == 1
        schema = actual[0]
        assert schema == expected_schema
```

The `source` will be unique to each original source data product, and the `_FillValue` appeared to be added automatically (I can't recall the specific issue with the latter though).

tests/test_aggregation.py
derekocallaghan: I checked and it would fail for the wrong reason (`KeyError`) in that situation, I've fixed it and will commit (also in `test_witers`)

pangeo_forge_recipes/aggregation.py
derekocallaghan: Will do

tests/test_writers.py
derekocallaghan: I originally stripped it in `data_generation.py` as you suggest, but I was sure that this was causing failures only in `test_writer.py`. I've restored this and all tests are passing, so I'm not sure what went wrong previously. Commit is on the way.

pangeo_forge_recipes/aggregation.py
derekocallaghan: I checked the latter again, and when `_FillValue` is retained, it's being set to `nan` in `expected_schema` (as generated by the original `ds.to_dict(data=False, encoding=True)`, but only for the `lat` and `lon` coords. However, the `actual` schema doesn't contain `_FillValue` for `lat`/`lon`, and the assert fails.

rabernat:
I've been thinking through this more, and and I'm worried that there is an important edge case that is not covered by this. Currently, if `encoding['units']` is not the same across all of the datasets, it will simply be dropped, following the logic in `_combine_attrs`:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/241ca87c618c30e609fd264d912d113b41ad5e7d/pangeo_forge_recipes/aggregation.py#L136-L145

So when pangeo forge goes to create the target dataset, it will not have any information about the time resolution and needed time encoding, and we will be back in the situation described by this comment:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/241ca87c618c30e609fd264d912d113b41ad5e7d/pangeo_forge_recipes/aggregation.py#L196-L202

This PR and the use case that motivated it assume that the input files will _all have the same encoding_. But what if they have different encoding?

For example, we could have input files representing hourly data with time encoded like this
```
0 hours since 2000-01-01 00:00:00
0 hours since 2000-01-01 01:00:00
0 hours since 2000-01-01 01:00:00
```
(yes I have seen this in real datasets)

since the time encoding is not uniform, `units` would just be dropped. And Xarray would probably pick `days since ...` as the encoding when initializing the target dataset. This would probably screw up the time coordinate.

However, xarray would handle the data fine in an `open_mfdataset` situation, because it would fist decode each time to an actual datetime type. Upon saving, it would determine the encoding using [this function](https://github.com/pydata/xarray/blob/522ee2210499cfa434a0ca0825790949a8e08ff0/xarray/coding/times.py#L389).

If we stick to our current approach of not reading and combining the actual coordinates, we will have to recreate some of this logic within `aggregate.py`. Specifically, we will want to determine the minimum frequency (days, hours, s, etc.) and reference date for the time encoding by intelligently combining the encoding of each dataset.

That could be a heavy lift. So perhaps, for the interim, we would just want to raise an error if the time encoding is different for different datasets?

More test cases would be useful at exposing edge cases.

derekocallaghan:
> 
> So when pangeo forge goes to create the target dataset, it will not have any information about the time resolution and needed time encoding, and we will be back in the situation described by this comment:
> 
> https://github.com/pangeo-forge/pangeo-forge-recipes/blob/241ca87c618c30e609fd264d912d113b41ad5e7d/pangeo_forge_recipes/aggregation.py#L196-L202
> 

Based on this, I should restore some of the original warning comment text.

> This PR and the use case that motivated it assume that the input files will _all have the same encoding_. But what if they have different encoding?

Yep, I believe the CCMP files all have the same encoding (units).

> 
> However, xarray would handle the data fine in an `open_mfdataset` situation, because it would fist decode each time to an actual datetime type. Upon saving, it would determine the encoding using [this function](https://github.com/pydata/xarray/blob/522ee2210499cfa434a0ca0825790949a8e08ff0/xarray/coding/times.py#L389).
> 

When I originally created the Zarr store before the Pangeo Forge recipe, I used `open_mfdataset()`, which worked and generated a suitable time encoding. (I was thinking of suggesting this just before I read your comment here).

> If we stick to our current approach of not reading and combining the actual coordinates, we will have to recreate some of this logic within `aggregate.py`. Specifically, we will want to determine the minimum frequency (days, hours, s, etc.) and reference date for the time encoding by intelligently combining the encoding of each dataset.
> 

I think replicating or reusing the functionality in [`xarray.coding.time.infer_datetime_units()`](https://github.com/pydata/xarray/blob/522ee2210499cfa434a0ca0825790949a8e08ff0/xarray/coding/times.py#L389) may be something worth looking at, possibly as a custom transform that precedes `DetermineSchema` in `StoreToZarr.expand()`, or instead in `OpenWithinXarray.expand()`. I need to something not too far removed for the ASCAT recipes, where I'm hoping to use a custom transform to reorder the collection based on a modified `time`, prior to `StoreToZarr`. I'm not sure yet whether this is possible, but I can look into it.

> That could be a heavy lift. So perhaps, for the interim, we would just want to raise an error if the time encoding is different for different datasets?
> 

That can definitely be done.

rabernat:
Actually, a nice middle ground would be to make it really easy to specify the desired time encoding. Like, if I know a priori that my data are daily resolution, I could just put the time encoding directly in the recipe. I'm wondering what would be the right way to specify this... 🤔

derekocallaghan:
> Actually, a nice middle ground would be to make it really easy to specify the desired time encoding. Like, if I know a priori that my data are daily resolution, I could just put the time encoding directly in the recipe. I'm wondering what would be the right way to specify this... thinking

Would this only work when all input files had the same time encoding? E.g. I guess we'd be back to the current issue in situations like your example above that has multiple units:

```
0 hours since 2000-01-01 00:00:00
0 hours since 2000-01-01 01:00:00
0 hours since 2000-01-01 02:00:00
```

rabernat:
> Would this only work when all input files had the same time encoding?

No, because the data are decoded before they are written, and then re-encoded to match the target encoding.

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/81a9ce53ef6e7e93e3a00d4cbf25735fc38f61b3/pangeo_forge_recipes/writers.py#L27-L36

derekocallaghan:
> > Would this only work when all input files had the same time encoding?
> 
> No, because the data are decoded before they are written, and then re-encoded to match the target encoding.
> 

I understand it now, i.e. the specified target encoding would override any potential variation in input file encodings.

rabernat:
I opened #480 to track the idea of specifying encoding in the recipe. We can address that in a follow-up PR. For now this is a big step forward.

PR #474, opened 2023-01-16
Make FSSpecTarget be 'pathlike'

yuvipanda:
The goal here is for recipe authors to be able to
pass *strings* to `target` parameters of StoreToZarr to specify where the data should be for this specific recipe, and we can inject the prefix with pangeo-forge-runner.

The primary use case is when we have a dictionary of recipes being returned from a single recipe file. The *name* of the recipe is unfortunately not really available during AST parsing, so the suffix the data should be put in needs to be explicitly specified.

### DIFF
--- a/pangeo_forge_recipes/storage.py
+++ b/pangeo_forge_recipes/storage.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import hashlib
 import io
 import json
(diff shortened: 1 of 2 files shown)


rabernat (approved):
No objections to adding this method. But where is it actually used? I just see the `__truediv__` method and its test.

Edit: ok I assume it is used by Runner.

PR #489, opened 2023-02-08
Remove prefect and dask executors

norlandrhagen:
https://github.com/pangeo-forge/pangeo-forge-recipes/issues/475

PR #491, opened 2023-02-13
Test python 3.9 and 3.10, not 3.8 and 3.9

yuvipanda:
Based on
https://docs.xarray.dev/en/stable/whats-new.html#breaking-changes, xarray's latest version no longer supports Python 3.8. This causes test to fail in
https://github.com/pangeo-forge/pangeo-forge-recipes/pull/489.

So we bump up to 3.9 and 3.10!

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -50,7 +50,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8, 3.9]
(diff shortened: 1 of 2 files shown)


cisaacstern (approved):
Thanks Yuvi! LGTM, and even better if we can figure out what's going on with pre-commit before merging?

yuvipanda:
@cisaacstern  fixed in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/492

PR #487, opened 2023-02-02
Port tutorial notebooks to use Beam

derekocallaghan:
This PR will be used to port tutorial notebooks to `beam-refactor`:

The first is an initial port of the *NetCDF Zarr Multi-Variable Sequential Recipe: NOAA World Ocean Atlas* notebook to use the `beam-refactor` approach:
* A custom transform is provided, containing the preprocessor function logic in the non-Beam notebook version. I've included some links to Beam documentation here but this could be expanded if required.
* The recipe approach is similar to that described in [`intro_tutorial_part2`](https://pangeo-forge--445.org.readthedocs.build/en/445/introduction_tutorial/intro_tutorial_part2.html#create-the-recipe-object-beam-composite-transform)
  * It currently creates and specifies a `target_path` to `StoreToZarr`. See #485 for discussion on this point as it may be subject to change.
* I've run it locally and rebuilt the docs, it seems to work without problems. The only issue I spotted was the final plot range is different to the non-Beam version, but I couldn't see what the problem was and I don't think it's related to Beam usage.
* Although it contains references to the [*Xarray-to-Zarr Sequential Recipe: NOAA OISST* notebook](https://pangeo-forge--445.org.readthedocs.build/en/445/pangeo_forge_recipes/tutorials/xarray_zarr/netcdf_zarr_sequential.html), I haven't ported this yet as it contains functionality which may no longer be provided by the `beam-refactor` branch. I'm going to look at this next. In the meantime, this current notebook may be useful as an example for porting other tutorial notebooks (@norlandrhagen).

Partly fixes #483

PR #495, opened 2023-02-16
Add store_name to StoreToZarr

yuvipanda:
When generating multiple Zarr datasets from the same recipe, we want to support specifying *where* each dataset will go inside the same target, as target is set by pangeo-forge-runner. Without, this subpath is specified as the *name* of the recipe itself. However, due to how ast rewriting works, we don't actually have the name of the recipe when constructing our beam pipeline, so we can not use that. So instead, it would have to be explicitly specified as a subpath inside the target.

### DIFF
--- a/pangeo_forge_recipes/transforms.py
+++ b/pangeo_forge_recipes/transforms.py
@@ -269,6 +269,8 @@ class StoreToZarr(beam.PTransform):
 
     :param combine_dims: The dimensions to combine
     :param target: Where to store the target Zarr dataset.
+    :param target_subpath: Subpath *inside* the target to store this Zarr dataset.
+                           Useful when generating multiple Zarr datasets from the same recipe.
     :param target_chunks: Dictionary mapping dimension names to chunks sizes.
         If a dimension is a not named, the chunks will be inferred from the data.
     """
@@ -277,12 +279,21 @@ class StoreToZarr(beam.PTransform):
     # Could be inferred from the pattern instead
     combine_dims: List[Dimension]
     target: str | FSSpecTarget
+    target_subpath: Optional[str] = None
     target_chunks: Dict[str, int] = field(default_factory=dict)
 
     def expand(self, datasets: beam.PCollection) -> beam.PCollection:
         schema = datasets | DetermineSchema(combine_dims=self.combine_dims)
         indexed_datasets = datasets | IndexItems(schema=schema)
+        if isinstance(self.target, str):
+            target_root = FSSpecTarget.from_url(self.target)
+        else:
+            target_root = self.target
+        if self.target_subpath:
+            full_target = target_root / self.target_subpath
+        else:
+            full_target = target_root
         target_store = schema | PrepareZarrTarget(
-            target=self.target, target_chunks=self.target_chunks
+            target=full_target, target_chunks=self.target_chunks
         )
         return indexed_datasets | StoreDatasetFragments(target_store=target_store)
--- a/pangeo_forge_recipes/transforms.py
+++ b/pangeo_forge_recipes/transforms.py
@@ -268,9 +268,9 @@ class StoreToZarr(beam.PTransform):
     """Store a PCollection of Xarray datasets to Zarr.
 
     :param combine_dims: The dimensions to combine
-    :param target: Where to store the target Zarr dataset.
-    :param target_subpath: Subpath *inside* the target to store this Zarr dataset.
-                           Useful when generating multiple Zarr datasets from the same recipe.
+    :param target: Location that will contain the target Zarr Store.
+    :param store_name: Name for the Zarr store. It will be created with this name
+                       under `target`.
     :param target_chunks: Dictionary mapping dimension names to chunks sizes.
         If a dimension is a not named, the chunks will be inferred from the data.
     """
@@ -279,7 +279,7 @@ class StoreToZarr(beam.PTransform):
     # Could be inferred from the pattern instead
     combine_dims: List[Dimension]
     target: str | FSSpecTarget
-    target_subpath: Optional[str] = None
+    store_name: str
     target_chunks: Dict[str, int] = field(default_factory=dict)
 
     def expand(self, datasets: beam.PCollection) -> beam.PCollection:
@@ -289,10 +289,7 @@ class StoreToZarr(beam.PTransform):
             target_root = FSSpecTarget.from_url(self.target)
         else:
             target_root = self.target
-        if self.target_subpath:
-            full_target = target_root / self.target_subpath
-        else:
-            full_target = target_root
+        full_target = target_root / self.store_name
         target_store = schema | PrepareZarrTarget(
             target=full_target, target_chunks=self.target_chunks
         )
--- a/pangeo_forge_recipes/transforms.py
+++ b/pangeo_forge_recipes/transforms.py
@@ -268,9 +268,9 @@ class StoreToZarr(beam.PTransform):
     """Store a PCollection of Xarray datasets to Zarr.
 
     :param combine_dims: The dimensions to combine
-    :param target: Location that will contain the target Zarr Store.
+    :param target_root: Location the Zarr store will be created inside.
     :param store_name: Name for the Zarr store. It will be created with this name
-                       under `target`.
+                       under `target_root`.
     :param target_chunks: Dictionary mapping dimension names to chunks sizes.
         If a dimension is a not named, the chunks will be inferred from the data.
     """
@@ -278,17 +278,17 @@ class StoreToZarr(beam.PTransform):
     # TODO: make it so we don't have to explictly specify combine_dims
     # Could be inferred from the pattern instead
     combine_dims: List[Dimension]
-    target: str | FSSpecTarget
+    target_root: str | FSSpecTarget
     store_name: str
     target_chunks: Dict[str, int] = field(default_factory=dict)
 
     def expand(self, datasets: beam.PCollection) -> beam.PCollection:
         schema = datasets | DetermineSchema(combine_dims=self.combine_dims)
         indexed_datasets = datasets | IndexItems(schema=schema)
-        if isinstance(self.target, str):
-            target_root = FSSpecTarget.from_url(self.target)
+        if isinstance(self.target_root, str):
+            target_root = FSSpecTarget.from_url(self.target_root)
         else:
-            target_root = self.target
+            target_root = self.target_root
         full_target = target_root / self.store_name
         target_store = schema | PrepareZarrTarget(
             target=full_target, target_chunks=self.target_chunks
--- a/tests/test_end_to_end.py
+++ b/tests/test_end_to_end.py
@@ -1,3 +1,5 @@
+import os
+
 import apache_beam as beam
 import pytest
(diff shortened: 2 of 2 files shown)


yuvipanda:
Would love to know what you think of this API too, @derekocallaghan

yuvipanda:
@derekocallaghan see https://github.com/pangeo-forge/staged-recipes/pull/213/files for an example!

cisaacstern (commented):
@yuvipanda I see why you're proposing this and technically speaking it seems reasonable to me.

I'm a little unclear on one point, which is, if we do not know the name (i.e. dict key) of the recipe at ast-rewrite time, how do we then generate a descriptive path name to `target_subpath`?

Or there is no way to have a human-readable/descriptive subpath name in the case of dictionaries of recipes?

yuvipanda:
@cisaacstern the writer of the recipe knows, but the *ast* does not. For example, in https://github.com/pangeo-forge/staged-recipes/pull/213/files#diff-fa5594d21a1885b459df45326bdabe05cde8685094f7acfddd4f53d132c729ecR96, the 'name' is assigned (as it is just the key of the dict), but that is separate from the construction of the pipeline itself. So given there are a million ways you can put a key in a dict (and even change it after), ast rewriting there would be pretty difficult.

yuvipanda:
> Or there is no way to have a human-readable/descriptive subpath name in the case of dictionaries of recipes?

No way to do this *automatically*. You'll see in the linked PR above it's done manually, and actually leads to what I think of as a better outcome - zarr stores are `<region>/<variable>`, which wouldn't be possible if it was just recipe name, as you can't put `/` in names.

cisaacstern:
> You'll see in the linked PR above it's done manually, and actually leads to what I think of as a better outcome - zarr stores are <region>/<variable>

To make sure I understand clearly, the keys the recipe contributor gives in the dict, _can_ be assigned (or transformed, then assigned) to the `target_subpath` at ast-rewrite time?

yuvipanda:
they can not be. If you're using dicts to write multiple zarr stores from same recipe, you must manually specify target_subpath.

cisaacstern:
> they can not be. If you're using dicts to write multiple zarr stores from same recipe, you must manually specify target_subpath.

In that case, if a dict of recipes is deployed automatically from GitHub (post-merge, e.g.) by Pangeo Forge Cloud, what logic do we use to assign `target_subpath` names to the recipes run from that dictionary?

yuvipanda:
@cisaacstern if there is only one recipe being run and `target_subpath` will be empty, so the data will reside directly wherever `target` is set. So if it's set to `gs://yuvipanda-test-bucket`, the zarr store will just live there. If instead it should be in `gs://yuvipanda-test-bucket/store1`, `store1` that should be specified as `target_subpath`. 

If there are multiple zarr stores from the same recipe, they will all have the same `target` - say `gs://yuvipanda-test-bucket`. So *each* will have a different `target_subpath`(like `store2`, `store3`, etc) so they don't all try to write to the exact same path `gs://yuvipanda-test-bucket`, but to `gs://yuvipanda-test-bucket/store2`, `gs://yuvipanda-test-bucket/store3` etc

In https://github.com/pangeo-forge/staged-recipes/pull/213/files#diff-fa5594d21a1885b459df45326bdabe05cde8685094f7acfddd4f53d132c729ecR90 you can see `target_subpath` is automatically generated by using the name of the variables and region. The key of the dict is also generated using those two variables (https://github.com/pangeo-forge/staged-recipes/pull/213/files#diff-fa5594d21a1885b459df45326bdabe05cde8685094f7acfddd4f53d132c729ecR96) just elsewhere.

So `target` will be automatically generated as it is now, but `target_subpath` will not be. I hope that makes sense?

cisaacstern (approved):
@yuvipanda thank you for patiently explaining that to me, and apologies that it took me a little bit to grok.

I now see that the _recipe contributor will give us the `target_subpath`_ (if one is needed), so design-wise we are in the same place as we were before, except that the subpath is given explicitly as a kwarg, rather than as a dict key. (And yes, a dict is still specified in the recipe, it's just that the key is not used to determine the subpath. Confusing things might happen if the dict keys and `target_subpaths` use different naming conventions, but that would be a documentation/linting/user-error concern for later on.)

This makes sense to me and I'm 👍

derekocallaghan:
Hi @yuvipanda, thanks for the link above (and for the conversation since). I understand the motivation having had a look through the daymet recipe and the code changes. The only thing I'd suggest might be to rename `target_subpath`, e.g. to `store_name` or something similar to reflect the `basename` aspect. This might be more intuitive to recipe authors, as if I understand, here they could specify `target_subpath` without also specifying `target`. I think `target` was recently renamed from `target_path` so now we'd have a `subpath` without a `path` (even though `target` is essentially a `path`).

This is only a minor point though.

yuvipanda:
No worries @cisaacstern - thank you for reasoning through with me!

@derekocallaghan given that this is optional, I'm hesitant to call it 'store_name' or something like that. Although, perhaps the way out of that is to make it *required*, and call it `store_name` - perhaps we *never* want the store to be directly at `target`?

derekocallaghan:
> @derekocallaghan given that this is optional, I'm hesitant to call it 'store_name' or something like that. Although, perhaps the way out of that is to make it _required_, and call it `store_name` - perhaps we _never_ want the store to be directly at `target`?

I think the latter suggestion might make more sense tbh, this way the recipe author has to explicitly name the recipe/transform output, but not necessarily where it's stored (`target`).

yuvipanda:
@derekocallaghan @cisaacstern what do you think now? I've called it `store_name` and made it required.

derekocallaghan:
> @derekocallaghan @cisaacstern what do you think now? I've called it `store_name` and made it required.

Thanks @yuvipanda, that looks good to me.

pangeo_forge_recipes/transforms.py
```diff
@@ -268,7 +268,9 @@ class StoreToZarr(beam.PTransform):
     """Store a PCollection of Xarray datasets to Zarr.
 
     :param combine_dims: The dimensions to combine
-    :param target: Where to store the target Zarr dataset.
+    :param target: Location that will contain the target Zarr Store.
```
cisaacstern, line 271:
```suggestion
    :param target_prefix: Location that will contain the target Zarr Store.
```

pangeo_forge_recipes/transforms.py
```diff
@@ -277,12 +279,18 @@ class StoreToZarr(beam.PTransform):
     # Could be inferred from the pattern instead
     combine_dims: List[Dimension]
     target: str | FSSpecTarget
```
cisaacstern, line 281:
```suggestion
    target_prefix: str | FSSpecTarget
```

cisaacstern (commented):
@yuvipanda this LGTM my only nit now that I think about it, is how do you feel about renaming `target` to something like, `target_prefix`, so it's a bit clearer how this relates, positionally, to the `store_name`?

cisaacstern:
@yuvipanda I'm not tied to `target_prefix` specifically (and if anything, it's imperfect, because maybe it implies the existence of something named "suffix" or "subpath", which does not exist)... but the bigger thing I'm after is trying to clarify the meaning of what is currently called `target`, especially considering that _in prior releases of pangeo-forge-recipes `target` means what we are now defining as `target` + `store_name`_... which feels possibly confusing to me?

Open to any suggestions on the specifics names.

keewis:
what about `target_root`?

cisaacstern:
@keewis, thanks for chiming in! Yes, `target_root` is ideal, IMO!

@yuvipanda, @derekocallaghan, thoughts?

pangeo_forge_recipes/transforms.py
```diff
@@ -268,7 +268,9 @@ class StoreToZarr(beam.PTransform):
     """Store a PCollection of Xarray datasets to Zarr.
 
     :param combine_dims: The dimensions to combine
-    :param target: Where to store the target Zarr dataset.
+    :param target: Location that will contain the target Zarr Store.
+    :param store_name: Name for the Zarr store. It will be created with this name
```
rabernat, line 272:
```suggestion
    :param group_path: Name of the group within the store in which to store the dataset.
```

tests/test_end_to_end.py
```diff
@@ -37,11 +39,35 @@ def test_xarray_zarr(
...
+):
+    pattern = netcdf_local_file_pattern_sequential
+    with pipeline as p:
+        (
+            p
+            | beam.Create(pattern.items())
+            | OpenWithXarray(file_type=pattern.file_type)
+            | StoreToZarr(
+                target=tmp_target_url,
+                store_name="subpath",
```
rabernat, line 67:
```suggestion
                group_path="subpath",
```

rabernat (commented):
I think we should be using the concept of a [**Zarr group**](https://zarr.readthedocs.io/en/stable/tutorial.html#groups) to address this scenario.

The bakery (or whatever we are calling it) will provide a **Zarr store**. The recipe may define a **group within that store** where the results should be placed.

cisaacstern:
> The bakery (or whatever we are calling it) will provide a Zarr store. The recipe may define a group within that store where the results should be placed.

Would this require changes to [`PrepareZarrTarget`](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/75d0d925cec7a3e8ca928950f952a874a5c997d7/pangeo_forge_recipes/transforms.py#L209)?

More generally, could you clarify what advantage(s) this hierarchical approach might offer (to "bakery" operators, recipe contributors, and/or data users)?

yuvipanda:
Hmm, that would mean that there *must* be a 1:1 correspondence between a *feedstock* and a zarr store. You can not produce multiple zarr stores from one feedstock. IMO, that feels like a pretty strong contraint, and a big difference vs how things are right now. While I agree that groups are probably a better fit, we can *add* that as functionality perhaps, allowing recipe writers to choose between multiple individual zarr stores vs zarr groups.

yuvipanda:
I like the `target_root` definition, @derekocallaghan so I've modified the PR to make it so. I'll make a companion PR to pangeo-forge-runner as well.

pangeo_forge_recipes/transforms.py
yuvipanda: Done as `target_root`

yuvipanda:
Also note that `store_name` is *required*, even if you are only doing a single zarr store output - so that to me also makes it a distinctly different feature than Zarr groups, as you don't *have* to use groups.

cisaacstern:
@rabernat, from syncing offline with @yuvipanda, I've learned that merging this PR today would be helpful to his upcoming presentation on Monday with @briannapagan. So I'm going to merge this as-is now, and I've opened #498 to track your suggestion above.

derekocallaghan:
> I like the `target_root` definition, @derekocallaghan so I've modified the PR to make it so. I'll make a companion PR to pangeo-forge-runner as well.

Thanks @yuvipanda, I'm just reading these comments this morning. I've modified my recent [recipe tutorial notebook changes to include use of `store_name` and `target_root`](https://pangeo-forge--487.org.readthedocs.build/en/487/pangeo_forge_recipes/tutorials/index.html) (#487).

PR #486, opened 2023-02-02
beam-refactor reference recipe

norlandrhagen:
This PR intends to add the ability for the `beam-refactor` branch of Pangeo-Forge to run `Kerchunk` based `ReferenceRecipes`.

https://github.com/pangeo-forge/pangeo-forge-recipes/pull/452

PR #509, opened 2023-04-11
Remove unused modules, functions, & tests

cisaacstern:
Tests in #486 are currently failing due to stale pre-`beam-refactor` tests being incompatible with upstream changes in Dask.

This PR removes unused modules, functions, and associated tests from `beam-refactor`, which are no longer used post-refactor.

Did I delete anything that shouldn't have been removed? Are there more things which can be removed?


cisaacstern:
Though reviews are still pending (and I didn't wait that long), I'm going to merge this because:

- I believe the deletions here are uncontroversial
- All beam-refactor tests pass
- We need this to get tests to pass in #486

Of course happy to follow up by restoring any of the deleted files, if someone later finds that I've made a mistake here. Onwards to the release of `beam-refactor`!

PR #512, opened 2023-04-21
Pass through inline_threshold

TomAugspurger:
`HDFReferenceRecipe` takes an `inline_threshold`, but it wasn't being passed through to the function that uses it.


rabernat:
Thanks Tom! We should make sure this change survives the beam refactor (ref #486)

cisaacstern:
Thanks @TomAugspurger! I will adapt the test here for #486. We do pass `inline_threshold` through there, but didn't have a unit test for it!

moradology:
This is pointing at the old default branch and seems to point to work superseded by `MultiZarrToZarr`

Issue #517, opened 2023-04-26
[beam-refactor] StoreToZarr - Cannot Combine Fragments Possible Error


norlandrhagen:
Hi there, I'm working through debugging a recipe using the beam-refactor branch. In the final step of the pipeline, `StoreToZarr` fails and gives an error: `ValueError: Cannot combine fragments. Expected a hypercube of shape [1] but got 4 fragments.`. 

Not sure if this is an actual error or some issue with my environment. I tried running the [multi_variable_recipe](https://pangeo-forge--516.org.readthedocs.build/en/516/pangeo_forge_recipes/tutorials/xarray_zarr/multi_variable_recipe.html) notebook, which also combines multiple variables into a single dataset and got the same error.

To create my environment, I ran:
```
mamba env create --file ci/py3.9.yml
mamba activate pangeo-forge-recipes
pip install git+https://github.com/pangeo-forge/pangeo-forge-recipes.git@beam-refactor
```

MRE:

```python

import os
from tempfile import TemporaryDirectory
import apache_beam as beam

from pangeo_forge_recipes.transforms import OpenURLWithFSSpec, OpenWithXarray, StoreToZarr, Indexed, T
from pangeo_forge_recipes.patterns import MergeDim, ConcatDim, FilePattern


start_year = 2000
end_year = 2001
variables = ["shum", "tmin", "tmax", "wind"]


def format_function(variable, time):
    return (f"https://rda.ucar.edu/thredds/dodsC/files/g/ds314.0/0.25deg/daily/{variable}_0p25_daily_{time}-{time}.nc")

class DropZDim(beam.PTransform):
    """A preprocessing function to drop z dim"""

    @staticmethod
    def _drop_z_dim(item: Indexed[T]) -> Indexed[T]:
        index, ds = item
        ds = ds.squeeze("z").drop('z')
        return index, ds

    def expand(self, pcoll: beam.PCollection) -> beam.PCollection:
        return pcoll | beam.Map(self._drop_z_dim)


concat_dim = ConcatDim("time",keys=range(start_year, end_year + 1),nitems_per_file=1)
variable_merge_dim = MergeDim("variable", keys=variables)
pattern = FilePattern(format_function, variable_merge_dim, concat_dim, file_type='opendap')


td = TemporaryDirectory()
target_root = td.name
store_name = "output.zarr"
target_store = os.path.join(target_root, store_name)

transforms = (
    beam.Create(pattern.items())
[...]

PR #490, opened 2023-02-08
Port of Terraclimate example to beam-refactor

norlandrhagen:
https://github.com/pangeo-forge/pangeo-forge-recipes/issues/483

PR #521, opened 2023-05-02
Include merge dim positions in group keys emitted by `split_fragments`

norlandrhagen:
Opening up a PR to start working on a bugfix for the error described in Issue #517 .

The error from #517 happens in L185  of `rechunking.py` in the `combine_fragments` function.

As @rabernat pointed out in #517, the error might be upstream in the `split_fragments` function.

> Basically, it splits up the original dataset pieces into fragments, 
> and then puts them back together. Each merge dim should end up in a separate group. 

So if I'm understanding this correctly, `split_fragments` does not take the `MergeDim` into account and outputs PCollections/Fragments that contain multiple variables. When fed into `combine_fragments`, we get the error: `Expected a hypercube of shape [1] but got 2 fragments`

There is the begining of a test named `test_split_fragment_merge_dim` in `test_rechunking.py`

cc @cisaacstern

### DIFF
--- a/tests/test_rechunking.py
+++ b/tests/test_rechunking.py
@@ -258,5 +258,3 @@ def test_combine_fragments_errors():
     index1 = Index({Dimension("time", CombineOp.CONCAT): IndexedPosition(2)})
     with pytest.raises(ValueError, match="are not consistent"):
         _ = combine_fragments(group, [(index0, ds), (index1, ds)])
-
-    #
--- a/tests/test_rechunking.py
+++ b/tests/test_rechunking.py
@@ -11,8 +11,7 @@ from .data_generation import make_ds
 
 
 @pytest.mark.parametrize("nt", [2])
-@pytest.mark.parametrize("nvars", [2])
-def test_split_and_combine_fragments_with_merge_dim(nt, nvars):
+def test_split_and_combine_fragments_with_merge_dim(nt):
     """Test if sub-fragments split from datasets with merge dims can be combined with each other."""
 
     t_offset = 0
--- a/tests/test_rechunking.py
+++ b/tests/test_rechunking.py
@@ -61,8 +61,8 @@ def test_split_and_combine_fragments_with_merge_dim(nt_dayparam, time_chunks):
     for g in sorted(groupkeys):
         # just confirms that grouping logic within this test is correct
         assert all([sf.groupkey == g for sf in grouped_subfragments[g]])
-        # for each subfragment in the current group, assert that there is only merge dimension
-        # positional value present. this verifies that `split_fragments` has not incorrectly
+        # for the merge dimension of each subfragment in the current group, assert that there
+        # is only one positional value present. this verifies that `split_fragments` has not
         # grouped distinct merge dimension positional values together under the same groupkey.
         merge_position_vals = [sf.content[0][merge_dim].value for sf in grouped_subfragments[g]]
         assert all([v == merge_position_vals[0] for v in merge_position_vals])
--- a/tests/test_rechunking.py
+++ b/tests/test_rechunking.py
@@ -15,24 +15,30 @@ from .data_generation import make_ds
 def test_split_and_combine_fragments_with_merge_dim(nt, nvars):
     """Test if sub-fragments split from datasets with merge dims can be combined with each other."""
 
-    t_offset = var_offset = 0
+    t_offset = 0
     target_chunks = {"time": 1}
 
-    # create two datasets, each with with `nt` time steps, and two variables
-    ds0 = make_ds(nt=nt)
-    ds1 = make_ds(nt=nt)
+    # create two datasets, each with with `nt` time steps, and one variable
+    ds0 = make_ds(nt=nt).drop_vars("foo")
+    assert "foo" not in ds0.data_vars
+    ds1 = make_ds(nt=nt).drop_vars("bar")
+    assert "bar" not in ds1.data_vars
 
-    # create an index with variable merge dim and time concat dim. mocks IndexItems transform.
-    index = Index(
-        {
-            Dimension("time", CombineOp.CONCAT): IndexedPosition(t_offset, dimsize=nt),
-            Dimension("variable", CombineOp.MERGE): IndexedPosition(var_offset, dimsize=nvars),
-        }
-    )
+    # replicates indexes created by IndexItems transform.
+    index0, index1 = [
+        Index(
+            {
+                Dimension("variable", CombineOp.MERGE): Position(i),
+                Dimension("time", CombineOp.CONCAT): IndexedPosition(t_offset, dimsize=nt),
+            }
+        )
+        for i in range(2)
+    ]
     # split the two (mock indexed) datasets into sub-fragments. once split according to
     # `target_chunks = {"time": 1}`, each list of splits should contain 2 sub-fragments
     ds0_splits, ds1_splits = [
-        list(split_fragment((index, ds), target_chunks=target_chunks)) for ds in (ds0, ds1)
+        list(split_fragment((index, ds), target_chunks=target_chunks))
+        for index, ds in zip([index0, index1], [ds0, ds1])
     ]
     assert len(ds0_splits) == len(ds1_splits) == 2
 
@@ -50,10 +56,12 @@ def test_split_and_combine_fragments_with_merge_dim(nt, nvars):
     # in the `Rechunk` transform.
     for i in range(len(ds0_subfragments)):
         assert ds0_subfragments[i].groupkey == ds1_subfragments[i].groupkey
-        combine_fragments(
+        _, ds_combined = combine_fragments(
             ds0_subfragments[i].groupkey,
             [ds0_subfragments[i].subfragment, ds1_subfragments[i].subfragment],
         )
+        assert "foo" in ds_combined.data_vars
+        assert "bar" in ds_combined.data_vars
 
 
 @pytest.mark.parametrize("offset", [0, 5])  # hypothetical offset of this fragment
--- a/tests/test_rechunking.py
+++ b/tests/test_rechunking.py
@@ -14,17 +14,17 @@ from .data_generation import make_ds
 
 
 @pytest.mark.parametrize(
-    "nt_resample",
-    [(2, "1D"), (5, "1D"), (10, "2D")],
+    "nt_dayparam",
+    [(5, "1D"), (10, "2D")],
 )
-@pytest.mark.parametrize("time_chunks", [1, 2])
-def test_split_and_combine_fragments_with_merge_dim(nt_resample, time_chunks):
+@pytest.mark.parametrize("time_chunks", [1, 2, 5])
+def test_split_and_combine_fragments_with_merge_dim(nt_dayparam, time_chunks):
     """Test if sub-fragments split from datasets with merge dims can be combined with each other."""
 
     target_chunks = {"time": time_chunks}
-    nt, resample = nt_resample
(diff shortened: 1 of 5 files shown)


cisaacstern:
@norlandrhagen thanks for getting this started.

Would it be okay if I take a shot at finishing this PR?

norlandrhagen:
@cisaacstern that would be great!

cisaacstern:
As of the last commit, the test added here now replicates the bug reported in #517:

```console
pytest -vx tests/test_rechunking.py -k test_split_and_combine_fragments_with_merge_dim
```

```
        ...
        total_size = functools.reduce(operator.mul, shape)
        if len(fragments) != total_size:
            # this error path is currently untested
>           raise ValueError(
                "Cannot combine fragments. "
                f"Expected a hypercube of shape {shape} but got {len(fragments)} fragments."
            )
E           ValueError: Cannot combine fragments. Expected a hypercube of shape [1] but got 2 fragments.
```

Thanks again to @norlandrhagen for kicking this off, and @rabernat for pairing on this yesterday. 🙏 

Now that the test is in place, I'll work on fixing the bug!

cisaacstern:
Summary to date:

- I believe the simplest way to make `combine_fragments` capable of merge (as well as concat), is to keep the existing concat logic unchanged, and to simply _first_ merge any fragments which require merging, and then pass those `merged_fragments` downstream to the existing concat logic. The work here in `rechunking.py` reflects this approach.
- I've parametrized the xarray -> zarr end-to-end test with the multivariable file patterns. Before the current changes to `rechunking.py` were added, this raised the same error as reported in #517. Following addition of these changes, all of the multivariable end-to-end tests pass. 🎉 
- All except one of the parameterizations of the unit test also pass. The failing case in the unit tests is for (`nt=10`, `resample="2D"`, `time_chunks=2`). It fails on the check of equality between `sizes` and `expected_sizes`... I suspect this may have to do with the way I am mocking the `IndexedPositions` emitted by the `IndexItems` transform in the unit test, because AFAICT this combination of parameters is also covered (and passes) in the end-to-end testing. Going to keep digging a bit on this.

cisaacstern:
> I suspect this may have to do with the way I am mocking the IndexedPositions emitted by the IndexItems transform in the unit test

Correction... the IndexedPositions look correct, but the subfragments generated by `split_fragment` in the unit test are a bit suspicious, being of len 14, with group keys:
```
[(('time', 0),), (('time', 0),), (('time', 0),), (('time', 0),), (('time', 1),), (('time', 1),), (('time', 1),), (('time', 1),), (('time', 1),), (('time', 1),), (('time', 2),), (('time', 2),), (('time', 2),), (('time', 2),)]
```
Note that `(('time', 1),)` appears 6 times... which is confusing to me. Getting closer to the issue here I think... 🤔

norlandrhagen:
Super exciting progress @cisaacstern !

cisaacstern:
I believe this is now quite close to the intended behavior. The relevant integration tests appear to all be passing, and most of the unit tests are as well. There are two unit test cases which are failing, I just need to figure out if that's a problem with some assumption in the unit test itself, of if I'm actually catching a meaningful corner case.

rabernat:
Great progress Charles! Let me know if I can be helpful here.

cisaacstern:
Thanks so much to @jbusecke for pairing on this, which helped me understand the specific failure mode of the failing test cases (which do appear to be a meaningful bug, and not a testing mistake...):

- All failing cases involve the parametrization `nt_dayparam=(10, "2D")`, wherein an initial aggregate dataset of 10 daily time steps and 2 variables is split into a collection of 2-day, single variable datasets, with a total collection size of
    ```
    (10 / 2 time steps) * (2 variables) = 10 datasets
    ```


- Each of the 10 datasets is of length 2 in the time dimension. Therefore, in the case of the `target_chunks={"time": 1}` parametrization, we would expect `split_fragments` to divide these 10 datasets into a total of 20 subfragments, each of length 1 in the time dimension. Indeed these subfragment datasets are emitted as expected, and they are grouped correctly according variable, but their time dimension groupkeys are incorrect. As shown in the table below, the time keys `('time', 1)`, `('time', 2)`, `('time', 3)`, and `('time', 4)` each have _**two time steps**_ grouped under them. IIUC, this is incorrect, as there should only be _one time step per time key_, for the chunking scheme `target_chunks={"time": 1}`. This would seem to be a bug in the splitting logic, which I will now try to track down.

  <details>
  <summary>Code for generating table</summary>
  
  ```python
  # this code was run from within a debugger console, opened from a breakpoint set
  # within `test_rechunking::test_split_and_combine_fragments_with_merge_dim`, which
  # was run with the `1-nt_dayparam1` parametrization (the same as described above)
  print("nfragment | len(ds.time) | date | vars | groupkey")
  print("--------- | ------------ | ---- | ---- | --------")
  for i, sf in enumerate(subfragments):
      ds = sf.content[1]
      print(
          i, "|",
          len(ds.time), "|", str(ds.time[0].values)[:10], "|",
          [k for k in ds.data_vars.keys()], "|",
          f"`{sf.groupkey}`",
      )
  ```
  </details>

  nfragment | len(ds.time) | date | vars | groupkey
  --------- | ------------ | ---- | ---- | --------
  0 | 1 | 2010-01-01 | ['bar'] | `(('time', 0), ('variable', 0))`
  1 | 1 | 2010-01-02 | ['bar'] | `(('time', 1), ('variable', 0))`
  2 | 1 | 2010-01-03 | ['bar'] | `(('time', 1), ('variable', 0))`
  3 | 1 | 2010-01-04 | ['bar'] | `(('time', 2), ('variable', 0))`
  4 | 1 | 2010-01-05 | ['bar'] | `(('time', 2), ('variable', 0))`
  5 | 1 | 2010-01-06 | ['bar'] | `(('time', 3), ('variable', 0))`
  6 | 1 | 2010-01-07 | ['bar'] | `(('time', 3), ('variable', 0))`
  7 | 1 | 2010-01-08 | ['bar'] | `(('time', 4), ('variable', 0))`
  8 | 1 | 2010-01-09 | ['bar'] | `(('time', 4), ('variable', 0))`
  9 | 1 | 2010-01-10 | ['bar'] | `(('time', 5), ('variable', 0))`
  10 | 1 | 2010-01-01 | ['foo'] | `(('time', 0), ('variable', 1))`
  11 | 1 | 2010-01-02 | ['foo'] | `(('time', 1), ('variable', 1))`
  12 | 1 | 2010-01-03 | ['foo'] | `(('time', 1), ('variable', 1))`
  13 | 1 | 2010-01-04 | ['foo'] | `(('time', 2), ('variable', 1))`
  14 | 1 | 2010-01-05 | ['foo'] | `(('time', 2), ('variable', 1))`
  15 | 1 | 2010-01-06 | ['foo'] | `(('time', 3), ('variable', 1))`
  16 | 1 | 2010-01-07 | ['foo'] | `(('time', 3), ('variable', 1))`
  17 | 1 | 2010-01-08 | ['foo'] | `(('time', 4), ('variable', 1))`
  18 | 1 | 2010-01-09 | ['foo'] | `(('time', 4), ('variable', 1))`
  19 | 1 | 2010-01-10 | ['foo'] | `(('time', 5), ('variable', 1))`

cisaacstern:
@rabernat, the test added in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/521/commits/f0d3159ee5fa09dfb383291b1f82bc9a81b1d781 reproduces the behavior documented in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/521#issuecomment-1612227064 above. To your eye, is this indeed a bug, or have I misunderstood what group keys to expect in this situation?

rabernat:
Trying to parse this example scenario, which is indeed pretty complicated. But the crux of it seems to be

> * IIUC, this is incorrect, as there should only be _one time step per time key_,

Based on my reading, this conclusion seems correct.

rabernat:
The bug in question seems purely related to the rechunking logic along the concat dim. So it's suspicious that it only arises once you bring in a merge dim. That seems like an important clue.

tests/test_rechunking.py
```diff
@@ -1,14 +1,117 @@
...
+    nt_total = 10  # the total size of the hypothetical dataset
+    time_chunks = 1
+    target_chunks = {"time": time_chunks}
+
+    nt = 2
+    ds = make_ds(nt=nt)  # this represents a single dataset fragment
+    dimension = Dimension("time", CombineOp.CONCAT)
+
+    offset = 1
+    index = Index([(dimension, IndexedPosition(offset, dimsize=nt_total))])
```
rabernat, line 92: This doesn't quite seem right to me. If each dataset has two days in it (`nt=2`), the only possible offsets are even (0, 2, 4, 6, and 8). So it feels like we are creating inconsistent test data here.

tests/test_rechunking.py
```diff
@@ -1,14 +1,117 @@
...
+    """Test if sub-fragments split from datasets with merge dims can be combined with each other."""
+
+    target_chunks = {"time": time_chunks}
+    nt, dayparam = nt_dayparam
+    ds = make_ds(nt=nt)
+    dsets, _, _ = split_up_files_by_variable_and_day(ds, dayparam)
+
+    # replicates indexes created by IndexItems transform.
+    unique_times = np.unique([ds.time[0].values for ds in dsets])
+    time_positions = {t: i for i, t in enumerate(unique_times)}
```
rabernat, line 31: This logic seems wrong. The index in `IndexedPosition` needs to be an actual offset from the beginning of the array.

cisaacstern:
@rabernat thanks so much for the review, your two inline comments were a serious a-ha moment for me. I thought I must've been misunderstanding something, and it turns out I was.

To recap (for my future self and any others reading this), there are actually _3 indexing spaces_ in play here:

- `FilePattern` index space: the arrangement of input files as defined by file pattern `CombineOp`s.
- Array index space: dataset-level indexing.
- Chunk index space: zarr chunk indexing.

I had misunderstood the `IndexedPosition.value` to be referencing an `FilePattern` index space, when it is in fact referencing array index space. Fixing this misunderstanding, per your comment, in the "possible bug test" https://github.com/pangeo-forge/pangeo-forge-recipes/pull/521/commits/b8130976777e1aa067e6e81c5ff6b0b81f6d9713 allows the test to pass (and reveals that this is not, in fact, a bug).

I'll now remove the possible bug test, and fix this issue in the unit test. 🚀

rabernat:
> I had misunderstood the `IndexedPosition.value` to be referencing an `FilePattern` index space, when it is in fact referencing array index space.

I am not super happy about how these types look, but I do believe this is documented.

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/5edf5ec438e6404f89c95ca7319e5be96e079253/pangeo_forge_recipes/types.py#L25-L30

cisaacstern:
@rabernat AFAICT this is good to go. Would love your final review, in case there's anything I've overlooked.

Two other gut checks in process:

- https://github.com/leap-stc/data-management/pull/21#discussion_r1247201859
- I'm currently running the terraclimate tutorial notebook offline, xref https://github.com/pangeo-forge/pangeo-forge-recipes/pull/490

IMHO the testing here is pretty robust, so I don't think we should wait on either of these items to merge, but whenever they both complete, they'll be great further verification of these changes.

rabernat (approved):
Nice work!

cisaacstern:
🥳

norlandrhagen:
Incredible @cisaacstern! Super exciting to have this PR merged in.

jbusecke:
Awesome @cisaacstern!

alxmrs:
So great to see the release within reach!!

PR #522, opened 2023-05-08
NASA CMR Tutorial 

norlandrhagen:
This PR intends to add a tutorial using the [pangeo-forge-cmr](https://github.com/yuvipanda/pangeo-forge-cmr) library with the `beam-refactor` branch.


cisaacstern:
@norlandrhagen thanks for opening this.

As a matter of organization, I'm not certain that we'll want to maintain tutorials that rely on external plugins as part of the docs here. That being said, this work definitely has a place as a demonstration _somewhere_, perhaps in pangeo-forge-cmr itself. Though the next thing I'll mention calls that suggestion a bit into question...

Regarding Earthdata login, I think @yuvipanda's https://github.com/yuvipanda/pangeo-forge-earthdatalogin may have what you are looking for. [See here for an example](https://github.com/pangeo-forge/GPM_3IMERGDL-feedstock/blob/beam-refactor/feedstock/recipe.py). I am not sure if/how this related to the EULA aspect.

So what we have here is a situation where a particular tutorial may rely on _multiple_ plugins. In this case arguably the plugin docs are also not the best place for maintaining the tutorial. This suggests that this type of tutorial might be best housed in https://github.com/pangeo-forge/docs. There is nothing in that repo yet, I just created it the other day as a place to aggregate docs that touch multiple modular components of Pangeo Forge.

For the time being, I'd recommend just keep working on this PR here, and get the notebook where you want it, and then we can confirm where to put it once you're happy with the content.

norlandrhagen:
Thanks for the detailed response @cisaacstern! Happy to move it to the new `docs` repo anytime.

PR #524, opened 2023-05-13
[WIP] Reference Recipes fail when Custom Transformations Applied [Tests]

norlandrhagen:
New test to document failure of custom transformations when using reference recipes.

### DIFF
--- a/tests/test_end_to_end.py
+++ b/tests/test_end_to_end.py
@@ -80,12 +80,11 @@ def test_xarray_zarr_subpath(
     xr.testing.assert_equal(ds.load(), daily_xarray_dataset)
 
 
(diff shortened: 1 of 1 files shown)


norlandrhagen:
Thinking this through a bit, this should be expected to fail as custom beam transformations act on a piece of an Xarray dataset dataset. 

In the [Kerchunk docs](https://fsspec.github.io/kerchunk/tutorial.html#preprocessing), it seems like `preprocessing` or `postprocessing` functions can be applied through `MultiZarrToZarr`.

norlandrhagen:
Ah, I see that example now. Thanks @cisaacstern!

PR #539, opened 2023-06-30
Update default branch to main in workflows

cisaacstern:
As the title indicates, this updates workflows to reflect that our default branch is now `main` (created from `beam-refactor`).

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -2,11 +2,11 @@ name: Tests
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


cisaacstern:
The only failing test is an intermittent error (which I previously saw in #521) apparently associated with the `upstream-dev` test case only. I've opened #540 to track this, but not going to hold up this PR to fix it, as it doesn't seem to be arising when using stable releases of our dependencies.

PR #546, opened 2023-07-16
WIP: Dynamic rechunking option for `StoreToZarr`

jbusecke:
This PR came out of the Saturday Sprint at Scipy (together with @amsnyder, @thodson-usgs, @alaws-USGS, @kjdoore). 

The proposed mechanism here is a generalization of my [prototype CMIP6 pangeo-forge feedstock](https://github.com/jbusecke/cmip6-leap-feedstock/blob/c83067ca2fbf7b9bafdb2f378f37905a18aa3fe7/feedstock/transforms.py#L13-L42). Over there I implemented dynamic rechunking along a single dimension according to a desired chunksize in memory (EDIT: I now realized that I could have probably achieved this with [dask auto chunking](https://docs.dask.org/en/latest/array-chunks.html#automatic-chunking) for this specific case).

The generalization was prompted by @rsignell-usgs. Rich wanted to have an option to specify the size of chunks and then have a fixed ration of total chunks between different dimensions. 

I think I have put together a solution that should work well in this case. The outline of the algo is the following: 
- Find even divisors of the dimension length for each individual dimension
- Iterate over each possible combination of chunks along the given dimensions, and compute the chunksize of a single variable for each of the combinations.
- Select viable combinations that fall within a range of sizes between [size - tolerance * size, size + tolerance * size]
- For these candidates then find the closest fit to the desired `chunking ratio`

> @alaws-USGS This is a different implementation than we originally came up with at the sprint, but I think the even division of chunks makes this quite attractive. 

I put some useage examples together [here](https://gist.github.com/jbusecke/fb546fbfd51fb764bc7ebd696b810df6).

## TODO:
- [x] Decide on how to expose this in the PGF-recipe API

I propose to integrate this functionality into `StoreToZarr`. The user could do something like this:

```python
...
StoreToZarr(..., target_chunk_size, target_chunk_ratio)
```

- [x] Enable string size input. Currently this only takes the size argument as an integer (specifying the number of bytes we want our chunks to have). I think it would be more intuitive to be able to specify size like `100MiB` etc. 

cc @cisaacstern @rabernat 

> @TomNicholas: I looked into the [dask auto chunking](https://docs.dask.org/en/latest/array-chunks.html#automatic-chunking), and I think that this covers the case where you strictly want to chunk only one dimension, but does not enable the user to specify the above ratio (they seem to aim for evenly sized chunks. But I might be overlooking something.

- [ ] Enable a fallback algorithm, that does loosen the constraint of even divisors to avoid raising an error for [certain CMIP6 models](https://github.com/leap-stc/cmip6-leap-feedstock/pull/9#issuecomment-1670207771)


thodson-usgs:
Thanks, Julius, 
BTW, @kjdoore was the one working on the implementation with you.

jbusecke:
Oh my bad for the confusion. My name buffer is super small and always overflows on conferences 😂

cisaacstern:
Nice work, @jbusecke, and everyone else named here who contributed.

> I propose to integrate this functionality into StoreToZarr

That seems reasonable to insert right above here

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/f0c7dac8a9a49533fe544dc2fd84e501f9d96508/pangeo_forge_recipes/transforms.py#L396-L398

as an alternate path for deriving `target_chunks`.

> Enable string size input.

At the risk of stating the obvious, would be great to reference whatever established parsing strategies / conventions exist in other packages for this, rather than inventing our own convention.

pangeo_forge_recipes/dynamic_target_chunks.py
```diff
@@ -0,0 +1,107 @@
...
+            divisors.append(n // i)
+    return divisors
+
+
+def dynamic_target_chunks_from_schema(
+    schema: XarraySchema,
+    target_chunk_nbytes: int,  # TODO: Accept a str like `100MB`
+    target_chunk_ratio: Dict[str, int],
+    nbytes_tolerance: float = 0.2,
+) -> dict[str, int]:
```
rabernat, line 43: I attempted to review this PR but realized I was missing some key context. Could we provide a docstring for this function which explains what this function does and what these parameters are? In particular, I don't understand `target_chunk_ratio`.

cisaacstern: FWIW, I was also confused by that term. So +1 on additional context. 🙏

pangeo_forge_recipes/dynamic_target_chunks.py
```diff
@@ -0,0 +1,107 @@
...
+    """Returns an estimate of memory size based on input chunks.
+    Currently this applies the chunks input to the dataset, then
+    iterates through the variables and returns the maximum.
+    """
+    ds_single_chunk = ds.isel({dim: slice(0, chunk) for dim, chunk in chunks.items()})
+    mem_size = max([ds_single_chunk[var].nbytes for var in ds_single_chunk.data_vars])
+    return mem_size
+
+
+def difference(a: np.ndarray, b: np.ndarray) -> np.ndarray:
```
TomNicholas, line 20: I would rename this to something like `distance` or `similarity`

pangeo_forge_recipes/dynamic_target_chunks.py
TomNicholas: Would it be slightly clearer to call it `target_chunk_aspect_ratio`?

tests/test_dynamic_target_chunks.py
```diff
@@ -0,0 +1,63 @@
+import dask.array as dsa
+import pytest
+import xarray as xr
+
+from pangeo_forge_recipes.aggregation import dataset_to_schema
+from pangeo_forge_recipes.dynamic_target_chunks import dynamic_target_chunks_from_schema
+
+
+class TestDynamicTargetChunks:
```
TomNicholas, line 9: This is potentially a good use case for hypothesis - you could parameterize a unit test with a hypothesis strategy that generates arbitrary (regular) chunks, then assert that the property that the returned target chunk size is within the specified tolerance.

TomNicholas:
> established parsing strategies / conventions exist in other packages

A function for parsing sizes expressed as strings with SI units into integers exists in both cubed and dask, and dask I think has a strategy that will try to generate chunks of a given size.

tests/test_dynamic_target_chunks.py
jbusecke: Yeah that is a good idea. I will have to dig into hypothesis a bit to understand how to encode the logic.

jbusecke:
Hey everyone, thanks for the great comments. Sorry for the patchwork PR (working in between flights), I should have marked this as draft in the meantime. 

To clarify the purpose (and naming) of `target_chunk_ratio` a bit and maybe crowdsource a better name: 
The idea here is to give the user some control over the ratio of total number of chunks along each dimension. For example if I specify `target_chunks_ratio={'x':1, 'y':10}` the algorithm should aim to get `len(ds_rechunked.chunks['x']/len(ds_rechunked.chunks['y']` as close to `1:10` as possible (within given constraints). This works (now, see most recent commit) independently of dataset size. I agree that the term `target_chunk_ratio` is confusing. 

How about something like `nchunks_aspect_ratio`? I am trying to reflect here that we are optimizing the ratio between absolute chunks, not the aspect ratio of each individual chunk.

pangeo_forge_recipes/dynamic_target_chunks.py
jbusecke: I added a docstring explaining the inputs.

jbusecke:
> A function for parsing sizes expressed as strings with SI units into integers exists in both cubed and dask, and dask I think has a strategy that will try to generate chunks of a given size.

I just implemented `dask.utils.parse_bytes` (with test). I chose the dask implementation since we already depend on dask here.

rsignell-usgs:
> How about something like nchunks_aspect_ratio? 
I think this conveys the concept well.   Thanks for working on this -- we intend to use this for the [HyTEST project](https://hytest-org.github.io/hytest/doc/About.html) as well!

tests/test_dynamic_target_chunks.py
TomNicholas: This PR to dask might give you what you need (I can also finish it if that would help)

https://github.com/dask/dask/pull/9374

cisaacstern:
> How about something like `nchunks_aspect_ratio`?

I favor `target_chunk_aspect_ratio` as proposed by @TomNicholas in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/546#discussion_r1265998042, for consistency with the `target_chunk_nbytes` kw on this function, and the `target_chunks` kw on the `StoreToZarr` transform, which feels in keeping with the spirit of ["call things the same thing"](https://macwright.com/2021/02/17/the-naming-of-things.html#call-things-the-same-thing).

Thanks for all the work on this, Julius!

cisaacstern:
> Or maybe `target_chunks_aspect_ratio` (plural "chunks")?

I think this is preferable, because the aspect ratio is between plural chunks. (As opposed to `target_chunk_nbytes` which does refer to the desired size of a single chunk.)

rsignell-usgs:
@jbusecke back to you to change https://github.com/jbusecke/pangeo-forge-recipes/blob/dynamic_chunks_2/pangeo_forge_recipes/dynamic_target_chunks.py#L42, right?

pangeo_forge_recipes/dynamic_target_chunks.py
```diff
@@ -0,0 +1,140 @@
...
+    divisors = []
+    for i in range(1, n + 1):
+        if n % i == 0:
+            divisors.append(n // i)
+    return divisors
+
+
+# FIXME: mypy did not correctly infer the return type of parse_bytes
+def parse_bytes_wrapper(target_chunk_size: Union[str, int]) -> int:
+    return parse_bytes(target_chunk_size)
```
jbusecke, line 41: If I do not wrap this with type hints, mypy is not able to detect that the result of parse_bytes is always an int and complains.

jbusecke:
Ok I have made some progress here. You can now call `StoreToZarr` either with static chunk sizes (via `target_chunks`) or with dynamic chunks sizes (by specifying both `target_chunk_size` and `target_chunks_aspect_ratio`). 

I have added lightweight [tests](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/39ff7e69612b3ee9049514c2a6ebe363279cdaa7/tests/test_transforms.py#L237-L282) to ensure that the user cannot enable static and dynamic chunking at the same time. These work but I am not sure about my actual implementation of defaults and checks in the `StoreToZarr` class.  

I am still getting some `mypy` errors that expose my limited experience with dataclasses I am afraid. 
```
pangeo_forge_recipes/transforms.py:399: error: Argument "default" to "field" has incompatible type "None"; expected "Dict[str, int]"  [arg-type]
pangeo_forge_recipes/transforms.py:401: error: Argument "default" to "field" has incompatible type "None"; expected "Dict[str, int]"  [arg-type]
pangeo_forge_recipes/transforms.py:402: error: Incompatible types in assignment (expression has type "None", variable has type "Union[str, int]")  [assignment]
Found 3 errors in 1 file (checked 3 source files)
```
I am not sure how to set a proper default here, that can then be checked in the [__post_init__](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/39ff7e69612b3ee9049514c2a6ebe363279cdaa7/pangeo_forge_recipes/transforms.py#L405-L427). 

Any help with this would be much appreciated. 

I would also love to get some advice on how to further test the behavior here. I want to refrain from adding more expensive integration tests, but it might be nice to check that the `self.target_chunks` attribute is set correctly in [this](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/39ff7e69612b3ee9049514c2a6ebe363279cdaa7/pangeo_forge_recipes/transforms.py#L433-L439) block. Any ideas on how to inspect that without writing out an actual zarr store?

pangeo_forge_recipes/transforms.py
```diff
@@ -383,16 +383,60 @@ class StoreToZarr(beam.PTransform, ZarrWriterMixin):
     :param combine_dims: The dimensions to combine
     :param target_chunks: Dictionary mapping dimension names to chunks sizes.
         If a dimension is a not named, the chunks will be inferred from the data.
+    :param target_chunk_aspect_ratio: Dictionary mapping dimension names to desired
+        aspect ratio of total number of chunks along each dimension.
+    :param target_chunk_size:
```
jbusecke, line 388: Oops need to add a description here.

pangeo_forge_recipes/dynamic_target_chunks.py
jbusecke: Maybe @TomNicholas or @cisaacstern know more here. This confused me since the type hinting on `dask.utils.parse_bytes` seemed to be equivalent (see [here](https://github.com/dask/dask/blob/a8f2acc2e3328707cda2a25cee6cfadec44ec815/dask/utils.py#L1455))

pangeo_forge_recipes/dynamic_target_chunks.py
```diff
@@ -0,0 +1,141 @@
...
+    schema: XarraySchema,
+    target_chunk_size: Union[int, str],
+    target_chunks_aspect_ratio: Dict[str, int],
+    size_tolerance: float,
+) -> dict[str, int]:
+    """Determine chunksizes based on desired chunksize (max size of any variable in the
+    dataset) and the ratio of total chunks along each dimension of the dataset. The
+    algorithm finds even divisors, and chooses possible combination that produce chunk
+    sizes close to the target. From this set of combination the chunks that most closely
+    produce the ratio of total chunks along the given dimensions.
```
cisaacstern, line 56: Not sure what the last sentence here means, maybe there's a typo?

jbusecke: Thanks for catching this. I completed the sentence in the latest commit. Does this make more sense to you now?

cisaacstern: Definitely closer! I'll make a small suggestion now.

pangeo_forge_recipes/transforms.py
```diff
@@ -383,20 +383,81 @@ class StoreToZarr(beam.PTransform, ZarrWriterMixin):
...
+        # and target_chunks_aspect_ratio are specified
+        if self.target_chunks is None and not (
+            self.target_chunk_size is not None and self.target_chunks_aspect_ratio is not None
+        ):
+            raise ValueError(
+                (
+                    "Must specify both target_chunk_size and "
+                    "target_chunks_aspect_ratio to enable dynamic chunking."
+                )
+            )
```
cisaacstern, line 436:
```suggestion
        if self.target_chunks is None and not (
            self.target_chunk_size and self.target_chunks_aspect_ratio
        ):
            raise ValueError(
                "To enable dynamic chunking, both `target_chunk_size` and "
                "`target_chunks_aspect_ratio` must be specified."
            )
```

- I think phrased this way, the ValueError message can stand in for the comment. 
- Also I think restating the conditional expression without `is not None` will work (same as above) and is more succinct.
- The extra layer of parentheses on the error message should not be necessary.

pangeo_forge_recipes/transforms.py
```diff
@@ -383,20 +383,81 @@ class StoreToZarr(beam.PTransform, ZarrWriterMixin):
...
+        # check that not both static and dynamic chunking are specified
+        if self.target_chunks is not None and (
+            self.target_chunk_size is not None or self.target_chunks_aspect_ratio is not None
+        ):
+            raise ValueError(
+                (
+                    "Cannot specify both target_chunks and "
+                    "target_chunk_size or target_chunks_aspect_ratio."
+                )
+            )
```
cisaacstern, line 424:
```suggestion
        if self.target_chunks and (self.target_chunk_size or self.target_chunks_aspect_ratio):
            raise ValueError(
                "If `target_chunks` is given, both of the dynamic chunking options "
                "(`target_chunk_size` and `target_chunks_aspect_ratio`) must be None."
            )
```

- I don't think we need a comment and an error message here. I've incorporated some of the comment language into the message.
- I'm pretty sure if any of these kwargs are `not None` then they will evaluate to `True` without the `is not None`... which is more succinct.
- We shouldn't need the extra layer of parentheses.

pangeo_forge_recipes/transforms.py
```diff
@@ -383,20 +383,81 @@ class StoreToZarr(beam.PTransform, ZarrWriterMixin):
...
     combine_dims: List[Dimension]
-    target_chunks: Dict[str, int] = field(default_factory=dict)
+    target_chunks: Optional[Dict[str, int]] = field(default=None)
+    target_chunks_aspect_ratio: Optional[Dict[str, int]] = field(default=None)
+    target_chunk_size: Optional[Union[str, int]] = field(
+        default=None
+    )  # ? Should we provide a default?
+    size_tolerance: float = 0.2
+
+    def __post_init__(self):
```
cisaacstern, line 406: Nice! Looks like this out first PTransform with a __post_init__, this is definitely a good place for this type of kw validation.

cisaacstern (commented):
Thanks for all this hard work, @jbusecke. I've made some stylistic suggestions as well as made comments in the tests inquiring re: motivation for class-based tests (which are not used elsewhere in `pangeo-forge-recipes`).

I'll confess the actual logic of `dynamic_target_chunks_from_schema` is still a bit opaque to me. I appreciate the detailed docstring and comments, but the details are still a bit hard for me to grok, and therefore I haven't commented on them specifically. One thought I had was, I wonder if this might be a situation where a few simple doctest examples in the docstring for that function could help developers more readily grasp what to expect from that function. Don't want to overcomplicate things but just trying to think of ways to make this more readily understandable for people like me who like to learn by example. (While the tests provide examples, with parametrization they are not necessarily readable at-a-glance the way a doctest example might be?.)

Also, I believe we've previously discussed the possibility of a constraint on this type of logic to make chunking divisions across logical dataset thresholds ... for example, a dataset with 1-hour time resolution might make sense to split into 24hr daily chunks, but the logic here (which isn't aware of that constraint), could end up splitting it into 27hr chunks or something, which could be awkward? Not sure it's essential to include this in this PR, just thinking out loud and wondering if your SciPy discussion touched on this.

pangeo_forge_recipes/dynamic_target_chunks.py
cisaacstern: Getting closer! Added a few comments in the review below.

jbusecke:
Just noting that I am testing this branch as part of my CMIP6 work, and I noticed that I need to make some more adjustments (see [here](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/546/commits/a5a182a643c8735b64c923c28b235b9d52f83db1)) which were not caught by the tests. To me this indicates that we need at least one end to end test for this functionality.

pangeo_forge_recipes/transforms.py
```diff
@@ -383,20 +383,86 @@ class StoreToZarr(beam.PTransform, ZarrWriterMixin):
...
     def expand(self, datasets: beam.PCollection) -> beam.PCollection:
         schema = datasets | DetermineSchema(combine_dims=self.combine_dims)
-        indexed_datasets = datasets | IndexItems(schema=schema)
-        rechunked_datasets = indexed_datasets | Rechunk(
-            target_chunks=self.target_chunks, schema=schema
+        target_chunks = beam.pvalue.AsSingelton(
+            schema
+            | beam.Map(
+                self.determine_target_chunks
+            )  # Would beam.pvalue.AsSingleton(self.determine_target_chunks(schema)) work?
```
cisaacstern, line 461: Or

```python
target_chunks = self.determine_target_chunks(beam.pvalue.AsSingleton(schema))
```

?

jbusecke:
Quickly responding to

> Also, I believe we've previously discussed the possibility of a constraint on this type of logic to make chunking divisions across logical dataset thresholds ... for example, a dataset with 1-hour time resolution might make sense to split into 24hr daily chunks, but the logic here (which isn't aware of that constraint), could end up splitting it into 27hr chunks or something, which could be awkward? Not sure it's essential to include this in this PR, just thinking out loud and wondering if your SciPy discussion touched on this.

The algo for now will only allow even divisions. I see this as partially fulfilling the requirements you mention. If the dataset e.g. has multiple "full" years (with no leap years) this logic will naturally arrive at some divisor that evenly splits a year or contains one or multiple full years per chunk. Dealing with more complex calendars will require us to allow uneven chunks. While this would be really cool to implement, I think it should be a completely different undertaking/PR.

pangeo_forge_recipes/transforms.py
jbusecke: Oh right, that makes a lot more sense. Changed.

jbusecke: Oh wait getting [this](https://github.com/leap-stc/cmip6-leap-feedstock/actions/runs/5744818073/job/15571779448?pr=4#step:5:46) over at the CMIP6 recipes.

pangeo_forge_recipes/dynamic_target_chunks.py
```diff
@@ -0,0 +1,141 @@
...
+    schema: XarraySchema,
+    target_chunk_size: Union[int, str],
+    target_chunks_aspect_ratio: Dict[str, int],
+    size_tolerance: float,
+) -> dict[str, int]:
+    """Determine chunksizes based on desired chunksize (max size of any variable in the
+    dataset) and the ratio of total chunks along each dimension of the dataset. The
+    algorithm finds even divisors, and chooses possible combination that produce chunk
+    sizes close to the target. From this set of combination the chunks that most closely
+    produce the aspect ratio of total chunks along the given dimensions is returned.
```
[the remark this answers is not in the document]
jbusecke: Thats clearer indeed. What I determine as max size is only one of many potential options which differ mostly if you have very differently shaped data_variables. But I think the choice here is a reasonable start and we can leave more options to the future.

tests/test_transforms.py
```diff
@@ -231,3 +233,79 @@ def _check_chunks(actual):
         indexed_datasets = datasets | IndexItems(schema=schema)
         rechunked = indexed_datasets | Rechunk(target_chunks=target_chunks, schema=schema)
         assert_that(rechunked, correct_chunks())
+
+
+class TestStoreToZarr:
```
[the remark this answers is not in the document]
jbusecke: I just did this as a convenience during local testing, but happy to remove.

pangeo_forge_recipes/transforms.py
jbusecke: I now settled on `target_chunks = beam.pvalue.AsSingleton(schema | beam.Map(self._get_target_chunks))` which seems to work, while just applying `AsSingelton` to the schema did not.

jbusecke:
First of all many thanks @cisaacstern for the thorough review. I believe I have addressed all of your concerns. If not please let me know. 
I also implemented some more input options to help with my CMIP work. 
Previously we errored out if `target_chunks_aspect_ratio` did not include all datasets dimensions. I have now changed this to assume a value (set by `default_ratio` input) for all missing dimensions. If left as default this will not chunk any of the missing dimensions, which I think is a reasonable behavior. 

Additionally I implemented a flag (`allow_extra_dims`) to be able to pass extra dimensions in `target_chunks_aspect_ratio` (the default is to error on this). This enables me to do something like this`target_chunks_aspect_ratio = {'time':10, 'x':1, 'lon':1}` and the returned target chunks for a dataset with dimensions ['time', 'x'] would be `{'time': some_value, 'x': another_value}` while for a dataset with dimensions ['time', 'lon'] would be `{'time': some_value, 'lon': another_value}`.
This will help tremendously with building a universal recipe for CMIP6 datasets with varying naming schemes.

rsignell-usgs:
@jbusecke and @cisaacstern , where do we stand with this?

jbusecke:
I have just added some logic to implement a fallback algorithm. This one is a lot more naive. It basically just determines the biggest chunk possible by dividing the length of each dimension by the corresponding value of `target_chunks_aspect_ratio`. Then I scale this maximum chunk by dividing it by increasing integers, until we find the closest match to the desired size. 
This can end up with really weird chunking schemes if applied in multiple dimensions, but should work fine with cases where we only chunk along a single dimension (https://github.com/leap-stc/cmip6-leap-feedstock/pull/9).
Ill give that a go over there, and report back. 

We still need to think about a way to check this functionality with an end-to-end test. @cisaacstern do you have any suggestions for this in general?

cisaacstern:
> We still need to think about a way to check this functionality with an end-to-end test. @cisaacstern do you have any suggestions for this in general?

@jbusecke can you remind me what about this is non-unit-testable?

jbusecke:
Just submitted the first batch with 100+ iids 🚀. 
A lot of successes, but there are quite a few oneas erroring out with this error: 
<details>
raceback (most recent call last): File "apache_beam/runners/common.py", line 1417, in apache_beam.runners.common.DoFnRunner.process File "apache_beam/runners/common.py", line 623, in apache_beam.runners.common.SimpleInvoker.invoke_process File "apache_beam/runners/common.py", line 1581, in apache_beam.runners.common._OutputHandler.handle_process_outputs File "apache_beam/runners/common.py", line 1694, in apache_beam.runners.common._OutputHandler._write_value_to_tag File "apache_beam/runners/worker/operations.py", line 239, in apache_beam.runners.worker.operations.SingletonElementConsumerSet.receive File "apache_beam/runners/worker/operations.py", line 198, in apache_beam.runners.worker.operations.ConsumerSet.update_counters_start File "apache_beam/runners/worker/opcounters.py", line 215, in apache_beam.runners.worker.opcounters.OperationCounters.update_from File "apache_beam/runners/worker/opcounters.py", line 267, in apache_beam.runners.worker.opcounters.OperationCounters.do_sample File "apache_beam/coders/coder_impl.py", line 1471, in apache_beam.coders.coder_impl.WindowedValueCoderImpl.get_estimated_size_and_observables File "apache_beam/coders/coder_impl.py", line 1482, in apache_beam.coders.coder_impl.WindowedValueCoderImpl.get_estimated_size_and_observables File "apache_beam/coders/coder_impl.py", line 1031, in apache_beam.coders.coder_impl.AbstractComponentCoderImpl.get_estimated_size_and_observables File "apache_beam/coders/coder_impl.py", line 377, in apache_beam.coders.coder_impl.FastPrimitivesCoderImpl.get_estimated_size_and_observables File "apache_beam/coders/coder_impl.py", line 441, in apache_beam.coders.coder_impl.FastPrimitivesCoderImpl.encode_to_stream File "apache_beam/coders/coder_impl.py", line 441, in apache_beam.coders.coder_impl.FastPrimitivesCoderImpl.encode_to_stream File "apache_beam/coders/coder_impl.py", line 411, in apache_beam.coders.coder_impl.FastPrimitivesCoderImpl.encode_to_stream UnicodeEncodeError: 'utf-8' codec can't encode characters in position 61-63: surrogates not allowed During handling of the above exception, another exception occurred: Traceback (most recent call last): File "/srv/conda/envs/notebook/lib/python3.9/site-packages/apache_beam/runners/worker/sdk_worker.py", line 284, in _execute response = task() File "/srv/conda/envs/notebook/lib/python3.9/site-packages/apache_beam/runners/worker/sdk_worker.py", line 357, in <lambda> lambda: self.create_worker().do_instruction(request), request) File "/srv/conda/envs/notebook/lib/python3.9/site-packages/apache_beam/runners/worker/sdk_worker.py", line 597, in do_instruction return getattr(self, request_type)( File "/srv/conda/envs/notebook/lib/python3.9/site-packages/apache_beam/runners/worker/sdk_worker.py", line 635, in process_bundle bundle_processor.process_bundle(instruction_id)) File "/srv/conda/envs/notebook/lib/python3.9/site-packages/apache_beam/runners/worker/bundle_processor.py", line 1003, in process_bundle input_op_by_transform_id[element.transform_id].process_encoded( File "/srv/conda/envs/notebook/lib/python3.9/site-packages/apache_beam/runners/worker/bundle_processor.py", line 227, in process_encoded self.output(decoded_value) File "apache_beam/runners/worker/operations.py", line 526, in apache_beam.runners.worker.operations.Operation.output File "apache_beam/runners/worker/operations.py", line 528, in apache_beam.runners.worker.operations.Operation.output File "apache_beam/runners/worker/operations.py", line 237, in apache_beam.runners.worker.operations.SingletonElementConsumerSet.receive File "apache_beam/runners/worker/operations.py", line 240, in apache_beam.runners.worker.operations.SingletonElementConsumerSet.receive File "apache_beam/runners/worker/operations.py", line 907, in apache_beam.runners.worker.operations.DoOperation.process File "apache_beam/runners/worker/operations.py", line 908, in apache_beam.runners.worker.operations.DoOperation.process File "apache_beam/runners/common.py", line 1419, in apache_beam.runners.common.DoFnRunner.process File "apache_beam/runners/common.py", line 1491, in apache_beam.runners.common.DoFnRunner._reraise_augmented File "apache_beam/runners/common.py", line 1417, in apache_beam.runners.common.DoFnRunner.process File "apache_beam/runners/common.py", line 623, in apache_beam.runners.common.SimpleInvoker.invoke_process File "apache_beam/runners/common.py", line 1581, in apache_beam.runners.common._OutputHandler.handle_process_outputs File "apache_beam/runners/common.py", line 1694, in apache_beam.runners.common._OutputHandler._write_value_to_tag File "apache_beam/runners/worker/operations.py", line 240, in apache_beam.runners.worker.operations.SingletonElementConsumerSet.receive File "apache_beam/runners/worker/operations.py", line 907, in apache_beam.runners.worker.operations.DoOperation.process File "apache_beam/runners/worker/operations.py", line 908, in apache_beam.runners.worker.operations.DoOperation.process File "apache_beam/runners/common.py", line 1419, in apache_beam.runners.common.DoFnRunner.process File "apache_beam/runners/common.py", line 1491, in apache_beam.runners.common.DoFnRunner._reraise_augmented File "apache_beam/runners/common.py", line 1417, in apache_beam.runners.common.DoFnRunner.process File "apache_beam/runners/common.py", line 623, in apache_beam.runners.common.SimpleInvoker.invoke_process File "apache_beam/runners/common.py", line 1581, in apache_beam.runners.common._OutputHandler.handle_process_outputs File "apache_beam/runners/common.py", line 1694, in apache_beam.runners.common._OutputHandler._write_value_to_tag File "apache_beam/runners/worker/operations.py", line 240, in apache_beam.runners.worker.operations.SingletonElementConsumerSet.receive File "apache_beam/runners/worker/operations.py", line 907, in apache_beam.runners.worker.operations.DoOperation.process File "apache_beam/runners/worker/operations.py", line 908, in apache_beam.runners.worker.operations.DoOperation.process File "apache_beam/runners/common.py", line 1419, in apache_beam.runners.common.DoFnRunner.process File "apache_beam/runners/common.py", line 1491, in apache_beam.runners.common.DoFnRunner._reraise_augmented File "apache_beam/runners/common.py", line 1417, in apache_beam.runners.common.DoFnRunner.process File "apache_beam/runners/common.py", line 623, in apache_beam.runners.common.SimpleInvoker.invoke_process File "apache_beam/runners/common.py", line 1581, in apache_beam.runners.common._OutputHandler.handle_process_outputs File "apache_beam/runners/common.py", line 1694, in apache_beam.runners.common._OutputHandler._write_value_to_tag File "apache_beam/runners/worker/operations.py", line 240, in apache_beam.runners.worker.operations.SingletonElementConsumerSet.receive File "apache_beam/runners/worker/operations.py", line 907, in apache_beam.runners.worker.operations.DoOperation.process File "apache_beam/runners/worker/operations.py", line 908, in apache_beam.runners.worker.operations.DoOperation.process File "apache_beam/runners/common.py", line 1419, in apache_beam.runners.common.DoFnRunner.process File "apache_beam/runners/common.py", line 1491, in apache_beam.runners.common.DoFnRunner._reraise_augmented File "apache_beam/runners/common.py", line 1417, in apache_beam.runners.common.DoFnRunner.process File "apache_beam/runners/common.py", line 837, in apache_beam.runners.common.PerWindowInvoker.invoke_process File "apache_beam/runners/common.py", line 981, in apache_beam.runners.common.PerWindowInvoker._invoke_process_per_window File "apache_beam/runners/common.py", line 1581, in apache_beam.runners.common._OutputHandler.handle_process_outputs File "apache_beam/runners/common.py", line 1694, in apache_beam.runners.common._OutputHandler._write_value_to_tag File "apache_beam/runners/worker/operations.py", line 240, in apache_beam.runners.worker.operations.SingletonElementConsumerSet.receive File "apache_beam/runners/worker/operations.py", line 907, in apache_beam.runners.worker.operations.DoOperation.process File "apache_beam/runners/worker/operations.py", line 908, in apache_beam.runners.worker.operations.DoOperation.process File "apache_beam/runners/common.py", line 1419, in apache_beam.runners.common.DoFnRunner.process File "apache_beam/runners/common.py", line 1491, in apache_beam.runners.common.DoFnRunner._reraise_augmented File "apache_beam/runners/common.py", line 1417, in apache_beam.runners.common.DoFnRunner.process File "apache_beam/runners/common.py", line 837, in apache_beam.runners.common.PerWindowInvoker.invoke_process File "apache_beam/runners/common.py", line 981, in apache_beam.runners.common.PerWindowInvoker._invoke_process_per_window File "apache_beam/runners/common.py", line 1581, in apache_beam.runners.common._OutputHandler.handle_process_outputs File "apache_beam/runners/common.py", line 1694, in apache_beam.runners.common._OutputHandler._write_value_to_tag File "apache_beam/runners/worker/operations.py", line 240, in apache_beam.runners.worker.operations.SingletonElementConsumerSet.receive File "apache_beam/runners/worker/operations.py", line 907, in apache_beam.runners.worker.operations.DoOperation.process File "apache_beam/runners/worker/operations.py", line 908, in apache_beam.runners.worker.operations.DoOperation.process File "apache_beam/runners/common.py", line 1419, in apache_beam.runners.common.DoFnRunner.process File "apache_beam/runners/common.py", line 1491, in apache_beam.runners.common.DoFnRunner._reraise_augmented File "apache_beam/runners/common.py", line 1417, in apache_beam.runners.common.DoFnRunner.process File "apache_beam/runners/common.py", line 623, in apache_beam.runners.common.SimpleInvoker.invoke_process File "apache_beam/runners/common.py", line 1581, in apache_beam.runners.common._OutputHandler.handle_process_outputs File "apache_beam/runners/common.py", line 1694, in apache_beam.runners.common._OutputHandler._write_value_to_tag File "apache_beam/runners/worker/operations.py", line 324, in apache_beam.runners.worker.operations.GeneralPurposeConsumerSet.receive File "apache_beam/runners/worker/operations.py", line 905, in apache_beam.runners.worker.operations.DoOperation.process File "apache_beam/runners/worker/operations.py", line 907, in apache_beam.runners.worker.operations.DoOperation.process File "apache_beam/runners/worker/operations.py", line 908, in apache_beam.runners.worker.operations.DoOperation.process File "apache_beam/runners/common.py", line 1419, in apache_beam.runners.common.DoFnRunner.process File "apache_beam/runners/common.py", line 1507, in apache_beam.runners.common.DoFnRunner._reraise_augmented File "apache_beam/runners/common.py", line 1417, in apache_beam.runners.common.DoFnRunner.process File "apache_beam/runners/common.py", line 623, in apache_beam.runners.common.SimpleInvoker.invoke_process File "apache_beam/runners/common.py", line 1581, in apache_beam.runners.common._OutputHandler.handle_process_outputs File "apache_beam/runners/common.py", line 1694, in apache_beam.runners.common._OutputHandler._write_value_to_tag File "apache_beam/runners/worker/operations.py", line 239, in apache_beam.runners.worker.operations.SingletonElementConsumerSet.receive File "apache_beam/runners/worker/operations.py", line 198, in apache_beam.runners.worker.operations.ConsumerSet.update_counters_start File "apache_beam/runners/worker/opcounters.py", line 215, in apache_beam.runners.worker.opcounters.OperationCounters.update_from File "apache_beam/runners/worker/opcounters.py", line 267, in apache_beam.runners.worker.opcounters.OperationCounters.do_sample File "apache_beam/coders/coder_impl.py", line 1471, in apache_beam.coders.coder_impl.WindowedValueCoderImpl.get_estimated_size_and_observables File "apache_beam/coders/coder_impl.py", line 1482, in apache_beam.coders.coder_impl.WindowedValueCoderImpl.get_estimated_size_and_observables File "apache_beam/coders/coder_impl.py", line 1031, in apache_beam.coders.coder_impl.AbstractComponentCoderImpl.get_estimated_size_and_observables File "apache_beam/coders/coder_impl.py", line 377, in apache_beam.coders.coder_impl.FastPrimitivesCoderImpl.get_estimated_size_and_observables File "apache_beam/coders/coder_impl.py", line 441, in apache_beam.coders.coder_impl.FastPrimitivesCoderImpl.encode_to_stream File "apache_beam/coders/coder_impl.py", line 441, in apache_beam.coders.coder_impl.FastPrimitivesCoderImpl.encode_to_stream File "apache_beam/coders/coder_impl.py", line 411, in apache_beam.coders.coder_impl.FastPrimitivesCoderImpl.encode_to_stream UnicodeEncodeError: 'utf-8 [while running 'Create|OpenURLWithFSSpec|OpenWithXarray|Preprocessor|StoreToZarr/StoreToZarr/DetermineSchema/Map(wrapper)-ptransform-84']' codec can't encode characters in position 61-63: surrogates not allowed
</details>

I have never seen this before. @cisaacstern do you have a clue what this means

jbusecke:
There is another distinct set of errors: 
image

Again I dont really know how to proceed...

jbusecke:
Some more observations to note:
- The submission of the recipes takes very long. Submission of ~80 iids took 20+ minutes. I wonder if there is a way to optimize this on the pangeo-forge-runner side?
- I am wondering how to efficiently batch a request like https://github.com/Timh37/CMIP6cex/issues/2. That is over 1000 iids. I probably need to split this into modules of ~100 iids?

jbusecke:
Basically all the tests that I currently wrote are for the logic in `pangeo_forge_recipes/dynamic_target_chunks.py`, but nothing touches the dynamic chunking logic in `transforms.py/StoreToZarr`! 
I am not suggesting that there isnt a way to unit-test this, I merely do not know how one would go about it. I did not find any example of using `StoreToZarr` in the tests (outside of the end-to-end tests).  The [current transform.py tests](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/main/tests/test_transforms.py) do not even import `StoreToZarr`.
I 100% understand the desire not to add more parametrizations to the end-to-end tests, but do you maybe have some guidance on how to make a 'synthetic' test that actually writes out a chunked store?

jbusecke:
> @jbusecke and @cisaacstern , where do we stand with this?

@rsignell-usgs, this PR is useable (I am running CMIP6 rechunking jobs with it right now), but I believe it needs some more testing and docs before it can get merged. 
I would love for you to give this a try on your end and see if the results are satisfying. You can see an implementation [here](https://github.com/leap-stc/cmip6-leap-feedstock/pull/9/files). The only specific thing to include to run from this PR is a [requirements.txt](https://github.com/leap-stc/cmip6-leap-feedstock/blob/dict_obj_refactor/feedstock/requirements.txt). NOTE: It was vital to include the `#egg=pangeo_forge_recipes` suffix to make this work on dataflow!

cisaacstern:
Per conversion with @jbusecke, we believe the best path to releasing this work is:

- [ ] Merge @yuvipanda's efforts in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/566 and https://github.com/pangeo-forge/pangeo-forge-runner/pull/87 which define a plugin system for `-runner` deploy-time injections into `-recipes`. (This is on me to review + merge, and is currently waiting on #563, bc I'd like to integration test all of this before it goes in.)
- [ ] Define a few new pluggable kwargs for `StoreToZarr`, something like:
    ```python
    class StoreToZarr(beam.PTransform):
        """
        
        :param dynamic_chunking_fn: A callable which _must_ take `schema` as first arg.
        """
        ...
        dynamic_chunking_fn: Optional[Callable] = None
        dynamic_chunking_fn_kwargs: Optional[dict] = field(default_factory=dict)

        def expand(...):
            schema = ...
            target_chunks = (
                dynamic_chunking_fn(schema, **dynamic_chunking_fn_kwargs)
                if self.dynamic_chunking_fn
                else self.target_chunks
            )          
            ...
    ```
- [ ] Move the work in this PR into a separate plugin package

This design allows users to bring their own dynamic chunking algorithm, and also keeps the rather large amount of code in this implementation out of the core `pangeo-forge-recipes` maintenance sphere.

yuvipanda:
Just wanted to confirm that this does not depend on my dynamic injections work :)

jbusecke:
Closing this in favor of #595. I have refactored the logic implemented here to [dynamic-chunks](https://github.com/jbusecke/dynamic_chunks) and am working on an implementation of the CMIP workflow [here](https://github.com/leap-stc/cmip6-leap-feedstock/pull/62)

PR #548, opened 2023-07-25
In combine_fragments, cast `fragments` to list if improperly decoded

cisaacstern:
Some additional logging to aid in debugging Dataflow-specific error reported in https://github.com/leap-stc/data-management/pull/33#issuecomment-1648522152

### DIFF
--- a/pangeo_forge_recipes/rechunking.py
+++ b/pangeo_forge_recipes/rechunking.py
@@ -162,6 +162,7 @@ def combine_fragments(
     :param fragments: indexed dataset fragments
     """
     if not isinstance(fragments, list):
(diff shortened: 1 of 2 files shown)


cisaacstern:
Thanks to @alxmrs for pointing out that Dataflow / Cloud Logging may not be able to materialize `logger.debug` calls, which appears to be the case, and that using `logger.info` calls would fix this issue (of not being able to see the logs on GCP). 🙃

cisaacstern:
Going to merge this to unblock https://github.com/leap-stc/data-management/pull/33 (so we don't have to pull from a PR branch of `-recipes` there).

IMO this is a temporary patch for #552 which should be reverted if/when a better fix for #552 is implemented.

PR #550, opened 2023-07-26
Attempt fixing typing issues raised with `runtime_type_check=True`

cisaacstern:
Thanks to @alxmrs, I now understand that the issue reported in https://github.com/leap-stc/data-management/pull/33#issuecomment-1648522152 (and discussed further in that thread), is almost definitely being caused by imprecise type hints ➡️ inaccurate beam coders ➡️ loss of type safety on decoding serialized PCollection values on Dataflow. The beam docs link Alex provided in https://github.com/leap-stc/data-management/pull/33#issuecomment-1650983722 is especially enlightening on this subject:

> https://beam.apache.org/documentation/sdks/python-type-safety/

Regarding the error in the linked issue, the comment here

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/f0c7dac8a9a49533fe544dc2fd84e501f9d96508/pangeo_forge_recipes/rechunking.py#L148-L150

is the best clue I've found so far, given that the error relates specifically to improperly decoded PCollection values passed to `combine_fragments`.

In this branch, I'm attempting to solve this particular issue, while also paying down a bit of typing debt more broadly, by setting `runtime_type_check=True` in the end-to-end tests, and seeing if I can resolve the issues Beam's type checker raises with that flag set.

### DIFF
--- a/pangeo_forge_recipes/openers.py
+++ b/pangeo_forge_recipes/openers.py
@@ -3,7 +3,7 @@
 import io
 import tempfile
 import warnings
-from typing import Dict, Optional, Union
+from typing import Dict, List, Optional, Union
 from urllib.parse import urlparse
 
 import xarray as xr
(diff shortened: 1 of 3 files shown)


cisaacstern:
This PR was an attempt to see if beam's type checker would catch anything relevant to #552. My conclusion is that it did not, and also that the changes here don't do much to improve things, as they are largely focused on resolving disagreements between mypy and beam. So closing now.

Issue #552, opened 2023-07-27
On dataflow, `combine_fragments` arg `fragments` is intermittently decoded incorrectly


cisaacstern:
As reported in https://github.com/leap-stc/data-management/pull/33#issuecomment-1648522152 (and discussed further in that thread), under certain circumstances on Dataflow, the values of the PCollection items passed from `GroupByKey` -> `MapTuple(combine_fragments)` here 

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/f0c7dac8a9a49533fe544dc2fd84e501f9d96508/pangeo_forge_recipes/transforms.py#L322-L323

are not properly decoded into the correct Python type when they arrive in the body of `combine_fragments`. Specifically, the `fragments` argument is type-hinted as a `List` in the signature of `combine_fragments`, but it sometimes arrives as a [`_ConcatSequence`](https://github.com/apache/beam/blob/c3f1675019254ba88ce996f3d946b3f973a13282/sdks/python/apache_beam/coders/coder_impl.py#L1087). As noted in the PR linked above, it's unclear to me what circumstance causes this.

Discussion with @alxmrs elucidated the category of error as being related to Beam's coders, which generally are inferred from type hints, and handle serialization/deserialization of PCollection values for handoff between pipeline stages in distributed contexts (e.g., Dataflow), as [described in this doc](https://beam.apache.org/documentation/sdks/python-type-safety/).

The commit history of https://github.com/pangeo-forge/pangeo-forge-recipes/pull/548 records various attempts I made to get the `fragments` coder to be correctly inferred (by trying a variety of different type-hinting strategies). Ultimately, I could not get type-hinting in any form (`.with_input_types(...)`, `PCollection[...]`, etc.) to fix this issue on Dataflow.

In its current form, #548 does solve this issue by simply explicitly casting `fragments` to `list` if it's found to be improperly decoded. This unblocks https://github.com/leap-stc/data-management/pull/33, but doesn't seem like a great solution, insofar as it pushes deserialization concerns into the body of `combine_fragments`, which is the wrong place for that.

Reading between the lines of https://github.com/apache/beam/pull/9344#issuecomment-525098325, I am left wondering if an alternative to #548 might be to `return` from `split_fragment`'s iteration, rather than `yield`ing...

[...]

PR #553, opened 2023-07-27
Return list from `split_fragment`

cisaacstern:
A possible alternative to #548 as a patch for #552. Even if this works, it might not be a good idea.

Going to test it out on https://github.com/leap-stc/data-management/pull/33 now.

### DIFF
--- a/pangeo_forge_recipes/rechunking.py
+++ b/pangeo_forge_recipes/rechunking.py
@@ -1,7 +1,7 @@
 import functools
 import itertools
 import operator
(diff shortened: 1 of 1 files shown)


cisaacstern:
As tested from CI triggered by this commit: https://github.com/leap-stc/data-management/pull/33/commits/7fa84ad1ee633b6374a84f660be6ef9daafaf2d8, this is _not_ a fix for the decoding problem. It does not throw the same error as reported in #552, but rather just hangs infinitely on the `Rechunk` transform.

PR #557, opened 2023-08-05
Concurrency limiting

### LINKED ISSUE #45: limit concurrency for input downloads
NOAA NCEI might not like it if we fire off hundreds of simultaneous requests to their servers. We would like to limit the concurrency of this step if possible. 

From an API perspective, the question is:
- Should a user have to specify concurrency limits as part of the recipe?
- Alternatively, should we try to auto-detect if flows access certain resources (e.g. specific FTP servers) and then automatically enforce concurrency limits?

In terms of implementation, Prefect cloud has a prefect solution: https://docs.prefect.io/orchestration/concepts/task-concurrency-limiting.html

However this _only works with cloud_. Some questions about this option are:

- Are we okay with getting locked into a prefect cloud feature?
- What convention do we use for the task tags to indicate concurrency? For example, I could imagine a tag like `www.ncei.noaa.gov`, allowing us to limit concurrency for all requests to that server from all flows simultaneously! That would be pretty useful.

If we don't want to get locked into cloud features, @jcrist made the following suggestion on the Prefect slack:

> I'd handle this with a distributed.Semaphore within your tasks for now. Alternatively, you could make use of dask's worker resources. Tasks tagged with tags of the form dask-resource:KEY=N will each take N amount of KEY resource. So you could limit active download tasks by creating a resource for downloading then tagging download tasks to mark that they require that resource. (edited) 
> That would mean that the total concurrency limit scales with the number of workers (so it isn't absolute across the whole run), but would also work and wouldn't block other tasks from running like the Semaphore would.

### LINKED ISSUE #389: Limit concurrency for caching
On the first production run of https://github.com/pangeo-forge/terraclimate-feedstock, Dataflow autoscaled the cluster to 1000 workers, in response to the slow throughput of caching ~882 inputs (totaling ~1.9 TB).

We should be able to limit concurrency for caching, given that the source file servers will generally be bandwidth-constrained. Dataflow provides a `max_num_workers` option to cap the size of the worker pool, but this issue is separate from that concern: concurrency should be limited _**only for the caching step**_, and then we should support larger scale-out after data is cached.

There must be a more formal discussion of this somewhere in the Beam docs, but for now the most direct discussion I've found is in the replies to https://stackoverflow.com/a/65634538, which suggest `GroupByKey` might be used to achieve this.

I believe this will require pulling caching out from [`OpenURLWithFSSpec`](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/bdb32f28d09fb3d2cb76296d988c5dcd64d9d80d/pangeo_forge_recipes/transforms.py#L75). Currently, if a `cache` argument is provided to `OpenURLWithFSSpec`, the input is cached and then immediately opened from the cache

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/bdb32f28d09fb3d2cb76296d988c5dcd64d9d80d/pangeo_forge_recipes/openers.py#L31-L32

In order to limit concurrency for the caching, but _not_ for the opening, I believe caching will need to be its own transform, the output of which is then passed to `OpenURLWithFSSpec`, which does not do any caching.

cc @rabernat @alxmrs, xref #376

cisaacstern:
WIP, will close #45 and #389 when complete.

Prioritizing this because it is blocking https://github.com/leap-stc/data-management/issues/36, and also because it's been a long standing feature request to unblock many different recipes. Implementation using `GroupByKey` adapted from example provided by @alxmrs's in linked issue, with this being the key couple lines:

> https://github.com/google/weather-tools/blob/bb2274d2be4ed51a9a6cc3bc79a8be110b250509/weather_mv/loader_pipeline/util.py#L380-L381

### DIFF
--- a/pangeo_forge_recipes/transforms.py
+++ b/pangeo_forge_recipes/transforms.py
@@ -57,7 +57,7 @@ Indexed = Tuple[Index, T]
 
 
 # TODO: replace with beam.MapTuple?
-def _add_keys(func: Callable):
+def _add_keys(func: Callable) -> Callable:
     """Convenience decorator to remove and re-add keys to items in a Map"""
     annotations = func.__annotations__.copy()
     arg_name, annotation = next(iter(annotations.items()))
@@ -75,7 +75,7 @@ def _add_keys(func: Callable):
     return wrapper
 
 
-def _add_keys_iter(func: Callable):
+def _add_keys_iter(func: Callable) -> Callable:
     """Convenience decorator to iteratively remove and re-add keys to items in a FlatMap"""
     annotations = func.__annotations__.copy()
     arg_name, annotation = next(iter(annotations.items()))
--- a/pangeo_forge_recipes/transforms.py
+++ b/pangeo_forge_recipes/transforms.py
@@ -5,13 +5,13 @@ import random
 from dataclasses import dataclass, field
 
 # from functools import wraps
-from typing import Dict, List, Optional, Tuple, TypeVar
+from typing import Dict, Iterable, Iterator, List, Optional, Tuple, TypeVar
 
 import apache_beam as beam
 
 from .aggregation import XarraySchema, dataset_to_schema, schema_to_zarr
 from .combiners import CombineMultiZarrToZarr, CombineXarraySchemas
-from .openers import open_url, open_urls, open_with_kerchunk, open_with_xarray
+from .openers import OpenFileType, open_url, open_with_kerchunk, open_with_xarray
 from .patterns import CombineOp, Dimension, FileType, Index, augment_index_with_start_stop
 from .rechunking import combine_fragments, split_fragment
 from .storage import CacheFSSpecTarget, FSSpecTarget
@@ -75,6 +75,10 @@ def _add_keys(func):
     return wrapper
 
 
+# FIXME: can `_add_keys_iter` this be made into a generic wrapper to replace `_open_urls`?
+# i.e.: `FlatMap(_add_keys_iter(open_url))` instead of `FlatMap(_open_urls)`?
+# if so, do that. if not, remove before merge
+
 # def _add_keys_iter(func: Callable):
 #     """Convenience decorator to iteratively remove and re-add keys to items in a FlatMap"""
 #     annotations = func.__annotations__.copy()
@@ -108,6 +112,24 @@ class DropKeys(beam.PTransform):
         return pcoll | "Drop Keys" >> beam.Map(_drop_keys)
 
 
+def _open_urls(
+    keyed_urls: Iterable,
+    cache: Optional[CacheFSSpecTarget] = None,
+    secrets: Optional[Dict] = None,
+    open_kwargs: Optional[Dict] = None,
+) -> Iterator[OpenFileType]:
+    """Calls `open_url` in a serial loop across an input collection of keyed urls."""
+
+    for elem in keyed_urls:
+        key, url = elem
+        opened_url = open_url(url, cache, secrets, open_kwargs)
+        yield key, opened_url
+
+
+def _assign_concurrency_group(elem, max_concurrency: int):
+    return (random.randint(0, max_concurrency - 1), elem)
+
+
 # This has side effects if using a cache
 @dataclass
 class OpenURLWithFSSpec(beam.PTransform):
@@ -123,10 +145,6 @@ class OpenURLWithFSSpec(beam.PTransform):
     secrets: Optional[dict] = None
     open_kwargs: Optional[dict] = None
     max_concurrency: Optional[int] = None
-
-    @staticmethod
-    def _assign_concurrency_group(elem, max_concurrency: int):
-        return (random.randint(0, max_concurrency - 1), elem)
 
     def expand(self, pcoll):
         if isinstance(self.cache, str):
@@ -146,10 +164,10 @@ class OpenURLWithFSSpec(beam.PTransform):
                 pcoll
                 # TODO: factor next three stages into a standalone PTransform
                 # so they can be reused elsewhere (e.g. OpenWithKerchunk)
-                | beam.Map(self._assign_concurrency_group, self.max_concurrency)
+                | beam.Map(_assign_concurrency_group, self.max_concurrency)
                 | beam.GroupByKey()
                 | DropKeys()
-                | f"Open with fsspec ({self.max_concurrency=})" >> beam.FlatMap(open_urls, **kws)
+                | f"Open with fsspec ({self.max_concurrency=})" >> beam.FlatMap(_open_urls, **kws)
             )
         )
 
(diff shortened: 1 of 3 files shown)


cisaacstern:
This appears to work in its current form. The climsim recipes I deployed with `OpenURLWithFSSpec(max_concurrency=20)` via https://github.com/leap-stc/data-management/pull/44 have now been running for over a day without erroring out:

Screen Shot 2023-08-09 at 9 33 22 PM

Indeed, following a quick spike to 20 at the start of the job, the number of workers has been locked at 10 for most of this time (with 2 vCPUs per worker, 10 workers means 20 vCPUs, so maybe this means two concurrency groups are running per worker?):

Screen Shot 2023-08-09 at 9 47 26 PM

At any rate, I've very encouraged by this (slow and steady) caching progress, as compared to previous deployment attempts of these recipes, wherein:

- **_No concurrency limit + unrestricted workers_**: crashed in ~20 minutes with `aiohttp` connection errors
- **_No concurrency limit + `max_num_workers=50`_**: crash in ~1 hr with `aiohttp` connection errors

In the day plus of running, the `climsim-highres-mli` recipe has cached a little over 100k files (~10 TB, at ~100 MB/file):

```python
import gcsfs
gcs = gcsfs.GCSFileSystem()
cache_mli = [f for f in gcs.ls("gs://leap-scratch/data-library/cache") if "mli" in f]
len(cache_mli)  # -> 100238
```

I'm working on an integration test which will hopefully be able to capture this behavior in a (much) smaller pipeline. In the meantime, this work is unit tested here, so I thought I'd open it up for review to see what others think.

I've made the concurrency-limiter here a standalone ptransform which can wrapped by other, more specific transforms, as demonstrated for `OpenURLWithFSSpec`. The approach is inspired by the above-linked code in `weather-tools` (thanks for that reference, @alxmrs!), but is obviously a bit simpler, insofar as it _only_ does concurrency limiting (no rate limiting/latency between invocations). In exchange for being less general and simpler than the `weather-tools` example, I do like the readability we gain with this implementation, which IMHO is easier to grok than the (much more general and fully featured!) `weather-tools` implementation.

My feeling is that merging some version of what's here is the easiest way to move forward with blocked recipes, and that if we need something more general down the line, we can address that later. That being said, if others see it another way, I'm all ears. Thanks in advance for taking a look!

cisaacstern:
A few additional notes/thoughts:

- I am curious if there is any approach better than trial-and-error for tuning the selection of a max concurrency value. I semi-arbitrarily chose `max_concurrency=20` as a conservative value for the ClimSim deployments above, after observing the `max_num_workers=50` experiment fail.
- Because the ongoing ClimSim jobs (my first real-world test of this logic), are still in the caching phase, I have not yet definitively confirmed that subsequent stages will scale back out to unrestricted numbers of workers. There is no reason I can think of that this _won't_ happen, but I suppose there's a small chance a reshuffle may be needed to trigger it? I will report back here with what I see in the ClimSim jobs, and also I'd like to see if it's possible to capture this scale-out as part of a forthcoming integration test.
- I wonder if it's worthwhile to add any backoff/retry logic for each call to `open_url`. In practice, this might be easiest (and most general) to implement as backoff/retries of the `result = func(...)` call here (in `OpenURLWithFSSpec`, this `func` is `open_url`):

  https://github.com/pangeo-forge/pangeo-forge-recipes/blob/3b4f0c50ad249c1a9dcf68e9a45d2ef5be616a15/pangeo_forge_recipes/transforms.py#L89-L92

    My thinking here is that a given concurrency group could be very large (in the current ClimSim jobs, each concurrency group is roughly ~10k elements). If there is an intermittent networking issue causing a single call to `open_url` to error out, perhaps the most graceful way to recover is to backoff and retry it at the worker code level? My concern would be that if we leave it to the runner to retry (IIUC Dataflow, e.g., has builtin retry routines), this might result in the _entire concurrency group_ being iterated over from the beginning? If so, that seems undesirable.

pangeo_forge_recipes/transforms.py
```diff
@@ -74,6 +75,26 @@ def wrapper(arg, **kwargs):
     return wrapper
 
 
+def _add_keys_iter(func: Callable):
```
rabernat, line 78: Add a return type

pangeo_forge_recipes/transforms.py
```diff
@@ -82,12 +103,51 @@ def _drop_keys(kvp):
...
+    number of concurrent calls. Useful for situations where the provided function requests data
+    from an external service that does not support an unlimited number of concurrent requests.
+
+    :param map_func: The function to pass to beam.Map (in the case of no concurrency limit)
+      or beam.FlatMap (if max_concurrency is specified).
+    :param op_name: The name of the operation performed by `map_func`.
+    :param map_func_kws: Kwargs passed to all invocations of `map_func`.
+    :param max_concurrency: The maximum number of concurrent invocations of `map_func`.
+      If unspecified, no limit is imposed by this transform (therefore the concurrency
+      limit will be set by the Beam Runner's configuration).
```
rabernat, line 128: I would try to make the signature of this PTransform stay as close as possible to that of the built-in Map transform: https://beam.apache.org/releases/pydoc/2.49.0/apache_beam.transforms.core.html#apache_beam.transforms.core.Map

rabernat (approved):
Overall this looks great!

A few nits on type hints and function signature.

pangeo_forge_recipes/transforms.py
cisaacstern: ✅ https://github.com/pangeo-forge/pangeo-forge-recipes/pull/557/commits/02d6e3c5ef5b57c029762e81990f97fa4787ada3

pangeo_forge_recipes/transforms.py
cisaacstern: ✅ https://github.com/pangeo-forge/pangeo-forge-recipes/pull/557/commits/95a9af2caad15904722bcacc22999c0014d5e192

Good call. This looks better.

pangeo_forge_recipes/transforms.py
```diff
@@ -56,7 +57,7 @@
 
 
 # TODO: replace with beam.MapTuple?
-def _add_keys(func):
+def _add_keys(func: Callable) -> Callable:
```
alxmrs, line 60: A Callable of what?

cisaacstern: Touché, @alxmrs, that was a cunning nerd snipe. 😆 

To find the answer, I just went down the (interesting, not-too-deep) rabbit hole of [PEP612](https://peps.python.org/pep-0612/), and arrived at https://github.com/pangeo-forge/pangeo-forge-recipes/pull/557/commits/5ebb0f72518eb6bd4e4984c5f40a9fe12f0fea27.

Do those hints look plausible/intelligible to you?

cisaacstern: I feel like these annotation are correct! To keep things moving, I'm going to merge, @alxmrs if you catch anything out-of-place, feel free to let me know and I'll fix it in a follow-on.

Thanks for encouraging more specificity here, I think this really helps with clarity. 😃

PR #561, opened 2023-08-09
Restore File Pattern Docs (with Globus workaround)

### LINKED ISSUE #558: Is the Globus workaround from #408 still supported with v0.10.x?
I am discussing moving data from GFDL into LEAP/m2lines cloud buckets (with @adcroft), and just scrolled through some past issues/PRs on the matter (#222, #408). In particular I am wondering if the approach in #408 is still valid with the beam refactor. 
The parts of the docs seem to have been removed, but I am not sure why something in the `file_pattern` API would have necessarily changed, even though I might be missing something here. 

If this is generally still applicable Id be happy to try this out.

jbusecke:
Closes #558. 

This so far is a simple copy and past based on @rabernat s [comment](https://github.com/pangeo-forge/pangeo-forge-recipes/issues/558#issuecomment-1670010751). Could someone confirm that all the information added is in fact working for the new beam version?

### DIFF
--- a/docs/pangeo_forge_recipes/recipe_user_guide/file_patterns.md
+++ b/docs/pangeo_forge_recipes/recipe_user_guide/file_patterns.md
@@ -9,6 +9,85 @@ kernelspec:
 
 # File Patterns
 
+## Pangeo Forge Pulls Data
+
+A central concept in Pangeo Forge is that data are "pulled", not "pushed" to
+the storage location. You tell Pangeo Forge where to find your data; when you
+execute a recipe, the data will automatically be downloaded and transformed.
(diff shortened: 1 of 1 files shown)


rabernat:
> Could someone confirm that all the information added is in fact working for the new beam version?

Unfortunately I think the best candidate to confirm this would be you. 😬

cisaacstern:
> Unfortunately I think the best candidate to confirm this would be you. 

@jbusecke, I would have to agree 😆 . Thanks for all your hard work of late!

jbusecke:
Looool. I would say LGTM and Ill promise to raise/correct issues once I test things (particularly with globus).

cisaacstern:
Sounds like a plan.

PR #562, opened 2023-08-10
Return target_store from StoreToZarr stage

jbusecke:
Superseeds https://github.com/jbusecke/pangeo-forge-recipes/pull/1 (see over there for discussion).


cisaacstern:
This looks good, and the change is from returning nothing (`store_dataset_fragment` returns `None`)

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/934e34b4bc87234e420b6e2c23c7305ddfd578d3/pangeo_forge_recipes/writers.py#L69-L71

to returning something, so I don't think we need to worry about a breaking change.

This is simple enough that I think we can leave it without an explicit test. Thanks @jbusecke, great enhancement!

jbusecke:
Yayyyy! Ill move on with building testing an cataloging stages in the CMIP6 repo. Thanks for the quick merge.

PR #565, opened 2023-08-19
Move almost everything to pyproject.toml

yuvipanda:
- Mixing pyproject.toml and setup.cfg the way we were doing seems to cause weird problems with entrypoints
- I've moved things to their equivalent values, except when they are the default - in which case I've left it out
- I don't understand if the mypy things are still needed, so I've left them as they are.
- flake8 doesn't support reading from pyproject.toml yet, so letting it be.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,11 +1,85 @@
-[tool.black]
-line-length = 100
-
-# pyproject.toml
 [build-system]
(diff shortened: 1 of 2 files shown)


cisaacstern:
LGTM, thanks @yuvipanda!

PR #567, opened 2023-08-19
Support python 3.11 too

yuvipanda:
Newer versions of apache beam now support python 3.11 too, no reason we can't.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -34,7 +34,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


yuvipanda:
Goes along with https://github.com/pangeo-forge/pangeo-forge-runner/pull/65, where we unpin version of apache_beam from pangeo-forge-runner. Funnily, recipes doesn't actually seem to have an explicit dependency on beam so the version of beam we got was always from runner I think.

cisaacstern:
Sorry for the delay on this, looks awesome @yuvipanda! Thanks so much!

PR #556, opened 2023-08-04
[WIP] Add ConsolidateDimensionCoordinates + ConsolidateZarrMetadata transforms

norlandrhagen:
This PR is intended to kickstart issue #554. It lays down some of the groundwork outlined in @cisaacstern's recommendation under issue 544.  

cc @abarciauskas-bgse @cisaacstern 

Currently fails when running end_to_end tests.

```bash
pytest tests/test_end_to_end.py -k 'test_xarray_zarr[netcdf_local_file_pattern_sequential-target_chunks0-netcdf_local_paths_sequential_1d]'
```

``` python
ValueError: Region (slice(1, 2, None),) does not align with Zarr chunks (10,). [while running 'StoreToZarr/StoreDatasetFragments/Map(store_dataset_fragment)']
```

Traceback:

<details>

```python

pytest tests/test_end_to_end.py -k 'test_xarray_zarr[netcdf_local_file_pattern_sequential-target_chunks0-netcdf_local_paths_sequential_1d]'
================================================================================================ test session starts ================================================================================================
platform darwin -- Python 3.10.12, pytest-7.4.0, pluggy-1.2.0
rootdir: /Users/nrhagen/Documents/carbonplan/pangeo_forge/pangeo-forge-recipes
configfile: setup.cfg
plugins: nbmake-1.4.1, timeout-2.1.0, cov-4.1.0, anyio-3.7.1, lazy-fixture-0.6.3
timeout: 30.0s
timeout method: signal
timeout func_only: False
collected 22 items / 21 deselected / 1 selected                                                                                                                                                                     

tests/test_end_to_end.py FE                                                                                                                                                                                   [100%]

====================================================================================================== ERRORS =======================================================================================================
____________________________________________ ERROR at teardown of test_xarray_zarr[netcdf_local_file_pattern_sequential-target_chunks0-netcdf_local_paths_sequential_1d] ____________________________________________

vname = 'time', var = <xarray.Variable (time: 1)>
array([1])
Attributes:
    units:     days since 2010-01-01
    calendar:  proleptic_gregorian
index = {Dimension(name='time', operation=<CombineOp.CONCAT: 2>): IndexedPosition(value=1, indexed=True, dimsize=10)}, zgroup = <zarr.hierarchy.Group '/'>

    def _store_data(vname: str, var: xr.Variable, index: Index, zgroup: zarr.Group) -> None:
        zarr_array = zgroup[vname]
        # get encoding for variable from zarr attributes
[...]

PR #575, opened 2023-08-23
Provide consolidate_metadata option in StoreToZarr

### LINKED ISSUE #568: Re-implement consolidated metadata in `StoreToZarr`
As noted here

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/561897d003231ee3c6e97427015847adaef1714e/pangeo_forge_recipes/transforms.py#L386C1-L388

consolidated metadata is not implemented in `StoreToZarr`. Here's the prior implementation:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/3b3c13c358259a9b01367c91951d28826b6c252e/pangeo_forge_recipes/recipes/xarray_zarr.py#L694-L697

Like #556, this also depends on #564, so I'll address that first.

cisaacstern:
Closes #568

Just need to add a test before merging

### DIFF
--- a/pangeo_forge_recipes/transforms.py
+++ b/pangeo_forge_recipes/transforms.py
@@ -494,6 +494,8 @@ class StoreToZarr(beam.PTransform, ZarrWriterMixin):
     :param combine_dims: The dimensions to combine
     :param target_chunks: Dictionary mapping dimension names to chunks sizes.
         If a dimension is a not named, the chunks will be inferred from the data.
+    :param consolidate_metadata: Whether to consolidate metadata in the resulting
+        Zarr dataset. Defaults to ``True``.
     """
 
     # TODO: make it so we don't have to explicitly specify combine_dims
--- a/pangeo_forge_recipes/transforms.py
+++ b/pangeo_forge_recipes/transforms.py
@@ -460,6 +460,33 @@ class WriteCombinedReference(beam.PTransform, ZarrWriterMixin):
         )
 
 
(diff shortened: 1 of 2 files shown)


cisaacstern:
I think this is close but I'm not sure I'm totally convinced by the testing yet, so will revisit that before merging.

cisaacstern:
@rabernat I'm a bit confused by the expected behavior here and would appreciate your insight. It looks like the zarr store created by `schema_to_zarr` here

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/db39eefc505a65f0ab2925ac6fd761d0a4380728/pangeo_forge_recipes/aggregation.py#L277

already contains a `.zmetadata` object, even though the call to `ds.to_zarr` does not specify `consolidated=True`.

Am I correct in assuming that there is some advantage to calling `zarr.consolidate_metadata` on the target store _after_ the dataset fragments have been stored? As I attempt in line 523 here

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/e90c5ce9f5a4aa9d2f387edfbc582a869f065654/pangeo_forge_recipes/transforms.py#L519-L523

... or is all of the consolidated metadata already encoded in the schema and what I'm doing here is actually just redundant?

cisaacstern:
Per discussion with @norlandrhagen, we're going to include this as part of #556, so closing this.

cisaacstern:
Per discussion with @norlandrhagen, we're going to include this as part of #556, so closing this.

PR #566, opened 2023-08-19
Tell pangeo-forge-runner what to inject

yuvipanda:
Tells pangeo-forge-runner, once https://github.com/pangeo-forge/pangeo-forge-runner/pull/87 lands,
what values to inject.

PR #588, opened 2023-08-30
Default value for `target_root`, to make recipe modules importable

### LINKED ISSUE #587: Recipe modules with injectable kws should still be importable (use defaults?)
We currently expect recipes to leave certain injectable kws empty, e.g. `StoreToZarr.target_root`, with the understanding that `pangeo-forge-runner` will inject values for these kws at deploy time. (Better documentation of this is forthcoming.)

`StoreToZarr.target_root` is required, meaning modules leaving it unassigned are not importable, raising

```
TypeError: __init__() missing 1 required positional argument: 'target_root'
```

This is suboptimal for at least two reasons:

- Linters/type-checkers don't like it
- We can't import this module to do other things with it outside the `pangeo-forge-runner` deployment context. An example (which motivated this opening issue for me) would be to import functions defined in the module for unit testing.

I see a few possible fixes:

- ❌ Use [`kw_only` dataclasses](https://medium.com/@aniscampos/python-dataclass-inheritance-finally-686eaf60fbb5) to allow the `ZarrWriterMixin` (from which `StoreToZarr` inherits `target_root`) to define an inheritable default for `target_root`. This is my favorite option but its not tractable because this feature is only available in Python 3.10 and not easily back-ported into 3.9.
- ❌ Set a default `target_root` on `ZarrWriterMixin` and give all args on its child classes defaults. I don't like this because some of the child class args are required (and therefore should not have defaults).
- ✅ Move `target_root` out of `ZarrWriterMixin` and just require that child classes declare this arg themselves. Possibly enforceable via a `__post_init__` on `ZarrWriterMixin`? This means the only value of the mixin is to define the `get_forge_target` method, which still seems worthwhile to me.

In the meantime, the workaround for this would be to just give a placeholder value to `StoreToZarr.target_root` in the recipe, under the assumption that `pangeo-forge-runner` will still overwrite it. 🤔

cisaacstern:
Closes #587

### DIFF
--- a/pangeo_forge_recipes/transforms.py
+++ b/pangeo_forge_recipes/transforms.py
@@ -438,6 +438,8 @@ class WriteCombinedReference(beam.PTransform, ZarrWriterMixin):
     :param output_json_fname: Name to give the output references file. Must end in ``.json``.
     """
 
+    store_name: str
+    target_root: Union[str, FSSpecTarget] = field(default_factory=os.getcwd)
     output_json_fname: str = "reference.json"
 
     def expand(self, reference: beam.PCollection) -> beam.PCollection:
--- a/pangeo_forge_recipes/transforms.py
+++ b/pangeo_forge_recipes/transforms.py
@@ -69,14 +69,17 @@ IndexedReturn = Tuple[Index, R]
 P = ParamSpec("P")
 
 
-class EmptyDefault:
-    """Default for values that must be set or injected."""
+class RequiredAtRuntimeDefault:
+    """Sentinel class to use as default for transform attributes which are required to run a
+    pipeline, but may not be available (or preferable) to define during recipe develoment; for
+    example, the ``target_root`` kwarg of a transform that writes data to a target location. By
+    using this sentinel as the default value for such an kwarg, a recipe module can define all
+    required arguments on the transform (and therefore be importable, satisfy type-checkers, be
+    unit-testable, etc.) before it is deployed, with the understanding that the attribute using
+    this sentinel as default will be re-assigned to the desired value at deploy time.
+    """
 
     pass
-
-
-def empty_default():
-    return EmptyDefault()
 
 
 # TODO: replace with beam.MapTuple?
@@ -452,7 +455,9 @@ class WriteCombinedReference(beam.PTransform, ZarrWriterMixin):
     """
 
     store_name: str
-    target_root: Union[str, FSSpecTarget, EmptyDefault] = field(default_factory=empty_default)
+    target_root: Union[str, FSSpecTarget, RequiredAtRuntimeDefault] = field(
+        default_factory=RequiredAtRuntimeDefault
+    )
     output_json_fname: str = "reference.json"
 
     def expand(self, reference: beam.PCollection) -> beam.PCollection:
@@ -480,7 +485,9 @@ class StoreToZarr(beam.PTransform, ZarrWriterMixin):
     # Could be inferred from the pattern instead
     combine_dims: List[Dimension]
     store_name: str
-    target_root: Union[str, FSSpecTarget, EmptyDefault] = field(default_factory=empty_default)
+    target_root: Union[str, FSSpecTarget, RequiredAtRuntimeDefault] = field(
+        default_factory=RequiredAtRuntimeDefault
+    )
     target_chunks: Dict[str, int] = field(default_factory=dict)
 
     def expand(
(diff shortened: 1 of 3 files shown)


pangeo_forge_recipes/transforms.py
```diff
@@ -437,6 +438,8 @@ class WriteCombinedReference(beam.PTransform, ZarrWriterMixin):
     :param output_json_fname: Name to give the output references file. Must end in ``.json``.
     """
 
+    store_name: str
+    target_root: Union[str, FSSpecTarget] = field(default_factory=os.getcwd)
```
yuvipanda, line 442: hmmmmm, I'm not a fan of getcwd as now behavior fundamentally differs based on where we're running this from. This leads to confusing to debug stuff...

I made a different suggestion for fixing this in https://github.com/pangeo-forge/pangeo-forge-recipes/issues/587#issuecomment-1699955200.

yuvipanda (changes requested):

pangeo_forge_recipes/transforms.py
cisaacstern: Fair enough!

What about a sentinel value (based on [this SO](https://stackoverflow.com/a/69239756) and [this PEP](https://peps.python.org/pep-0661/#using-class-objects)), something like:
```python
class InjectableDefault:
   """Serves as a default for values which we expect to be either set or injected."""
   pass

def injectable_default():
    return InjectableDefault()

@dataclass
class StoreToZarr(...):
    ...
    target_root: Union[str, FSSpecTarget, InjectableDefault] = field(default_factory=injectable_default)
```

cisaacstern: Does that possibly make debugging clearer because it just won't work?

Trying to see if we can add this as a small patch without dropping a whole Python version support...

cisaacstern: A possible benefit of this `InjectableDefault` approach is that it provides a type hint in the code which can flag a given argument as also being declared in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/566 (once that lands)? So when we read the `transforms` module... we know what we expect to be injected, without cross-referencing to the injection spec added by that PR...?

yuvipanda: Nah, we should just drop the python version support! If we want to do something with types, we can do that separately from this - but IMO in this case, the right way to tackle this is to drop 3.9 support.

cisaacstern: > IMO in this case, the right way to tackle this is to drop 3.9 support.

Just realizing that the only thing dropping 3.9 solves is where we set the default, not what the default is.

With 3.10, we can set the default on `ZarrWriterMixin` and still have required fields on child classes. With 3.9, we need to set the default on the child classes.

But in either case, we need to chose _what_ the default is. https://github.com/pangeo-forge/pangeo-forge-recipes/pull/588/commits/5a37cb7a7c5a17412517e176951a97e6a07e3a7d demonstrates the sentinel class idea (again, this could be thought of as agnostic to 3.9 or 3.10, we need a default either way).

I think the empty sentinel works nicely, and indeed prevents the recipe from being run in the first place.

What if we use something like this here, so we can just merge this, and drop 3.9 in a more deliberate PR?

pangeo_forge_recipes/transforms.py
```diff
@@ -435,11 +444,15 @@ def expand(self, references: beam.PCollection) -> beam.PCollection:
...
+    :param store_name: Name for the Zarr store. It will be created with
+        this name under `target_root`.
+    :param target_root: Root path the Zarr store will be created inside;
+        `store_name` will be appended to this prefix to create a full path.
     :param output_json_fname: Name to give the output references file. Must end in ``.json``.
     """
 
     store_name: str
-    target_root: Union[str, FSSpecTarget] = field(default_factory=os.getcwd)
+    target_root: Union[str, FSSpecTarget, EmptyDefault] = field(default_factory=empty_default)
```
yuvipanda, line 455:
```suggestion
    target_root: Union[str, FSSpecTarget, EmptyDefault] = field(default_factory=EmptyDefault)
```

yuvipanda: Classes are also callable, so this will directly work.

yuvipanda (changes requested):
Let's call it `RequiredAtRuntimeDefault` or something of that sort?

cisaacstern:
> Let's call it RequiredAtRuntimeDefault or something of that sort?

Love it... updating now...

cisaacstern:
@yuvipanda how does this look now?

cisaacstern:
Going to merge:

✅ All checks passing
✅ All review items addressed

🙏 Thanks @yuvipanda! This is a lot better thanks to your feedback.

Issue #569, opened 2023-08-22
Docs + integration testing reboot


cisaacstern:
Umbrella issue for a docs + integration testing reboot following the `0.10.0` release. These two topics are grouped together because I propose we model our docs on FastAPI, where all of the docs tutorials ["are actually tested Python files"](https://fastapi.tiangolo.com/tutorial/#run-the-code). This is achieved there by:

- Writing the tutorial code as a plain python file, e.g. https://github.com/tiangolo/fastapi/blob/master/docs_src/first_steps/tutorial001.py
- Having a pytest test for each file, e.g. https://github.com/tiangolo/fastapi/blob/master/tests/test_tutorial/test_first_steps/test_tutorial001.py 
- In the docs, rendering the tested file contents (and _not writing any code as text in the markdown!), e.g. https://github.com/tiangolo/fastapi/blob/master/docs/en/docs/tutorial/first-steps.md#first-steps
- Which all comes together to produce a tested tutorial, e.g. https://fastapi.tiangolo.com/tutorial/first-steps/#first-steps

In #563, I've done some experimenting and convinced myself this approach is indeed adaptable for our use case. In addition to the structural exploration, #563 also proves out how we can make integration tests _**look & run like actual feedstocks**: stored as `.py` files, run with `pangeo-forge-runner`, etc. This should go a long way to ensuring compatibility with `-runner` going forward.

This reboot is probably cleanest if spread across a few PRs:

- [ ] Implement integration testing separate from any docs changes. I will do this as a first step in #563 (or perhaps another PR which supersedes it).
- [ ] Revert the nested hierarchy of the current docs (Pangeo Forge Recipes, and Pangeo Forge Cloud headings) in favor of making these docs "just" about `pangeo-forge-recipes`. The Cloud docs are now out-of-date, and doing this should make the final item easier, insofar as it impacts the overall narrative arc of the docs.
- [ ] Adapt current notebook tutorials to reference the exact code of these integration tests, as shown in the FastAPI example.

I will get started on the first two items here, and anticipate that the final item (adapting the tutorial notebooks) is where I could use the most help from @derekocallaghan (who's expressed interest in this). Derek, as noted here, still hammering out the foundation for this, I'll ping you as soon as I have something actionable for you to look at.

cisaacstern:
> Adapt current notebook tutorials to reference the exact code of these integration tests, as shown in the FastAPI example.

[...]

PR #593, opened 2023-08-31
Try sphinx 6.2.1

cisaacstern:
Dependabot keeps suggesting we upgrade sphinx (e.g. #592), and docs builds lately fail on those PRs, so I've been closing them.

I know we can have Dependabot ignore sphinx upgrades, but before we do that, I wanted to see if we can bump to this version, because we admittedly are on a quite old sphinx version currently.

I've chosen 6.2.1 because its the latest version in the 6.*.* series, and our base theme, `sphinx-book-theme`, [supports `<7`](https://github.com/executablebooks/sphinx-book-theme/blob/a9794b73cec766e01bee35a96f7d44e22851f278/pyproject.toml#L25).

Presumably the first build will fail here due to incompatibility with other pinned docs requirements. I'll clean that up in a follow-on commit here.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,4 @@
-sphinx==3.5.4
+sphinx==6.2.1
 pangeo-sphinx-book-theme==0.2
 myst-parser==0.13.6
 myst-nb==0.12.3


cisaacstern:
This is failing with

```
The conflict is caused by:
    The user requested sphinx==6.2.1
    pangeo-sphinx-book-theme 0.2 depends on sphinx
    myst-parser 0.13.6 depends on sphinx<4 and >=2
```
so seems like it would require a fix in https://github.com/pangeo-forge/pangeo-sphinx-book-theme/

cisaacstern:
> so seems like it would require a fix in https://github.com/pangeo-forge/pangeo-sphinx-book-theme/

I misunderstood the error. Deeper dive on this locally leaves me to conclude that `myst-nb`/`myst-parser` interactions are a bit of a blocker here, so I think we should pause on this until #569 is complete, at which point I think we can drop `myst-nb`.

cisaacstern:
Closing as a failed experiment.

PR #594, opened 2023-08-31
Adding parquet reference file storage

norlandrhagen:
The current version of pangeo-forge-recipes supports reference recipe creation. The transform defaults to saving the reference files as `.json`. Semi-recently, Kerchunk added the ability to save reference files as `parquet`. [ex1](https://fsspec.github.io/kerchunk/advanced.html?highlight=parquet#parquet-storage), [ex2](https://projectpythia.org/kerchunk-cookbook/notebooks/foundations/04_kerchunk_reference_storage.html).

For large reference recipes, implementing the ability to write references as parquet could be beneficial. 

Some open questions:
- The parquet references are written with `fsspec.LazyReferenceMapper` along with a few other specific kwargs. Is it possible to impliment these to an already created MultiZarrToZarr object in `WriteCombinedReference` instead of in the complex `CombineReferences` transform?


cisaacstern:
> The parquet references are written with fsspec.LazyReferenceMapper

Is there an example or documentation of this you can share?

> Is it possible to impliment these to an already created MultiZarrToZarr object in WriteCombinedReference instead of in the complex CombineReferences transform?

Seems like you've gotten a start on this already here?

norlandrhagen:
Note: pre-commit seems to want to add in a bunch of notebooks from docs.

cisaacstern:
@norlandrhagen I fixed the pre-commit issue in #602. To get back on track with pre-commit here:

- Revert the changes it make to `docs/` so that `docs/` here is equal with `docs/` in `main`
- Merge `main` into this PR
- Should be good to go

norlandrhagen:
Wonderful thanks!

norlandrhagen:
@cisaacstern, I got the .pre-commit fix from main merged in. Any chance you mind taking a look at this PR?

pangeo_forge_recipes/transforms.py
```diff
@@ -447,24 +447,23 @@ def expand(self, references: beam.PCollection) -> beam.PCollection:
...
+    :param output_file_name: Name to give the output references file.
+    :param output_file_type: Filetype to store reference files as. Options are: "parquet" or "json".
+    :param concat_dims: concat_dims kwarg to pass to write_combined_reference if using
+    .parquet as a storage format.
     """
 
-    store_name: str
-    target_root: Union[str, FSSpecTarget, RequiredAtRuntimeDefault] = field(
-        default_factory=RequiredAtRuntimeDefault
-    )
```
cisaacstern, line 460: I think the call below to

```python
full_target=self.get_full_target()
```
will fail without these kwargs.

Did you mean to remove these?

pangeo_forge_recipes/transforms.py
```diff
@@ -447,24 +447,23 @@ def expand(self, references: beam.PCollection) -> beam.PCollection:
...
-    output_json_fname: str = "reference.json"
+    output_file_name: str = "reference"
+    output_file_type: str = "json"
+    concat_dims: List[str] = field(default_factory=list)
 
     def expand(self, reference: beam.PCollection) -> beam.PCollection:
         return reference | beam.Map(
             write_combined_reference,
             full_target=self.get_full_target(),
```
cisaacstern, line 463: See comment above. This relies on `store_name` and `target_root`.

cisaacstern (commented):
Nice progress, @norlandrhagen.

Left a few comments about removing `store_name` and `target_root`. This seems possibly unintentional? If not, can you explain the reasoning there? I am in favor of keeping these for now if possible, unless there's a good reason not to.

pangeo_forge_recipes/transforms.py
norlandrhagen: I did not! Reverted.

cisaacstern:
FYI these test failures:

> https://github.com/pangeo-forge/pangeo-forge-recipes/actions/runs/6100768929/job/16555841114?pr=594#step:9:28578

Appear to be a general compatibility problem with our ci envs + test suite, as they've just started cropping up in all open PRs.

I'm investigation a fix now.

cisaacstern:
> I'm investigation a fix now.

Fixed by #603, merging `main` should get you back on track here 🙂

norlandrhagen:
@cisaacstern, thanks for the upstream test fix! I just tried this branch locally on a larger recipe and it worked great. What do you think about merging?

pangeo_forge_recipes/writers.py
```diff
@@ -95,22 +97,44 @@ def store_dataset_fragment(
...
             f.write(ujson.dumps(multi_kerchunk).encode())
+
+    elif output_file_type == "parquet":
+
+        # Creates empty parquet store to be written to
+        if os.path.exists(outpath):
+            import shutil
+
+            shutil.rmtree(outpath)
+        os.makedirs(outpath)
```
cisaacstern, line 122: This will presumably fail against cloud storage.

Rather than `shutil` and `os`, we should be using the cloud-storage-aware methods on the `full_target` kw passed to this function. Available methods on the class definition here:


https://github.com/pangeo-forge/pangeo-forge-recipes/blob/356ea2a64e7e15dff27c2497f232a4312b2debc7/pangeo_forge_recipes/storage.py#L85

Let's try to make it work wit the existing methods there, but if not we could add methods as needed.

pangeo_forge_recipes/transforms.py
```diff
@@ -446,25 +445,33 @@ def expand(self, references: beam.PCollection) -> beam.PCollection:
...
-
     :param store_name: Name for the Zarr store. It will be created with
         this name under `target_root`.
     :param target_root: Root path the Zarr store will be created inside;
         `store_name` will be appended to this prefix to create a full path.
-    :param output_json_fname: Name to give the output references file. Must end in ``.json``.
+    :param target_chunks: Dictionary mapping dimension names to chunks sizes.
+        If a dimension is a not named, the chunks will be inferred from the data.
+    :param output_file_name: Name to give the output references file.
+    :param output_file_type: Filetype to store reference files as. Options are: "parquet" or "json".
```
cisaacstern, line 455:
```suggestion
    :param output_file_name: Name to give the output references file.
    :param output_file_type: Filetype to store reference files as. Options are: "parquet" or "json".
```

Presumably `target_chunks` was not meant to be here?

Also, is it possibly simpler to just have `output_file_name` include a suffix of either `.json` or `.parquet` and infer the file type from the suffix in the body of `write_combined_references`? I see you removed the following below

```python
file_ext = os.path.splitext(output_json_fname)[-1]

if file_ext == ".json":
```

Was there a reason to not just use that? I feel like its friendlier to the user to just specify a single kwarg here.

pangeo_forge_recipes/transforms.py
```diff
@@ -446,25 +445,33 @@ def expand(self, references: beam.PCollection) -> beam.PCollection:
...
+    .parquet as a storage format.
     """
 
     store_name: str
     target_root: Union[str, FSSpecTarget, RequiredAtRuntimeDefault] = field(
         default_factory=RequiredAtRuntimeDefault
     )
-    output_json_fname: str = "reference.json"
+    output_file_name: str = "reference"
+    output_file_type: str = "json"
```
cisaacstern, line 465: See question above about have a single kw and inferring the file type from the suffix.

pangeo_forge_recipes/writers.py
```diff
@@ -95,22 +97,44 @@ def store_dataset_fragment(
 def write_combined_reference(
     reference: MultiZarrToZarr,
     full_target: FSSpecTarget,
-    output_json_fname: str,
+    concat_dims: List[str],
+    output_file_name: str,
+    output_file_type: str,
```
cisaacstern, line 102: Note if we infer file type from suffix we would change consolidate these into one kw.

cisaacstern (commented):
Very close, @norlandrhagen 🙏 

Two main points in the review below:

- (big one) I believe we need to use the methods on `full_target` in order to operate against non-local filesystems (rather than `os` and `shutil`).
- (smaller one) Is there a reason we can't infer the file type from the file extension as we were doing before? I thought that was perhaps a bit of a simpler interface for the user.

norlandrhagen:
Thanks for the detailed review @cisaacstern!

Issue #604, opened 2023-09-07
Speed up unit test runs


cisaacstern:
Working on #603 just now, I noted that our unit tests take a long time to run. It would be good to speed this up, as the slow running tests do impede development iteration cycles to some degree.

Some areas that could potentially speed things up:

- [ ] Slim down test envs (though the current caching does provide some performance here). This could also entail simply removing the `ci/py3*.yaml` files in favor of various (smaller) sets of pip-installable optional dependencies, which would have a benefit for the realism with which our test environment replicates user environments. Note reverting #603 may be part of this work if the issue linked there is resolved (or if we don't use conda).
- [ ] Reduce future redundancy. I'm to blame for this one. A while ago, a fixtures refactor I did introduced a fair amount of grouping/lazy-fixturing in conftest such that a lot of tests that only need to run once, actually end up running in multiple parametrizations.
- [ ] (Re)consider role http server plays in tests. We start a real (local) http server in the tests to serve files back to ourselves. Not sure how much performance we'd gain by using this more sparingly, but might be worth considering.

yuvipanda:
Yeah I was super surprised by the complex conda setups here - IMO they add a lot of complexity and slow everything down. We can and *should* actually go pip only - once https://github.com/pangeo-forge/pangeo-forge-runner/pull/90 lands, there's no conda in the default image. I also don't use conda locally for anything, and haven't had problems with running the tests. There's also well developed pip caching for github actions supported upstream.

PR #605, opened 2023-09-07
Pip for ci

cisaacstern:
Towards #604

This PR tries just using `pip` for testing, with goal of:

- simplifying workflow files
- speeding up test runtimes
- making testing install more closely reflect users' installs


cisaacstern:
This seems like an overall win to me and an important, albeit incremental, improvement in test runtime. Specifically I view this as a:

- Big improvement in intelligibility and maintainability of the test workflow. This is _a lot_ simpler. Among other things, the _number_ of checks per commit is reduced by half (because each check previously had a `prepare-env` step which is now removed), which is big win for readability of the checks results which were getting quite cluttered.
- Big improvement in degree to which our test environment replicates users pip installed environments. This PR surfaced two big discoveries in that realm: **(1)** We had not specified `apache-beam` as a dependency 😱 ; and **(2)** GRIB pipelines should run in conda environments (due to the dependency on `cfgrib`), but other than that, everything here works from pip.
- Incremental improvement in test run time. My anecdotal comparison to checks run times on some other open PRs is that we might save about 1 minute of the ~11-15 minute test running time. That's not nothing, but bigger improvements will probably come with reducing fixture redundancy.

The simplification and user environment replication alone is worth it here, IMO, and the marginal speed up is a nice add-on. Before merging noting a few future directions for follow-on work:

- [ ] Restore GRIB testing with a separate `requires-conda` test setup somewhere
- [ ] Catching issues like #603 is nice, so we should generally try to test against conda environments somewhere, but maybe on `conda-forge/pangeo-forge-recipes-feedstock`?
- [ ] It's good to have the `upstream-dev` testing (and we should probably include `Zarr-python` there) but it's a big heavy to run on every commit. Maybe we can move that to nightly runs on a chron schedule?

Issue #606, opened 2023-09-08
Pip for ci (PR 605) follow-on work


cisaacstern:
A few future directions for follow-on work to complement #605:

- [ ] Restore GRIB testing with a separate `requires-conda` test setup somewhere
- [ ] Catching issues like #603 is nice, so we should generally try to test against conda environments somewhere, but maybe on `conda-forge/pangeo-forge-recipes-feedstock`?
- [ ] It's good to have the `upstream-dev` testing (and we should probably include `Zarr-python` there) but it's a big heavy to run on every commit. Maybe we can move that to nightly runs on a chron schedule?

_Originally posted by @cisaacstern in https://github.com/pangeo-forge/pangeo-forge-recipes/issues/605#issuecomment-1711077834_

keewis:
A few hints here:
- for `conda` in CI `mamba-org/setup-micromamba` is a good option as it allows environment caching, which on cache hit means that the environment can be set up in <10s (but this appears to heavily depend on the CI worker spec, as windows and macos CI often take longer than that).
- for scheduled upstream-dev I like using `xarray-contrib/issue-from-pytest-log` (though I'm the main author of that, so :shrug:) to open an issue as soon as the upstream-dev CI fails, which makes failures a bit more visible

PR #607, opened 2023-09-08
Schedule upstream dev tests

cisaacstern:
Towards #606

IMO we don't need to be testing against upstream dev versions on every commit of every PR.

This PR instead tests against upstream dev versions once nightly, which should still allow us to surface any potential issues in a timely fashion, while also reducing clutter on PR commit checks.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -9,6 +9,8 @@ on:
     branches: [ "main" ]
     paths-ignore:
     - 'docs/**'
+  schedule:
+    - cron: '0 4 * * *'  # run once a day at 4 AM
 
 env:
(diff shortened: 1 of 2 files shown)


cisaacstern:
This PR also tests against upstream dev installs of zarr-python and apache-beam. Zarr is installed from GitHub, like the other upstream dev requirements. Beam appears to follow a much more [regimented release candidate cycle ](https://pypi.org/project/apache-beam/#history) than any of the other upstream dev requirements, so we install that from any available release candidates instead of from GitHub.

cisaacstern:
This gets us down to only 6 checks per commit currently, nice! Much more approachable and developer-friendly IMHO. Merging! 😃 

Screen Shot 2023-09-07 at 11 19 02 PM

PR #620, opened 2023-09-14
Parquet storage option for reference recipe

norlandrhagen:


pangeo_forge_recipes/writers.py
```diff
@@ -95,23 +97,47 @@ def store_dataset_fragment(
...
+    elif file_ext == ".parquet":
+
+        # Creates empty parquet store to be written to
+        if full_target.exists(output_file_name):
+            full_target.rm(output_file_name, recursive=True)
+        full_target.makedir(output_file_name)
+
+        # kwargs to pass to MultiZarrToZarr
+        fs = fsspec.filesystem("file")
+        out = LazyReferenceMapper.create(refs_per_component, outpath, fs)
```
cisaacstern, line 126:
```suggestion
        out = LazyReferenceMapper.create(refs_per_component, outpath, full_target.fs)
```

I think we need the `full_target.fs` instead of `fsspec.filesystem("file")` here, since we want to be able to write to remote storage?

pangeo_forge_recipes/writers.py
```diff
@@ -95,23 +97,47 @@ def store_dataset_fragment(
...
+
+        # Calls MultiZarrToZarr on a MultiZarrToZarr object and adds kwargs to write to parquet.
+        MultiZarrToZarr(
+            [reference.translate()], concat_dims=concat_dims, remote_protocol="memory", out=out
```
cisaacstern, line 130:
```suggestion
        MultiZarrToZarr(
            [reference.translate()], concat_dims=concat_dims, remote_protocol=full_target.fs.protocol, out=out
```
Same concept as above comment, IIUC we need the more general `full_target.fs.protocol` here.

cisaacstern (commented):
Almost there, @norlandrhagen! Thanks so much for your work on this.

Just two suggestions below to generalize storage definitions for non-local storage. As we discussed in person last week, I think `minio` testing would help us catch these points in testing, and I've opened https://github.com/pangeo-forge/pangeo-forge-recipes/issues/623 to track that, but we don't need to let `minio` testing become a blocker to this work, just something to keep in mind for the future.

pangeo_forge_recipes/writers.py
```diff
@@ -1,9 +1,11 @@
 import os
-from typing import Protocol, Tuple, Union
+from typing import List, Protocol, Tuple, Union
 
+import fsspec
```
cisaacstern, line 4:
```suggestion
```

@norlandrhagen looks like pre-commit is failing bc this is now unused.

PR #630, opened 2023-09-21
Add `verify_existing` option to cache

cisaacstern:
Adds the option to disable verification of files in the cache against their remote sizes.

For long-running / large caching operations which need to be restarted, simply verifying remote sizes can itself represent a large time cost. In such cases, it may be preferable to disable verification of remote sizes, to allow for more expedient continuation of the caching operation.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ dependencies = [
     "cftime",
     "dask >= 2021.11.2",
     "fastparquet",
-    "fsspec[http] >= 2021.6.0",
+    "fsspec[http] >= 2023.4.0",
     "h5netcdf",
     "h5py >= 3.3.0",
     "kerchunk >= 0.0.7",


cisaacstern:
Related work:

- CLI passthrough: https://github.com/pangeo-forge/pangeo-forge-runner/pull/106
- Real world deployment test: https://github.com/pangeo-forge/aqua-modis-feedstock/pull/7

moradology (approved):
LGTM

pangeo_forge_recipes/storage.py
```diff
@@ -171,19 +171,26 @@ def _full_path(self, path: str) -> str:
...
     def cache_file(self, fname: str, secrets: Optional[dict], **open_kwargs) -> None:
         # check and see if the file already exists in the cache
         logger.info(f"Caching file '{fname}'")
-        if self.exists(fname):
+        exists = self.exists(fname)
+        if exists and self.verify_existing:
             cached_size = self.size(fname)
             remote_size = _get_url_size(fname, secrets, **open_kwargs)
             if cached_size == remote_size:
                 # TODO: add checksumming here
```
moradology, line 188: For future reference, do we have a way of doing this without downloading the file and computing the checksum? S3 provides that information but arbitrary HTTP endpoints can't be relied upon to provide the SHA in a predictable way

moradology: OK, so digging into fsspec source a bit, it appears that fsspec's `info` method ought to be used here. It will, by default, avoid opening the file and instead attempt to get both size and checksum from a `head` request. It looks to me as though there's no benefit to caching with the current method if `verify_existing` is `True`, as it will get the size of an opened file rather than using a small `HEAD` request

Would you like me to amend this PR with some logic to lean on those methods? Probably good to provide a warning in case it is absolutely necessary to fully load the file up (as happens here) for size/checksum validation

moradology:
I'm afraid we let this get pretty stale. It might be easier to incorporate the intended changes here in related changes in #750 ... I'm not happy to do it, but the rebase/merge here gets a bit nastier than it ought to

PR #632, opened 2023-09-21
Added parsing logic if multiple protocols are supplied

norlandrhagen:
https://github.com/norlandrhagen/pangeo-forge-recipes/pull/new/protocol

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ dependencies = [
     "cftime",
     "dask >= 2021.11.2",
     "fastparquet",
-    "fsspec[http] >= 2023.4.0",
+    "fsspec[http,s3] >= 2023.4.0",
     "h5netcdf",
     "h5py >= 3.3.0",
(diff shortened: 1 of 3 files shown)


norlandrhagen:
Totally see how `minio` would be helpful! For now I just added a simple test, but it would be nice to include that in the future todo.

pyproject.toml
```diff
@@ -29,7 +29,7 @@ dependencies = [
     "cftime",
     "dask >= 2021.11.2",
     "fastparquet",
-    "fsspec[http] >= 2023.4.0",
+    "fsspec[http,s3] >= 2023.4.0",
```
cisaacstern, line 32: Do we need `s3` here if we're including `s3fs` in dev dependencies below?

cisaacstern:
Probably need to pass `anon=True` to workaround the missing credentials error being thrown in the tests. 🙂

cisaacstern:
LGTM! Thanks so much @norlandrhagen!

Issue #634, opened 2023-09-27
Nest `CombineReferences` inside `WriteCombinedReferences`


cisaacstern:
Working on #610 has me thinking about how to easily communicate common recipe styles. Over there, I'm centering on the idea that the basic version of all common pipeline styles should be expressible as: `FilePattern | Opener | Writer`. `StoreToZarr` fits this style, but our kerchunk recipes do not:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/f6af0e03725caf93691cda4fc5724159a2671898/examples/feedstock/hrrr_kerchunk_concat_step.py#L48-L61

To align kerchunk recipes with this style, we can nest `CombineReferences` inside `WriteCombinedReferences`, making the basic kerchunk recipe expressible as:

```python

recipe = (
    beam.Create(pattern.items())
    | OpenWithKerchunk(...)
    | WriteCombinedReferences(...)
)
```

norlandrhagen:
This seems like a good call!

norlandrhagen:
Closed by #635

PR #635, opened 2023-09-27
Nests `CombineReference` inside of `WriteCombineReference` Transform

norlandrhagen:
PR for issue: https://github.com/pangeo-forge/pangeo-forge-recipes/issues/634

cc @cisaacstern

### DIFF
--- a/pangeo_forge_recipes/transforms.py
+++ b/pangeo_forge_recipes/transforms.py
@@ -429,8 +429,8 @@ class CombineReferences(beam.PTransform):
       dimensionality will match that of the accumulator.
     """
 
-    concat_dims: List[str]
-    identical_dims: List[str]
+    concat_dims: List[str] = field(default_factory=list)
+    identical_dims: List[str] = field(default_factory=list)
     mzz_kwargs: dict = field(default_factory=dict)
     precombine_inputs: bool = False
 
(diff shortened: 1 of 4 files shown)


pangeo_forge_recipes/transforms.py
```diff
@@ -448,6 +448,18 @@ def expand(self, references: beam.PCollection) -> beam.PCollection:
 @dataclass
 class WriteCombinedReference(beam.PTransform, ZarrWriterMixin):
     """Store a singleton PCollection consisting of a ``kerchunk.combine.MultiZarrToZarr`` object.
+    :param concat_dims: Dimensions along which to concatenate inputs.
```
cisaacstern, line 451: Looks like this is duplicated in the docstring (see bottom of param list).

pangeo_forge_recipes/transforms.py
```diff
@@ -457,14 +469,24 @@ class WriteCombinedReference(beam.PTransform, ZarrWriterMixin):
     .parquet as a storage format.
     """
 
-    store_name: str
+    concat_dims: List[str] = field(default_factory=list)
+    identical_dims: List[str] = field(default_factory=list)
+    mzz_kwargs: dict = field(default_factory=dict)
+    precombine_inputs: bool = False
+    store_name: Optional[str] = None
```
cisaacstern, line 476: `store_name` is required so let's not make that Optional if at all possible.

As for `concat_dims` & `identical_dims`, I think those are both required for `CombineReferences` so I think we should not be using a default factory. If either of these kws are not passed, we _do want_

```python
TypeError: __init__() missing 1 required positional argument:
```
to be raised, which the default factory will prevent.

pangeo_forge_recipes/transforms.py
```diff
@@ -429,8 +429,8 @@ class CombineReferences(beam.PTransform):
       dimensionality will match that of the accumulator.
     """
 
-    concat_dims: List[str]
-    identical_dims: List[str]
+    concat_dims: List[str] = field(default_factory=list)
+    identical_dims: List[str] = field(default_factory=list)
```
cisaacstern, line 433: Let's not use default factories here if at all possible.

IIUC these are required, so if they are not passed, we _do want_ a TypeError to be raised (which the default factory will prevent).

See related comment below.

tests/test_end_to_end.py
```diff
@@ -148,11 +146,9 @@ def test_reference_grib(
             p
             | beam.Create(pattern.items())
             | OpenWithKerchunk(file_type=pattern.file_type)
-            | CombineReferences(
+            | WriteCombinedReference(
                 concat_dims=[pattern.concat_dims[0]],
                 identical_dims=["latitude", "longitude"],
-            )
-            | WriteCombinedReference(
```
cisaacstern, line 155: This is beautiful diff 💐 😄 !

cisaacstern (commented):
Thanks for the lighting fast ⚡ pickup on this one, @norlandrhagen!

Looking good, a few comments below re: required args.

Also, before merging could you make this change to the two kerchunk recipes in https://github.com/pangeo-forge/pangeo-forge-recipes/tree/main/examples/feedstock as well?

Thanks so much!

norlandrhagen:
Thanks for the review @cisaacstern. Pretty sure all of the feedback is addressed!

cisaacstern:
Thanks so much @norlandrhagen !

Issue #608, opened 2023-09-08
`WriteCombinedReference` should emit a `zarr.storage.FSStore` (like `StoreToZarr`)


cisaacstern:
`StoreToZarr` emits a singleton PCollection containing a `zarr.storage.FSStore`. `WriteCombinedReference` should as well.

This is very useful for designing pipelines that _do_ something with the data once it's written, such as:

- Validate it with some tests
- Catalog it somewhere
- etc.

In #590 I am relying on this feature of `StoreToZarr` to do integration testing, so I'm blocked by this from integration testing kerchunk stores in the same manner.

cisaacstern:
This should be as simple as returning a `zarr.storage.FSStore` from this function

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/f094b024716485956d81b38c730aee2eaf6ee7da/pangeo_forge_recipes/writers.py#L95-L99

which is what `WriteCombinedReference` calls into here

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/f094b024716485956d81b38c730aee2eaf6ee7da/pangeo_forge_recipes/transforms.py#L464-L465

norlandrhagen:
I think this was addressed in the Parquet Kerchunk option PR: https://github.com/pangeo-forge/pangeo-forge-recipes/pull/620/files. On main, `write_combined_reference` now returns an [fsspec target](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/7e34030e6ea58d389641f7fa15dfb4e34be9467d/pangeo_forge_recipes/writers.py#L151C1-L151C23).

cisaacstern:
@norlandrhagen, looks like we're quite close but not all the way there: `full_target` is an `pangeo_forge_recipes.storage.FSSpecTarget`. The aim of this issue is to return a `zarr.storage.FSStore`. In practical terms, what we're aiming for is to be able to pass this `store` value directly into `xr.open_dataset`, i.e.

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/7e34030e6ea58d389641f7fa15dfb4e34be9467d/examples/feedstock/noaa_oisst.py#L24-L32

[...]

Issue #623, opened 2023-09-19
Test against `minio` target storage


cisaacstern:
In `pangeo-forge-runner`, @yuvipanda did an elegant job of setting up a `minio` instance to test recipes against s3-compatible storage. This is a nice way to catch oversights related to insufficiently generalized storage definitions. Probably a good thing to do here as well.

PR #636, opened 2023-10-02
WriteCombinedReference returns `zarr.storage.FSStore`

norlandrhagen:
PR to address Issue #608


pangeo_forge_recipes/writers.py
```diff
@@ -148,7 +148,14 @@ def write_combined_reference(
     else:
         raise NotImplementedError(f"{file_ext = } not supported.")
 
-    return full_target
+    # Return an fsspec mapper that can be read with Xarray
+    return ReferenceFileSystem(
+        outpath,
+        remote_protocol=remote_protocol,
+        remote_options={"anon": True},
+        target_protocol="file",
```
cisaacstern, line 156: Comments on each of these:

- `remote_protocol`: IIUC, this is the protocol the data lives on, which is not what this variable has been assigned to above. This variable is assigned to to the protocol that the reference is written to.
- `remote_options`: We don't know for sure that these are the options users will need to access the remote data. This needs to be configurable as a kwarg. Possibly `anon:True` is a sensible default for that kwarg.
- `target_protocol`: IIUC, this is the protocol on which the reference file exists. So it is the the value that the `remote_protocol` variable is currently assigned to.

pangeo_forge_recipes/writers.py
```diff
@@ -112,13 +112,15 @@ def write_combined_reference(
 ) -> FSSpecTarget:
     """Write a kerchunk combined references object to file."""
 
-    import ujson  # type: ignore
+    from fsspec.implementations.reference import ReferenceFileSystem
 
     file_ext = os.path.splitext(output_file_name)[-1]
-
     outpath = full_target._full_path(output_file_name)
+    remote_protocol = _select_single_protocol(full_target)
```
cisaacstern, line 119:
```suggestion
    target_protocol = _select_single_protocol(full_target)
```

See other comment re: difference between `remote_protocol` and `target_protocol`. IIUC, we have these backwards at the moment.

cisaacstern (commented):
Getting close, @norlandrhagen.

A few comments below re: generalizing the arguments to ReferenceFileSystem.

pangeo_forge_recipes/writers.py
norlandrhagen: So where/how can the remote protocol be inferred? Or should it be passed explicitly? 

Also, should all the the kwargs that we pass to `MultiZarrToZarr` and `ReferenceFileSystem` be packaged as mzz_kwargs and unpacked or separated?

norlandrhagen:
@cisaacstern good call!

Yeah, it seems like `remote_protocol` should be inferable from the `MultiZarrToZarr` object. 

I just pushed a commit that specifies the `remote_protocol` == 'file' upstream in `OpenWithKerchunk`. However in `write_combined_references` the `MZZ` object `reference.remote_protocol` returns None? So maybe the remote protocol is getting dropped somewhere or I'm missing something.

cisaacstern:
@norlandrhagen the `reference` received by `write_combine_references` is emitted by the [`CombineMultiZarrToZarr` here](https://github.com/norlandrhagen/pangeo-forge-recipes/blob/72b02c6e32c07b0529f628726a9a7739b6b876f2/pangeo_forge_recipes/transforms.py#L440C13-L446).

I suspect that the issue may have something to do with the (lack of?) remote protocol being passed explicitly (or otherwise handled) there.

The difficulties of implementing this do make me think it may be worth pausing to implement https://github.com/pangeo-forge/pangeo-forge-recipes/issues/623 first, which would give us a much better testbed.

pangeo_forge_recipes/writers.py
```diff
@@ -108,17 +108,21 @@ def write_combined_reference(
     full_target: FSSpecTarget,
     concat_dims: List[str],
     output_file_name: str,
+    # remote_protocol:  Optional[str],
+    remote_options: Optional[dict] = {"anon": True},
     refs_per_component: int = 1000,
 ) -> FSSpecTarget:
```
sbquinlan, line 114: nit: update return typehint

norlandrhagen:
Closing this in favor of PR: https://github.com/pangeo-forge/pangeo-forge-recipes/pull/660

Issue #645, opened 2023-11-06
Minio tests failing


cisaacstern:
We need a follow-on to #638 that restores the

```python
assert proc.returncode == 0
```
because both of the minio test cases are failing:

> https://github.com/pangeo-forge/pangeo-forge-recipes/actions/runs/6777276247/job/18420542826#step:7:32

> https://github.com/pangeo-forge/pangeo-forge-recipes/actions/runs/6777276247/job/18420542826#step:7:317

but without that assert they are registering to pytest as passes.

_Originally posted by @cisaacstern in https://github.com/pangeo-forge/pangeo-forge-recipes/issues/638#issuecomment-1796994135_

ranchodeluxe:
and we're good after that merge: https://github.com/pangeo-forge/pangeo-forge-recipes/actions/runs/6852898330/

PR #646, opened 2023-11-06
Disable `fsspec` Cache on Integration Tests

ranchodeluxe:
Addresses #645 

I can't find the deeper bug here to ticket with `zarr` or `fsspec` but potentially getting close

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -82,6 +82,7 @@ def minio_confpath(minio, tmp_path_factory: pytest.TempPathFactory):
     fsspec_args = {
         "key": minio["username"],
         "secret": minio["password"],
+        "use_listings_cache": False,
         "client_kwargs": {"endpoint_url": minio["endpoint"]},
     }
 
@@ -131,7 +132,4 @@ def test_integration(confpath_option: str, recipe_id: str, request):
         "JOB_NAME": f"{recipe_id}-{str(int(time.time()))}",
     }
     proc = subprocess.run(cmd, capture_output=True, env=env, text=True)
-    if proc.returncode != 0:
-        print("Command failed!")
-        print("Stdout:", proc.stdout)
-        print("Stderr:", proc.stderr)
+    assert proc.returncode == 0


ranchodeluxe:
Super fascinating bug. I don't think this is actually `minio` in the end (for reasons I will explain later). Wondering if this is something with `s3fs` 🤔

norlandrhagen:
Oof, well thanks for digging into it!

ranchodeluxe:
> Oof, well thanks for digging into it!

### Breakdown

Here's some evidence for it not being *just* `minio`:

1. I remove any dynamic paths [here](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/main/tests/test_integration.py#L81) and hardcode a reusable one using `pathlib.Path('/tmp')`

2. I comment out spinning up a `pytest` `minio` instnace and hardcode the secrets for my minio container running on my host (see next step) [here](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/main/tests/test_integration.py#L41)

3. I then run a dedicated `minio` container on my host:
```bash
docker run \
   -p 9000:9000 
   --name minio666 \
   -e "MINIO_ACCESS_KEY=YOURACCESSKEY" \
   -e "MINIO_SECRET_KEY=YOURSECRETKEY" \
   -v /tmp/:/data \                                  
   minio/minio server /data 
```

4. I can ssh into that container and trace requests using:
```bash
mc alias set 'myminio' 'http://localhost:9000' 'YOURACCESSKEY' 'YOURSECRETKEY'
mc admin trace -v -a myminio
```

6. Since the problem is intermittent I wait for  a successful run to happen and check outputs on `minio` using `mc ls <target>`. If everything looks good I know the next runs should ALWAYS succeed b/c the `lat` key (for the `noaa-oisst` test specifically) exists already and we are reusing the hardcoded target from step 1 above for the next run

7. More runs either succeed and fail intermittently even with this set up (`pytest -s tests/test_integration.py -k 'test_integration[minio_confpath-noaa-oisst]' --run-integration --timeout=500`)

8. There is nothing in the verbose `mc trace` logging that shows bad `PUT`(s) or `500`(s) or anything

9. My hunch is if this was just `minio` we wouldn't be seeing the tests failing constantly for exactly the same key (which is `lat` for `noaa-oisst` test) but different keys across runs. I'm not seeing that

### Proposals

* Is there a reason these tests can't use an actually dedicated  s3 or google storage bucket? Tests run fine against them

* I imagine we don't want to use `fsspec.implementations.local.LocalFileSystem` for the tests b/c it's not really flexing what this project is about?

* If we truly think it's just `minio` a possible work around here that will take more time is to do what `pangeo-forge-runner` does and set up a k3's instance and use `minio` there. Thoughts? 🤔

ranchodeluxe:
> Proposals
> 
>     Is there a reason these tests can't use an actually dedicated s3 or google storage bucket? Tests run fine against them
> 
>     I imagine we don't want to use fsspec.implementations.local.LocalFileSystem for the tests b/c it's not really flexing what this project is about?
> 
>     If we truly think it's just minio a possible work around here that will take more time is to do what pangeo-forge-runner does and set up a k3's instance and use minio there. But if it's not just minio it's all for nothing. Thoughts? 🤔

well, if I run `minio` as a service in k8s locally everything works fine so 🤷 looks like that's the way forward. won't get to that today though

ranchodeluxe:
wowzers, things just got more interesting [with these failures](https://github.com/pangeo-forge/pangeo-forge-recipes/actions/runs/6815908965/job/18536039496?pr=646)

can't get it to fail locally with a similar set up

looks like we cannot ignore the subtle interaction that is causing this and have to weed it out

ranchodeluxe:
It was s3fs the whole time!!!!!!! I had to disable cache settings and now everything works 😞

norlandrhagen (approved):
LGTM!

PR #651, opened 2023-11-15
Add pytest-xdist to speed up tests

jbusecke:
An attempt to speed up the extremely long-running tests.


norlandrhagen:
Didn't know about pytest-xdist. Seems very cool!

jbusecke:
works even better locally when you have more cores available 😃

PR #654, opened 2023-11-17
Add ConsolidateMetadata transform, making it compatible with kerchunk

sbquinlan:


cisaacstern:
Thanks for kicking this off, @sbquinlan!

How do you propose harmonizing this work with #556?

Open to any suggestions but the easiest thing might be to finish #556 and then put this in on top of it?

sbquinlan:
> How do you propose harmonizing this work with https://github.com/pangeo-forge/pangeo-forge-recipes/pull/556? 

@cisaacstern Oh thanks for pointing that one out! I didn't realize it was already being worked on. I'll leave my feedback on that PR, but at a high level it's "we should provide composable transforms with a common interface". For this change it means not doing boolean configuration flags on a central transform (ie "StoreToZarr") and we should change the typing on `ConsolidateMetadata` so that it works on both Kerchunk and Zarr datastores.

sbquinlan:
Also keep in mind this is a draft and there's still some issues that I need to fix.

cisaacstern:
>  I'll leave my feedback on that PR

Awesome, thanks!

> but at a high level it's "we should provide composable transforms with a common interface". For this change it means not doing boolean configuration flags on a central transform (ie "StoreToZarr")

Yes that's a very good criticism, and I fully agree. I am to blame for the existing presence of boolean config flags & inconsistent return types, both in the linked PR and in `OpenWithKerchunk` here:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/32d3864a596322b1bf0c135c69a3e2b896fcb96a/pangeo_forge_recipes/transforms.py#L242

By way of background, these boolean flags came up as a way to keep recipe composition as simple as possible for basic use cases. But I agree they hinder composability so we should probably do away with this approach.

> and we should change the typing on ConsolidateMetadata so that it works on both Kerchunk and Zarr datastores.

💯 agree, great!

sbquinlan:
This is ready but it needs a version of both #636 and #556 so I'll leave it in draft for now.

norlandrhagen:
Hey @sbquinlan, wondering if you have plans to do any more work off of this PR or do you mind if I wrap up your work on it?

Issue #656, opened 2023-11-28
Output target chunks to logs?


jbusecke:
With the additions of the dynamic chunking function I think it would be great to add some logging that shows the target chunks that are actually passed internally. 
I believe that will make debugging recipes on e.g. Dataflow much easier.

cisaacstern:
Makes total sense, and in general we are too thin on logs.

As I think we've discussed, logging seems to need to be at INFO level or above to get picked up by Dataflow.

PRs welcome! 😄

PR #657, opened 2023-11-29
Add logging to StoreToZarr

jbusecke:
#656 
I also added the full target here, because that might be helpful to have?

### DIFF
--- a/pangeo_forge_recipes/transforms.py
+++ b/pangeo_forge_recipes/transforms.py
@@ -547,7 +547,7 @@ class StoreToZarr(beam.PTransform, ZarrWriterMixin):
                 | beam.Map(self.dynamic_chunking_fn, **self.dynamic_chunking_fn_kwargs)
             )
         )
-        logger.warn(f"Storing Zarr with {target_chunks =} to {self.get_full_target()}")
+        logger.info(f"Storing Zarr with {target_chunks =} to {self.get_full_target()}")
         rechunked_datasets = indexed_datasets | Rechunk(target_chunks=target_chunks, schema=schema)
         target_store = schema | PrepareZarrTarget(
             target=self.get_full_target(),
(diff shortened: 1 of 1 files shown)


pangeo_forge_recipes/transforms.py
```diff
@@ -547,6 +547,7 @@ def expand(
                 | beam.Map(self.dynamic_chunking_fn, **self.dynamic_chunking_fn_kwargs)
             )
         )
+        logger.warn(f"Storing Zarr with {target_chunks =} to {self.get_full_target()}")
```
cisaacstern, line 550:
```suggestion
        logger.info(f"Storing Zarr with {target_chunks =} to {self.get_full_target()}")
```
Seems like this is `info`, since nothing potentially adverse (i.e. warn-worthy) is happening?

cisaacstern:
@jbusecke have you confirmed anecdotally by running this that the logs write what you expect/desire them to?

jbusecke:
Hmmm not really. I havent run things on a PR branch in a bit lol. Let me check this, BRB

jbusecke:
Testing over at https://github.com/leap-stc/cmip6-leap-feedstock/pull/79

jbusecke:
Hmmmm so over at https://github.com/leap-stc/cmip6-leap-feedstock/pull/79 I specified to install -recipes from [this branch](https://github.com/leap-stc/cmip6-leap-feedstock/pull/79/files), and the [install seems to work](https://github.com/leap-stc/cmip6-leap-feedstock/actions/runs/7214078871/job/19655377315?pr=79#step:5:32).

But for e.g. one of [the successful jobs](https://console.cloud.google.com/dataflow/jobs/us-central1/2023-12-14_12_17_09-7089901999214371794;logsSeverity=INFO;graphView=0?project=leap-pangeo&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22))) im not able to find anything matching `Storing Zarr with`. 

So either this does not work, or I am a moron with the logs (highly likely). @cisaacstern if you have a minute, could you take a stab at the logs (you should have access) and confirm my finding?

moradology:
I'm curious about the way this was run. If it was done in a local runner with multi_processing, I'd expect logging to be swallowed on executor processes even if this is defined correctly. Either way, the logging is written correctly here and if there are more general logging difficulties we should address them elsewhere. Going to merge this in as it is incredibly unlikely to cause trouble for us

PR #660, opened 2023-12-05
Kerchunk WriteCombinedReference return fsstore

norlandrhagen:
This PR is a clean-slate version of this PR: https://github.com/pangeo-forge/pangeo-forge-recipes/pull/636
It borrows heavily from @sbquinlan's excellent work in this PR. https://github.com/pangeo-forge/pangeo-forge-recipes/pull/654. Hopefully enabling that PR to proceed. 

- `test_CombineReferences` is now passing
- The integration test that tries hrrr_kerchunk_concat_step is now passing and the xfail is removed.

### DIFF
--- a/.github/workflows/test-integration.yaml
+++ b/.github/workflows/test-integration.yaml
@@ -56,10 +56,14 @@ jobs:
       - name: set up conda environment
         uses: mamba-org/setup-micromamba@v1
         with:
+          environment-name: pgf
+          create-args: >-
+            cfgrib
           init-shell: >-
             bash
           cache-environment: true
           cache-downloads: true
+          
           post-cleanup: 'all'
 
       - name: 🎯 Check cache hit
@@ -70,11 +74,11 @@ jobs:
           python -m pip install ${{ matrix.runner-version }}
           python -m pip install -e ".[test,minio]"
 
-      # - name: Install cfgrib deps
-      #   shell: bash -l {0}
-      #   run: |
-      #     sudo apt-get install libeccodes-dev
-      #     python -m pip install cfgrib
+      - name: Install cfgrib deps
+        shell: bash -l {0}
+        run: |
+          sudo apt-get install libeccodes-dev
+          python -m pip install cfgrib
 
 
         # order reversed to fix https://github.com/pangeo-forge/pangeo-forge-recipes/pull/595#issuecomment-1811630921
--- a/.github/workflows/test-integration.yaml
+++ b/.github/workflows/test-integration.yaml
@@ -36,13 +36,16 @@ jobs:
         ]
     steps:
       - uses: actions/checkout@v4
-      - name: 🔁 Setup Python
-        id: setup-python
-        uses: actions/setup-python@v4
+
+      - name: Setup conda
+        uses: mamba-org/setup-micromamba@v1
         with:
-          python-version: ${{ matrix.python-version }}
-          cache: pip
-          cache-dependency-path: pyproject.toml
+          environment-file: ${{ matrix.python-version }}
+      - name: Install kerchunk
+        shell: bash -l {0}
+        run: |
+          python -m pip install ${{ matrix.runner-version }}
+          python -m pip install -e ".[test,minio]"
 
       - name: 'Setup minio'
         run: |
@@ -52,34 +55,8 @@ jobs:
 
           minio --version
 
-
-      - name: set up conda environment
-        uses: mamba-org/setup-micromamba@v1
-        with:
-          environment-name: pgf
-          create-args: >-
-            cfgrib
-          init-shell: >-
-            bash
-          cache-environment: true
-          cache-downloads: true
-          
-          post-cleanup: 'all'
-
       - name: 🎯 Check cache hit
         run: echo '${{ steps.setup-python.outputs.cache-hit }}'
-      - name: 🌈 Install pangeo-forge-recipes & pangeo-forge-runner
-        shell: bash -l {0}
-        run: |
-          python -m pip install ${{ matrix.runner-version }}
-          python -m pip install -e ".[test,minio]"
-
-      - name: Install cfgrib deps
-        shell: bash -l {0}
-        run: |
-          sudo apt-get install libeccodes-dev
-          python -m pip install cfgrib
-
 
         # order reversed to fix https://github.com/pangeo-forge/pangeo-forge-recipes/pull/595#issuecomment-1811630921
         # this should however be fixed in the runner itself
--- a/.github/workflows/test-integration.yaml
+++ b/.github/workflows/test-integration.yaml
@@ -41,7 +41,7 @@ jobs:
         uses: mamba-org/setup-micromamba@v1
         with:
           create-args: python==${{ matrix.python-version }}
-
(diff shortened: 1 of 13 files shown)


pangeo_forge_recipes/transforms.py
```diff
@@ -446,6 +447,33 @@ def expand(self, references: beam.PCollection) -> beam.PCollection:
...
+    :param concat_dims: Dimensions along which to concatenate inputs.
+    :param target_root: Root path the Zarr store will be created inside; ``store_name``
+      will be appended to this prefix to create a full path.
+    :param output_file_name: Name to give the output references file
+      (``.json`` or ``.parquet`` suffix).
+    """
+
+    store_name: str
+    concat_dims: List[str]
+    target_root: Union[str, FSSpecTarget, RequiredAtRuntimeDefault] = field(
```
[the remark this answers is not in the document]
norlandrhagen: Great catch! I updated injections.py to add `WriteReference`. Any idea if any more is required for that?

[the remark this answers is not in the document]
cisaacstern: Can take a closer look tomorrow but off the cuff I wonder if the failure is because we need to bump the version of runner used in the integration tests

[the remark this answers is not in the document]
norlandrhagen: I bumped runner to the two latest tags. First kerchunk integration test is passing! 
Getting a failure on the minio version though. 


```
tests/test_integration.py::test_python_json_configs_identical PASSED
tests/test_integration.py::test_integration[local_confpath-hrrr-kerchunk-concat-valid-time] XFAIL
tests/test_integration.py::test_integration[local_confpath-hrrr-kerchunk-concat-step] PASSED
tests/test_integration.py::test_integration[local_confpath-gpcp-from-gcs-dynamic-chunks] PASSED
tests/test_integration.py::test_integration[local_confpath-terraclimate] XFAIL
tests/test_integration.py::test_integration[local_confpath-narr-opendap] XFAIL
tests/test_integration.py::test_integration[local_confpath-noaa-oisst] PASSED
tests/test_integration.py::test_integration[local_confpath-gpcp-from-gcs] PASSED
tests/test_integration.py::test_integration[minio_confpath-hrrr-kerchunk-concat-valid-time] XFAIL
tests/test_integration.py::test_integration[minio_confpath-hrrr-kerchunk-concat-step] FAILED
```

```
raise NoCredentialsError()\nRuntimeError: botocore.exceptions.NoCredentialsError: Unable to locate credentials [while running \'Create|OpenWithKerchunk|WriteCombinedReference|Test dataset/WriteCombinedReference/WriteReference/Map(write_combined_reference)\']\n\n').returncode
```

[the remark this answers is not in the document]
cisaacstern: This is actually great, insofar as it proves the value of the `minio` test! So much better to catch this here, rather than merge this and only find out in prod! 😃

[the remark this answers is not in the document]
norlandrhagen: 100%

norlandrhagen:
@sbquinlan @cisaacstern got the tests passing! If either one of you would mind taking one last look at the PR, it would be appreciated.

.github/workflows/test-integration.yaml
```diff
@@ -31,18 +31,23 @@ jobs:
...
-          cache: pip
-          cache-dependency-path: pyproject.toml
+          create-args: python==${{ matrix.python-version }}
+          environment-name: pangeo-forge
+
+      - name: Install pangeo-forge recipes and runner
+        shell: bash -l {0}
+        run: |
+          python -m pip install ${{ matrix.runner-version }}
+          python -m pip install -e ".[test,minio,grib]"
```
cisaacstern, line 50: If we are using pip to install GRIB dependencies, do we need to be using micro mamba? Can't we just keep Setup Python the way it was?

cisaacstern: Also, I'm somewhat surprised this works? The `cf-grib`/`eccodes` docs seems to suggest system-level dependencies need to be installed via conda?

.github/workflows/test-integration.yaml
```diff
@@ -54,11 +61,6 @@ jobs:
 
       - name: 🎯 Check cache hit
         run: echo '${{ steps.setup-python.outputs.cache-hit }}'
-      - name: 🌈 Install pangeo-forge-recipes & pangeo-forge-runner
-        shell: bash -l {0}
-        run: |
-          python -m pip install ${{ matrix.runner-version }}
-          python -m pip install -e ".[test,minio]"
```
cisaacstern, line 61: Sorry for the piecemeal review, @norlandrhagen.

Since we're not doing anything different re: pip, can we just preserve this block for a cleaner diff?

I would prefer to _not_ support a user-facing `grib` optional dependency (at least as part of this PR), as I'm concerned that without putting that in a conda environment we can really ensure that will work reliably in a cross-platform way.

So I'd prefer just a separate workflow step here called "install grib deps" or similar, and not to have those deps as part of our pyproject.toml.

pyproject.toml
```diff
@@ -51,6 +51,8 @@ test = [
     "s3fs",
     "scipy",
 ]
+grib = ["ecmwflibs", "eccodes", "cfgrib"]
+
```
cisaacstern, line 55:
```suggestion
```

See comment on test-integration.yaml

cisaacstern (commented):
Another comment about how we're bringing in cfgrib for the test.

.github/workflows/test-integration.yaml
norlandrhagen: Good call, just reverted that and the integration tests are passing.

.github/workflows/test-integration.yaml
norlandrhagen: Seems like a good call!

cisaacstern:
Thanks for pushing this forward, @norlandrhagen, clearly we're very close.

My comments thus far are on packaging/GH workflows.

I haven't really digested the changes to the actual code. If @sbquinlan is able to take a look at that, great! If not, I can try to circle back to that section tomorrow.

.github/workflows/test-integration.yaml
norlandrhagen: If I understand.. I removed the grib deps from pyproject, reverted the install to`python -m pip install -e ".[test,minio]"` and added another step in the CI to install the grib deps via pip.

norlandrhagen:
Thanks for the CI feedback so far!

norlandrhagen:
Thanks for the review @sbquinlan !

pangeo_forge_recipes/transforms.py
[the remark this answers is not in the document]
ranchodeluxe: I just rediscovered this bug since I didn't see the need to pass `remote_options` or `target_options` given we had dep injected `target_root`.

So the breaking change between `0.10.4` and what this PR merged into `main` is us removing [this context manager](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/0.10.4/pangeo_forge_recipes/writers.py#L123)

PR #673, opened 2024-01-22
try to remove `conftest.py:tmp_target_url`

ranchodeluxe:
~~IMHO our tests should be using `FSSpecTarget` but I'm probably wrong 😆 So we'll see~~

easier than expected 😎


ranchodeluxe:
gonna fold this into my existing PR so closing now

PR #676, opened 2024-01-25
WIP: Failing Test Case

ranchodeluxe:


ranchodeluxe:
Tests are now not raising `KeyError` b/c `xr.open_dataset` can open the `mapper` of the `ReferenceFileSystem`. My hunch is the `zarr.hierarchy.Group` returned from `zarr.conslidate_metadata` isn't suitable to opened by `xarray` but it's not that complicated of a  `zarr.hierarchy.Group`(no nested groups). 

So more investigation needed

ranchodeluxe:
closing to work on Raphael's new branch

Issue #675, opened 2024-01-25
Xarray Cannot Read `ConsolidateMetadata` Reference Output


ranchodeluxe:
### Versions:

zarr.__version__: `'2.16.1'`
xarray.__version__: `'2023.12.0'`
pangeo-forge-recipes: `@main`

### Problem:

Given a `WriteCombinedReference` beam pipeline like this:

```python
    with pipeline as p:
        (
            p
            | beam.Create(pattern.items())
            | OpenWithKerchunk(file_type=pattern.file_type)
            | WriteCombinedReference()
            | ConsolidateMetadata()
        )
```

we can only read the reference output with `zarr.open_consolidated` but not `xr.open_dataset(..., consolidated=True)`

```python
(Pdb) xr.open_dataset(store, engine='zarr', consolidated=True)
*** KeyError: '.zmetadata'
```

indeed looking at the raw `reference.json` it doesn't contain a `.zmetadata` key even though `zarr.consolidate_metadata` should be writing one.  Totaly guess, but  might have to do with the version branching here: https://github.com/zarr-developers/zarr-python/blob/main/zarr/convenience.py#L1250-L1263

abarciauskas-bgse:
I also don't know if this is helpful but I'm a bit confused as well how zarr.consolidate_metadata is intended to or ever was working. I was also looking at convenience.py and noticed that `consolidate_metadata` returns `open_consolidated` and `open_consolidated` uses, by default an `r+` mode which has a docstring:

>'r+' means
        read/write (must exist) although only writes to data are allowed,
        changes to metadata including creation of new arrays or group
        are not allowed.

Which stuck out to me because this method is intended to add a metadata group, I think.

Overall, I'm not seeing where the `.zmetadata` key would be written to disk, either as its own file or into the reference json file. But I'm probably missing something?

maxrjones:
@martindurant suggested in https://github.com/fsspec/kerchunk/issues/240#issuecomment-1305729813 adding consolidation in MultiZarrtoZarr, if I understand correctly. In general, I haven't seen examples of Zarr Python being used to modify kerchunk reference files, so I'm wondering if that functionality would need to be provided by kerchunk?

rabernat:
> I also don't know if this is helpful but I'm a bit confused as well how zarr.consolidate_metadata is intended to or ever was working.

You can have a look at how Xarray does it.
[...]

PR #678, opened 2024-01-26
WIP - [BugFix Attempt] ConsolidateMetadata 

norlandrhagen:
`ConsolidateMetadata` doesn't seem to be creating a `.zmetadata` in the reference recipe tests.

Based on Martin Durants comments here: https://github.com/fsspec/kerchunk/issues/240#issuecomment-1290739134, this PR updates how the `.zmetadata` is being created.

Some open questions:

- It seemed like the previous tests of `zarr.open_consolidated` were working...without a .zmetadata file?
- This method `path.fs.save_json(ref_path)` calls **save_json`**. How do we modify the parquet stores?

cc @abarciauskas-bgse @ranchodeluxe

### DIFF
--- a/pangeo_forge_recipes/writers.py
+++ b/pangeo_forge_recipes/writers.py
@@ -84,8 +84,10 @@ def consolidate_metadata(store: MutableMapping) -> MutableMapping:
     if isinstance(store, zarr.storage.FSStore):
         path = store.path
 
-    zc = zarr.consolidate_metadata(path)
-    return zc
+    zarr.convenience.consolidate_metadata(path)
+    # How do we update a parquet reference file?
+    path.fs.save_json(ref_path)
+    return path
 
 
 def store_dataset_fragment(
--- a/tests/test_writers.py
+++ b/tests/test_writers.py
@@ -182,7 +182,6 @@ def test_reference_netcdf(
     netcdf_local_file_pattern_sequential,
     pipeline,
     tmp_target,
-    # why are we not using tmp_target?
     output_file_name,
 ):
     pattern = netcdf_local_file_pattern_sequential
@@ -203,6 +202,6 @@ def test_reference_netcdf(
         )
 
     full_path = os.path.join(tmp_target.root_path, store_name, output_file_name)
-
     mapper = fsspec.get_mapper("reference://", fo=full_path)
     assert zarr.open_consolidated(mapper)
+    assert xr.open_zarr(mapper, consolidated=True)


abarciauskas-bgse:
Thanks for looking at this @norlandrhagen ! I was just reading about the parquet implementation and noticed it says:

>For now, [kerchunk.combine.MultiZarrToZarr](https://fsspec.github.io/kerchunk/reference.html#kerchunk.combine.MultiZarrToZarr) only operates on JSON/dict input. Therefore, refs_to_dataframe can only be used on the final output reference set.

Since consolidate_metadata also seems to assume JSON/dict input I would not expect to be able to pass geoparquet to consolidate_metadata either.

Did you consolidate metadata before converting to geoparquet for the CMIP6 dataset you worked on?

norlandrhagen:
Update:

Thanks @abarciauskas-bgse & @ranchodeluxe for talking through this! 

According to discussion in this PR: https://github.com/pangeo-forge/pangeo-forge-recipes/issues/675. It seems like there isn't really a point to consolidate the metadata for a Kerchunk reference. 

- We should probably remove any Kerchunk specific logic for consolidating references. 
- Consolidate_Metadata works on Kerchunk json references (Although it's probably not needed at all)
- ConsolidateMetadata fails on Kerchunk parquet references.

ranchodeluxe:
> Update:
> 
> Thanks @abarciauskas-bgse & @ranchodeluxe for talking through this!
> 
> According to discussion in this PR: #675. It seems like there isn't really a point to consolidate the metadata for a Kerchunk reference.
> 
>     * We should probably remove any Kerchunk specific logic for consolidating references.
> 
>     * Consolidate_Metadata works on Kerchunk json references (Although it's probably not needed at all)
> 
>     * ConsolidateMetadata fails on Kerchunk parquet references.

(just adding these bullets also)

* Include the fix for `write_combined_reference`

* Make sure ConsolidateMetadata takes `target_root` and passes along fsspec kwargs

ranchodeluxe (approved):
Thanks for doing this @norlandrhagen 🥳  Can't wait to get this merged so I can double check if my assumption about fsspec auth args is correct or not

norlandrhagen:
@cisaacstern would you mind taking a look at this when you get a chance?

pangeo_forge_recipes/writers.py
```diff
@@ -79,12 +79,16 @@ def consolidate_metadata(store: MutableMapping) -> MutableMapping:
     import zarr
 
     if isinstance(store, fsspec.FSMap) and isinstance(store.fs, ReferenceFileSystem):
-        ref_path = store.fs.storage_args[0]
-        path = fsspec.get_mapper("reference://", fo=ref_path)
+        raise ValueError(
+            "Creating consolidated metadata for Kerchunk references has"
+            " no preformance benefits: Issue: "
+            " https://github.com/pangeo-forge/pangeo-forge-recipes/issues/675"
```
abarciauskas-bgse, line 85: Should we link to this issue here? It doesn't really explain or prove that there are no performance benefits.

I did a quick test (see this gist https://gist.github.com/abarciauskas-bgse/4cb32678f11aa04bcb281517a5e1862c) to see if I could tell if consolidating metadata mutates the store in any way and if there is performance benefits in that case. This was not an extensive test by any means but what I found was:

1. The json is the same between the two datasets except the consolidated version places all the zarr data in a `refs` key under the root, as opposed to directly under the root and the `refs` value holds all the data with an additional `.zmetadata` key which duplicates the data under the `.zarray` of the root object in the unconsolidated store.
2. Just as a sanity check and in case the libraries read the data differently when `consolidated=True` vs `consolidate=False`, I ran those performance tests a few times and, as noted in the notebook, the first time you generate the time series always takes the longest (4-5 seconds), regardless of whether the metadata is consolidated or not. Re-running the time series generation takes 1-1.5 seconds regardless as well.

Like I said, this was not a very exhaustive test but I don't think we have any reason to believe that performance would be different as the main point of consolidating metadata is just to read all the zarr metadata from one file which is already satisfied with the kerchunk reference. kerchunk is basically already consolidated metadata, but alongside the data itself (in a sense).

norlandrhagen: Good call! I'll remove the link to the issue. Thanks for digging into this @abarciauskas-bgse.

PR #687, opened 2024-02-06
Return FSStore from consolidate_metadata

sbquinlan:
It's more convienent to return an FSStore than to open the store, because most of the other provided utilites also return and/or take an FSStore. This just simply removes the call to open_consolidated and returns the FSStore.


moradology:
Looks like this change made it in elsewhere

PR #688, opened 2024-02-07
Pluralize items in CombineMultiZarrToZarr#add_input

abarciauskas-bgse:

PR #689, opened 2024-02-08
Fix CombineReference merge ordering

moradology:
precombine_inputs is kind of awkward and it is really only used (so far as I can tell) to deal with the fact that GRIB2 reads will produce a list of ZARR refs. This could be simpler. ~We could just flatmap over the list of lists, producing a list of references that can be accumulated, shipped to the driver, and then combined via MultiZarrToZarr all at once.~

FlatMap is gone, as it is convenient to keep grib references grouped under their key until `CombineReferences` is called. Instead, MapTuple is used and we hang onto keys. Keys are used to construct batches that keep neighboring data on the same worker and merging is handled much more consciously than is possible in a `beam.CombineGlobally`

~This also happens to be much closer to the strategy described as 'tree reduction' in the kerchunk docs and implemented in their auto_dask method so maybe it will resolve some *very weird* bugs that currently appear to be associated with the (bad) assumption that MultiZarrToZarr merger has the properties of associativity and idempotence. Maybe 🤞?~
We now have testing (courtesy of @abarciauskas-bgse) which verifies correctness.

This also supersedes #688 as it resolves the item pluralization issue @abarciauskas-bgse identified
Incorporates #691

### DIFF
--- a/pangeo_forge_recipes/combiners.py
+++ b/pangeo_forge_recipes/combiners.py
@@ -1,11 +1,10 @@
 import operator
-from dataclasses import dataclass, field
+import sys
+from dataclasses import dataclass
 from functools import reduce
-from typing import Dict, List, Optional, Sequence
+from typing import Sequence
 
 import apache_beam as beam
-import fsspec
-from kerchunk.combine import MultiZarrToZarr
 
 from .aggregation import XarrayCombineAccumulator, XarraySchema
 from .types import CombineOp, Dimension, Index, Indexed
@@ -46,54 +45,32 @@ class CombineXarraySchemas(beam.CombineFn):
         return accumulator.schema
 
 
-@dataclass
-class CombineZarrRefs(beam.CombineFn):
-    """A beam ``CombineFn`` for combining Kerchunk reference objects.
+def build_reduce_fn(accumulate_op, merge_op, initializer):
+    """Factory to construct reducers without so much ceremony"""
 
-    :param concat_dims: Dimensions along which to concatenate inputs.
-    :param identical_dims: Dimensions shared among all inputs.
-    :param remote_options: Storage options for opening remote files
-    :param remote_protocol: If files are accessed over the network, provide the remote protocol
-      over which they are accessed. e.g.: "s3", "gcp", "https", etc.
-    :mzz_kwargs: Additional kwargs to pass to ``kerchunk.combine.MultiZarrToZarr``.
-    :param target_options: Target options dict to pass to the MultiZarrToZarr
+    class AnonymousCombineFn(beam.CombineFn):
+        def create_accumulator(self):
+            return initializer
 
-    """
+        def add_input(self, accumulator, input):
+            return accumulate_op(accumulator, input)
 
-    concat_dims: List[str]
-    identical_dims: List[str]
-    target_options: Optional[Dict] = field(default_factory=lambda: {"anon": True})
-    remote_options: Optional[Dict] = field(default_factory=lambda: {"anon": True})
-    remote_protocol: Optional[str] = None
-    mzz_kwargs: dict = field(default_factory=dict)
+        def merge_accumulators(self, accumulators):
+            acc = accumulators[0]
+            for accumulator in accumulators[1:]:
+                acc = merge_op(acc, accumulator)
+            return acc
 
-    def to_mzz(self, references):
-        return MultiZarrToZarr(
-            references,
-            concat_dims=self.concat_dims,
-            identical_dims=self.identical_dims,
-            target_options=self.target_options,
-            remote_options=self.remote_options,
-            remote_protocol=self.remote_protocol,
-            **self.mzz_kwargs,
-        )
+        def extract_output(self, accumulator):
+            return accumulator
 
-    def create_accumulator(self) -> list[dict]:
-        return []
+    return AnonymousCombineFn
 
-    def add_input(self, accumulator: list[dict], item: dict) -> list[dict]:
-        accumulator.append(item)
-        return accumulator
 
-    def merge_accumulators(self, accumulators: list[list[dict]]) -> list[dict]:
-        return [self.to_mzz(accumulator).translate() for accumulator in accumulators]
-
-    def extract_output(self, accumulator: list[dict]) -> fsspec.FSMap:
-        return fsspec.filesystem(
-            "reference",
-            fo=self.to_mzz(accumulator).translate(),
-            storage_options={
-                "remote_protocol": self.remote_protocol,
-                "skip_instance_cache": True,
-            },
-        ).get_mapper()
+# Find minimum/maximum/count values
+# Count done as a slight optimization to avoid multiple passes across the distribution
+MinMaxCountCombineFn = build_reduce_fn(
+    accumulate_op=lambda acc, input: (min(acc[0], input), max(acc[1], input), acc[2] + 1),
+    merge_op=lambda accL, accR: (min(accL[0], accR[0]), max(accL[1], accR[1]), accL[2] + accR[2]),
+    initializer=(sys.maxsize, -sys.maxsize, 0),
+)
--- a/pangeo_forge_recipes/combiners.py
+++ b/pangeo_forge_recipes/combiners.py
@@ -56,18 +56,6 @@ class CombineZarrRefs(beam.CombineFn):
     :param remote_protocol: If files are accessed over the network, provide the remote protocol
       over which they are accessed. e.g.: "s3", "gcp", "https", etc.
     :mzz_kwargs: Additional kwargs to pass to ``kerchunk.combine.MultiZarrToZarr``.
-    TODO: precombine_inputs is very adhoc here, perhaps should be resolved in other ways
-        flatmapping at a prior stage resolves the awkward list[dict] item type, so absolves
-        GRIB2 inputs of their need for this
-    :precombine_inputs: If ``True``, precombine each input with itself, using
-      ``kerchunk.combine.MultiZarrToZarr``, before adding it to the accumulator.
-      Used for multi-message GRIB2 inputs, which produce > 1 reference when opened
-      with kerchunk's ``scan_grib`` function, and therefore need to be consolidated
-      into a single reference before adding to the accumulator. Also used for inputs
-      consisting of single reference, for cases where the output dataset concatenates
-      along a dimension that does not exist in the individual inputs. In this latter
-      case, precombining adds the additional dimension to the input so that its
-      dimensionality will match that of the accumulator.
     :param target_options: Target options dict to pass to the MultiZarrToZarr
 
     """
--- a/pangeo_forge_recipes/combiners.py
+++ b/pangeo_forge_recipes/combiners.py
@@ -98,7 +98,7 @@ class CombineZarrRefs(beam.CombineFn):
         return accumulator
 
     def merge_accumulators(self, accumulators: list[list[dict]]) -> list[dict]:
-        return [item for accumulator in accumulators for item in accumulator]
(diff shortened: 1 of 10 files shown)


moradology:
Converted to a draft for now, pending a bit of discussion. I feel pretty confident that the path I'm on is superior on the assumption that there's not something important I don't understand!

moradology:
@cisaacstern Awesome. I think we've verified that this does work. Now to produce some meaningful tests that demonstrate failure and the current path's success. I always find it hard not to second-guess when the elephant is so big and I can touch so little of it at any one time 😅

ranchodeluxe:
Does this feel ready @moradology? @abarciauskas-bgse wrote a nice test case in https://github.com/pangeo-forge/pangeo-forge-recipes/pull/692 that I think we should add to this PR

pangeo_forge_recipes/transforms.py
```diff
@@ -443,37 +435,89 @@ class CombineReferences(beam.PTransform):
...
+            reference_lists
+            | "Handle special case of gribs" >> beam.Map(self.handle_gribs)
+            | "Bucket to preserve order"
+            >> beam.Map(
+                self.bucket_by_position,
+                global_position_min_max_count=beam.pvalue.AsSingleton(min_max_count_positions),
+            )
+            | "Group by buckets for ordering" >> beam.GroupByKey()
+            | "Distributed reduce" >> beam.MapTuple(lambda k, refs: self.to_mzz(refs).translate())
+            | "Assign global key for collecting to executor" >> beam.Map(lambda ref: (None, ref))
```
ranchodeluxe, line 518: As we talked about: our assumption is this will route all work to one worker for the final reduce but we need to think (across runners) how we'd identify and right-size that worker in terms of resources to handle this. So I'll need to find what Flink says here

pangeo_forge_recipes/combiners.py
```diff
@@ -46,69 +45,32 @@ def extract_output(self, accumulator) -> dict:
...
-      ``kerchunk.combine.MultiZarrToZarr``, before adding it to the accumulator.
-      Used for multi-message GRIB2 inputs, which produce > 1 reference when opened
-      with kerchunk's ``scan_grib`` function, and therefore need to be consolidated
-      into a single reference before adding to the accumulator. Also used for inputs
-      consisting of single reference, for cases where the output dataset concatenates
-      along a dimension that does not exist in the individual inputs. In this latter
-      case, precombining adds the additional dimension to the input so that its
-      dimensionality will match that of the accumulator.
-    :param target_options: Target options dict to pass to the MultiZarrToZarr
+def build_reduce_fn(accumulate_op, merge_op, initializer):
```
abarciauskas-bgse, line 48: should these arguments be typed?

pangeo_forge_recipes/transforms.py
```diff
@@ -443,37 +435,89 @@ class CombineReferences(beam.PTransform):
...
+            concat_dims=self.concat_dims,
+            identical_dims=self.identical_dims,
+            target_options=self.target_options,
+            remote_options=self.remote_options,
+            remote_protocol=self.remote_protocol,
+            **self.mzz_kwargs,
+        )
+
+    def handle_gribs(self, indexed_references: Tuple[Index, list[dict]]) -> Tuple[Index, dict]:
+        """Handles the special case of GRIB format files by combining multiple references."""
```
abarciauskas-bgse, line 463: 👍🏽

pangeo_forge_recipes/combiners.py
moradology: 100%

ranchodeluxe (changes requested):
Nice 🥳 Looked over the code again (especially `CombineReferences`) and have the following comments:

0. I like how `CombineReferences.expand` got much more readable and clear on it's steps
1. `CombineReferences.bucket_by_position` is still opague to me even though I know what it's attempting to do

~~I'd request we write some `parametrize` tests to explain the inputs and outputs for `bucket_by_position` clearly and use one of those in the docs string for so folks can understand it immediately. Though now that I mention it `bucket_by_position` is hard to test by itself b/c it's a class function~~ (this PR https://github.com/pangeo-forge/pangeo-forge-recipes/pull/697 shows how we can do it as a class function fine)

moradology:
> Though now that I mention it `bucket_by_position` is hard to test by itself b/c it's a class function

Should be able to break it out for testing. May come in handy again?

ranchodeluxe:
> > Though now that I mention it `bucket_by_position` is hard to test by itself b/c it's a class function
> 
> Should be able to break it out for testing. May come in handy again?

Ha, spoke too soon and took a quick stab at it myself to help elucidate my understanding @ `gcorradini/npz/feature/simplify_zarr_merge`. Let's see what passes

pangeo_forge_recipes/transforms.py
```diff
@@ -443,37 +435,89 @@ class CombineReferences(beam.PTransform):
...
+        else:
+            raise ValueError("No references produced for {idx}. Expected at least 1.")
+
+    def bucket_by_position(
+        self, indexed_references: Tuple[Index, dict], global_position_min_max_count
+    ) -> Tuple[int, dict]:
+        """Assigns a bucket based on index position to help preserve data order."""
+        idx = indexed_references[0]
+        ref = indexed_references[1]
+        global_min, global_max, global_count = global_position_min_max_count
```
ranchodeluxe, line 481: Isn't the assumption of `bucket_by_position`  that we are always dealing with contiguous data? If true then wouldn't `count` always be the same as the `max` position?

pangeo_forge_recipes/transforms.py
```diff
@@ -443,37 +435,89 @@ class CombineReferences(beam.PTransform):
...
+            | "Bucket to preserve order"
+            >> beam.Map(
+                self.bucket_by_position,
+                global_position_min_max_count=beam.pvalue.AsSingleton(min_max_count_positions),
+            )
+            | "Group by buckets for ordering" >> beam.GroupByKey()
+            | "Distributed reduce" >> beam.MapTuple(lambda k, refs: self.to_mzz(refs).translate())
+            | "Assign global key for collecting to executor" >> beam.Map(lambda ref: (None, ref))
+            | "Group globally" >> beam.GroupByKey()
+            | "Global reduce" >> beam.MapTuple(lambda k, refs: self.global_combine_refs(refs))
```
abarciauskas-bgse, line 520: So just for my own understanding, what I think is happening is this:
1. In the case of grib, merge multiple references so we start from a flat list of references
2. The next map operation buckets the references while preserving their order, such that each bucket will have a set of ordered references, ordered by the last `concat_dim` dimension.
3. The first GroupByKey ensures the order of the buckets is preserved.
4. Create a multizarrtozarr ref for each bucket (`to_mzz`)
5. This part I'm less clear about, from what I can tell we're assigning a new key to each group - but is each group a group of single references or merged references?
6. Merge everything and write to the target.

pangeo_forge_recipes/transforms.py
```diff
@@ -443,37 +435,89 @@ class CombineReferences(beam.PTransform):
...
+        return (
+            reference_lists
+            | "Handle special case of gribs" >> beam.Map(self.handle_gribs)
+            | "Bucket to preserve order"
+            >> beam.Map(
+                self.bucket_by_position,
+                global_position_min_max_count=beam.pvalue.AsSingleton(min_max_count_positions),
+            )
+            | "Group by buckets for ordering" >> beam.GroupByKey()
+            | "Distributed reduce" >> beam.MapTuple(lambda k, refs: self.to_mzz(refs).translate())
```
abarciauskas-bgse, line 517: I'm still curious if we know that it's faster to do this intermediate `to_mzz` call - if it is then I'm fine with leaving this part in but it seems to add complexity since we have to re-group by key.

pangeo_forge_recipes/transforms.py
ranchodeluxe: just over here trying to see if we can simply the last couple operations and leave out `range_size`

pangeo_forge_recipes/transforms.py
moradology: I don't have specific analysis in the beam context at hand but it appears to parallelize pretty well from what I've seen: https://github.com/fsspec/kerchunk/issues/200#issuecomment-1220547197
At this point the shuffle isn't so bad; mostly just passing around some json so I would expect the groupby to be pretty snappy. We could and perhaps should consider moving the groupbykey logic up to the point of Index/URL `PCollection` creation, where the shuffle would be extremely minimal but that may require a bit of thought in terms of what kind of flexibility we want there. At some point, I'll wager we will have to come up with a few alternative data localization strategies - perhaps even to get performant pyramiding?

pangeo_forge_recipes/transforms.py
moradology: The global reduce here is kind of weird, I agree. I would love a `beam.Collect` or something. So it goes. What we have to do, so far as I can tell, is to assign *the same key* to all records if we want to ensure that they are realized together on one executor

pangeo_forge_recipes/transforms.py
moradology: If` we assume that the index starts at 0, count would be max+1, but that quibble aside you're right (if the assumption holds true). Might be sensible to just eat the very slight count cost though to keep it more general and perhaps to sanity check results (e.g by comparing max-min to count)?

ranchodeluxe: okay, right, maybe we just `assert` the assumption somewhere either here or in `MinMaxCountCombiner`?

pangeo_forge_recipes/transforms.py
ranchodeluxe: > Merge everything and write to the target.

Just want to point out that nothing is being written in `global_combine_refs` just setting up an `fsspec` target for the next transform that does the writing

abarciauskas-bgse: > assign the same key to all records (None here) and then beam.GroupByKey if we want to ensure that they are realized together on one executor

Thanks that is super helpful explanation. In that case how are we assured that the final combine is done in order on the last single executor?

moradology: Experiments I've run locally w/ `multiprocessing` appear to demonstrate that the list fed to `MultiZarrToZarr` can have any ordering at all so long as there exists some ordering of the elements that would be contiguous. It is when you have separate `mzz` calls that you run into the risk of one of the sub-lists not having contiguity among its would-be merged references
Messing with ordering here should illustrate the behavior: https://gist.github.com/moradology/0e207a7ea4afd7251b0df604ac10958e#file-wtf-py-L69-L71

pangeo_forge_recipes/transforms.py
abarciauskas-bgse: That video is 💯

moradology: And damned relevant for building indexes of arbitrary precision (e.g. if you're trying to query data for processing 'zoom levels' on a pyramid)!
https://opensource.googleblog.com/2008/08/uzaygezen-multi-dimensional-indexing.html
https://link.springer.com/referenceworkentry/10.1007/978-0-387-39940-9_350
https://tildesites.bowdoin.edu/~ltoma/teaching/cs3225-GIS/fall17/Lectures/gis-spaceFC.pdf
https://www.grin.com/document/322063?lang=en
https://github.com/locationtech/sfcurve

And some prior art in roughly our space:
https://github.com/locationtech/geotrellis/blob/af71a4ae027bfeaa9a930bc086ca98186e701953/store/src/main/scala/geotrellis/store/index/hilbert/HilbertSpatialKeyIndex.scala#L49

ranchodeluxe (approved):

PR #691, opened 2024-02-08
comment out precombine_inputs

abarciauskas-bgse:


moradology:
I'm wondering if we can get away from pre_combine_inputs/bool rot here without the flatmap to iron out the unique case of GRIB2 outputs returning potentially more than one zarr reference

abarciauskas-bgse:
Closing, superceded by #689

PR #692, opened 2024-02-09
Add test for running in parallel

abarciauskas-bgse:


tests/conftest.py
```diff
@@ -235,6 +242,19 @@ def pipeline(scope="session"):
         yield p
 
 
+@pytest.fixture
+def pipeline_parallel(scope="session"):
+    # TODO: make this True and fix the weird ensuing type check errors
+    options = PipelineOptions(
+        runtime_type_check=False,
+        direct_num_workers=4,
```
abarciauskas-bgse, line 250: Using direct_num_workers was the key here.

tests/test_end_to_end.py
```diff
@@ -122,6 +122,47 @@ def test_reference_netcdf(
...
+    # 25 keys expected for netcdf_local_paths_sequential_multivariable_1d:
+    # .zgroup + .zattrs + (3 dimensions * (.zattrs + .zarray + 1 chunk)) + \
+    #   (2 variables * (.zattrs + .zarray + 5 chunks))
+    # 35 keys expected for netcdf_local_paths_sequential_multivariable_2d
+    file_ext = os.path.splitext(output_file_name)[-1]
+
+    if file_ext == ".json":
+        mapper = fsspec.get_mapper("reference://", fo=full_path)
+        ds = xr.open_dataset(mapper, engine="zarr", backend_kwargs={"consolidated": False})
+        xr.testing.assert_equal(ds.load(), daily_xarray_dataset)
```
abarciauskas-bgse, line 163: This is expected to fail right now, until merged with npz/feature/simplify_zarr_merge

tests/conftest.py
```diff
@@ -129,7 +129,14 @@ def make_local_paths(
...
     for ds, path in zip(datasets, full_paths):
         save_method = getattr(ds, method)
-        save_method(path, **kwargs)
+        # Question: Should we enable explicity chunking when saving to netcdf?
+        # by default it is encoding data with chunks, such as "chunks": [2,18,36].
+        # It might be useful to set chunking explicitly.
+        # encoding = {
+        #     'foo': {'chunksizes': (1, 2, 2)},
+        #     'bar': {'chunksizes': (1, 2, 2)}
+        # }
```
abarciauskas-bgse, line 138: ⬆️

tests/test_end_to_end.py
```diff
@@ -122,6 +122,47 @@ def test_reference_netcdf(
...
+                target_root=tmp_target,
+                store_name=store_name,
+                concat_dims=["time"],
+                output_file_name=output_file_name,
+            )
+        )
+    full_path = os.path.join(tmp_target.root_path, store_name, output_file_name)
+
+    # Question: Should we add a test here for the number of keys in the reference
+    # or is that moot since the xarray dataset equality check is already in place?
```
abarciauskas-bgse, line 150: ⬆️

tests/conftest.py
ranchodeluxe: I don't have strong opinion either way. I guess if it helps this immediate test case then yes, otherwise maybe just leave it as is.

tests/test_end_to_end.py
ranchodeluxe: yeah, I assume the equality check is handling that already

tests/test_end_to_end.py
ranchodeluxe: Nice! Is there anyway we can change the target of this to be @moradology PR https://github.com/pangeo-forge/pangeo-forge-recipes/pull/689 and just merge it into there since he's trying to write tests already?

tests/conftest.py
abarciauskas-bgse: I started to right up an issue about this, to create a test that explicitly tests for chunks in the origin netcdf being replicated in destination kerchunk, but I'm not sure if that's a test more appropriate for kerchunk or pangeo-forge-recipes 🤔

tests/test_end_to_end.py
abarciauskas-bgse: Thanks @moradology let me know when that's ready. I've removed obsolete comments from this PR so it should be good to merge

tests/conftest.py
moradology: Sounds like a kerchunk test, as you suggest. Might be useful for us to figure out the next steps around merge-func disambiguation over there, as things might make more sense at that point https://github.com/fsspec/kerchunk/issues/416#issuecomment-1937817235

moradology (approved):
This is really good. puts to shame my local attempts to reproduce via a `multiprocessing` simulation of the run 😭 .

I rebased the branch this is to aim at so perhaps introduced merge conflicts. will sort those out before pulling the trigger here

PR #694, opened 2024-02-14
Use MapTuple, put higher-order index type in `types` file

moradology:
Saw a TODO that was absolutely the right path forward regarding `_add_keys`: don't use it because it is more confusing to modify functions rather than simply compose functions via MapTuple

This also moves the higher order type `Indexed` into types as that is likely to find reuse (elsewhere, as it did in this PR)


cisaacstern:
This is looking great, thanks @moradology 🎉

ranchodeluxe:
oh yeah, like how readable `beam.MapTuple` makes this 🥳

PR #697, opened 2024-02-16
Tests to illustrate `CombineReferences.bucket_by_position`

ranchodeluxe:


ranchodeluxe:
I think I'm starting to understand the contiguous assumption here by writing these tests 👍

PR #700, opened 2024-02-19
Remove add_keys_iter

moradology:
This PR simply removes _add_keys_iter in favor of function composition


ranchodeluxe (approved):
yep, works

PR #669, opened 2024-01-17
[WIP - DRAFT] Ndpyramid

norlandrhagen:
Draft PR to explore generating pyramids via pangeo-forge-recipes. Uses [ndpyarmid](https://github.com/carbonplan/ndpyramid).

PR #711, opened 2024-03-18
Retain original _FillValue in encoding

maxrjones:
This is a follow-up to https://github.com/pangeo-forge/pangeo-forge-recipes/issues/471 that adds `_FillValue` to the retained encoding options. Based on https://github.com/pangeo-forge/pangeo-forge-recipes/pull/471#discussion_r1072218998, `_FillValue` was initially excluded because it causes the tests to fail when asserting that the expected and actual schemas are equal. That fails because it's common to specify `_FillValue` as `np.nan` and nans do not equal each other. This PR adds a `_assert_schema_equal` that will treat NaNs in the schema as equal, so that the original `_FillValue` can be retained.

This PR is a prerequisite to #669 because we need to make all encoding options stored in `.zmetadata` follow [valid JSON](https://datatracker.ietf.org/doc/html/rfc8259).

### DIFF
--- a/pangeo_forge_recipes/aggregation.py
+++ b/pangeo_forge_recipes/aggregation.py
@@ -147,7 +147,12 @@ def _combine_attrs(a1: dict, a2: dict) -> dict:
     common_attrs = set(a1) & set(a2)
     new_attrs = {}
     for key in common_attrs:
-        if isinstance(a1[key], np.floating) and isinstance(a2[key], np.floating) and np.isnan(a1[key]) and np.isnan(a2[key]):
+        if (
+            isinstance(a1[key], np.floating)
+            and isinstance(a2[key], np.floating)
+            and np.isnan(a1[key])
+            and np.isnan(a2[key])
+        ):
             new_attrs[key] = a1[key]
         elif a1[key] == a2[key]:
             new_attrs[key] = a1[key]
--- a/pangeo_forge_recipes/aggregation.py
+++ b/pangeo_forge_recipes/aggregation.py
@@ -31,7 +31,6 @@ def dataset_to_schema(ds: xr.Dataset) -> XarraySchema:
     # Remove redundant encoding options
     for v in ds.variables:
         for option in ["source"]:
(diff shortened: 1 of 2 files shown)


pangeo_forge_recipes/aggregation.py
```diff
@@ -30,7 +30,7 @@ def dataset_to_schema(ds: xr.Dataset) -> XarraySchema:
 
     # Remove redundant encoding options
     for v in ds.variables:
-        for option in ["_FillValue", "source"]:
+        for option in ["source"]:
             # TODO: should be okay to remove _FillValue?
```
norlandrhagen, line 34: Should we remove this comment?

maxrjones: good call!

pangeo_forge_recipes/aggregation.py
```diff
@@ -147,7 +147,14 @@ def _combine_attrs(a1: dict, a2: dict) -> dict:
     common_attrs = set(a1) & set(a2)
     new_attrs = {}
     for key in common_attrs:
-        if a1[key] == a2[key]:
+        if (
```
norlandrhagen, line 150: Maybe it's worth adding a comment about the change and what we're checking for?

norlandrhagen (approved):

cisaacstern:
Thanks @maxrjones! Based on discussion following https://github.com/pangeo-forge/pangeo-forge-recipes/pull/471#discussion_r1072191103, it does not seem as though there was a specific justification for excluding `_FillValue` initially, aside from the testing issues resolved by this PR. This track with what you shared in the meeting today! So in principle, this is good by me. I'll just wait to make sure tests pass.

cisaacstern:
Thanks! @norlandrhagen please feel free to release per https://pangeo-forge.readthedocs.io/en/latest/contributing.html#releasing if it's useful for you all!

maxrjones:
thanks for the prompt reviews @cisaacstern and @norlandrhagen!

PR #712, opened 2024-03-19
Remove confusing docstring for consolidating coordinate dimensions

jbusecke:
The docstring did not reflect the new Consolidation Strategy (via separate stages rather than in StoreToZarr)


norlandrhagen:
Nice catch!

PR #718, opened 2024-03-27
Crush warning about ds.dims deprecation

jbusecke:
Got a bunch of these:
```
 FutureWarning: The return type of `Dataset.dims` will be changed to return a set of dimension names in future, in order to be more consistent with `DataArray.dims`. To access a mapping from dimension names to lengths, please use `Dataset.sizes`.
```
Hopefully this will be an easy fix

### DIFF
--- a/pangeo_forge_recipes/rechunking.py
+++ b/pangeo_forge_recipes/rechunking.py
@@ -57,7 +57,7 @@ def split_fragment(
             dimsize = getattr(index[concat_dim], "dimsize", 0)
             concat_position = index[concat_dim]
             start = concat_position.value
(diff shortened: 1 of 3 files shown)


norlandrhagen:
Does [this](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/714) cover this @jbusecke ?

jbusecke:
yes yes it does. my bad...

jbusecke:
Actually I think this PR removes use of ds.dims globally (there are a few occasions where it is used).

jbusecke:
but still sorry for not checking before.

moradology (approved):

Issue #447, opened 2022-12-05
Making appending work in the beam refactor


rabernat:
This issue is for @alxmrs, who indicated some willingness to work on it.

In order to append to existing Zarr datasets, we need to update the [StoreToZarr](https://pangeo-forge--445.org.readthedocs.build/en/445/pangeo_forge_recipes/api_reference.html#pangeo_forge_recipes.transforms.StoreToZarr) PTransform such that it can handle an existing dataset at the `target_url`.

That PTransform first figures out the schema of the input data, then pass this to the `PrepareZarrTarget` schema:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/0e16f15155db31dd13e448c6620e872b158be4b1/pangeo_forge_recipes/transforms.py#L205-L227

which ultimately calls the function `schema_to_zarr`

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/0e16f15155db31dd13e448c6620e872b158be4b1/pangeo_forge_recipes/aggregation.py#L233-L242

One way to implement this would be to add an `append` option to this  function. If active, we could try to first open the target zarr dataset and see if it is compatible with the schema (e.g. same variables and dimensions [other than the concat dim]). If so, we could skip initializing / overwriting it. Instead, we would simply have to resize the arrays (e.g. using the zarr [resize](https://zarr.readthedocs.io/en/stable/tutorial.html#resizing-and-appending) function). The original zarr dataset size would be used to determine an offset for future writes. (Note there is a tricky edge case if the offset does not evenly divide by the target chunk size.)

The other change that would have to be made is in `store_dataset_fragment`

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/0e16f15155db31dd13e448c6620e872b158be4b1/pangeo_forge_recipes/writers.py#L54-L76

We would need some way to pass the offset (determined in `schema_to_zarr` as described above) through the pipeline so that the new data is written at the correct location.

DarshanSP19:
@rabernat The first step is fine to add an option `append` and verify the dimensions and variables. For the other half can we just append the arrays with new data. zarr has already a function for appending the data. [Here](https://zarr.readthedocs.io/en/stable/tutorial.html#resizing-and-appending). Can this be an approach to consider?

alxmrs:
[...]

PR #721, opened 2024-03-29
Append mode for StoreToZarr

cisaacstern:
Towards #447

This first pass focuses just on `StoreToZarr` (not reference recipes).

Will open for review once it looks like it's working.

### DIFF
--- a/pangeo_forge_recipes/aggregation.py
+++ b/pangeo_forge_recipes/aggregation.py
@@ -290,11 +290,16 @@ def schema_to_zarr(
     append_dim: Optional[str] = None,
 ) -> zarr.storage.FSStore:
     """Initialize a zarr group based on a schema."""
+    if append_dim:
+        # if appending, only keep schema for coordinate to append. if we don't drop other
+        # coords, we may end up overwriting existing data on the `ds.to_zarr` call below.
+        schema["coords"] = {k: v for k, v in schema["coords"].items() if k == append_dim}
     ds = schema_to_template_ds(schema, specified_chunks=target_chunks, attrs=attrs)
-    # using mode="w" makes this function idempotent
+    # using mode="w" makes this function idempotent when not appending
     ds.to_zarr(
         target_store,
         append_dim=append_dim,
+        mode=("a" if append_dim else "w"),
         compute=False,
         consolidated=consolidated_metadata,
         encoding=encoding,
--- a/pangeo_forge_recipes/aggregation.py
+++ b/pangeo_forge_recipes/aggregation.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from copy import deepcopy
 from dataclasses import dataclass, field
-from typing import Dict, Literal, Optional, TypedDict
+from typing import Dict, Optional, TypedDict
 
 import cftime
 import dask.array as dsa
@@ -287,14 +287,14 @@ def schema_to_zarr(
     attrs: Optional[Dict[str, str]] = None,
     consolidated_metadata: Optional[bool] = True,
     encoding: Optional[Dict] = None,
-    mode: Literal["w", "a"] = "w",
+    append_dim: Optional[str] = None,
 ) -> zarr.storage.FSStore:
     """Initialize a zarr group based on a schema."""
     ds = schema_to_template_ds(schema, specified_chunks=target_chunks, attrs=attrs)
     # using mode="w" makes this function idempotent
     ds.to_zarr(
         target_store,
-        mode=mode,
+        append_dim=append_dim,
         compute=False,
         consolidated=consolidated_metadata,
         encoding=encoding,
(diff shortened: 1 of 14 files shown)


rabernat:
I would recommend using `append_dim: str | None` instead of `mode`.

cisaacstern:
This now works! The best "documentation" of it so far is the [end to end test](https://github.com/pangeo-forge/pangeo-forge-recipes/pull/721/files#diff-a4a614c9bc88758962ff69a4056481f9f9ae8930d92e4ce1ad8daab54a452b30).

Tagged a bunch of you for review, for visibility. Also @thodson-usgs since I know you're interesting in this feature.

To state the (presumably) obvious, this minimal implementation requires that the user/deployer know what range of data should be appended to the existing dataset, and does not provide any idempotency guarantees or much in the way of logical checks. It will happily append time ranges that do not monotonically increase from the existing concat dim, or re-append the same data multiple twice.

But because resizing of the contact dimension is done with `xarray.Dataset.to_zarr`, we _**do**_ get the benefit of [some built-in consistency checks](https://github.com/pydata/xarray/blob/6af547cdd9beac3b18420ccb204f801603e11519/xarray/backends/api.py#L1828-L1842) there.

pangeo_forge_recipes/aggregation.py
```diff
@@ -287,13 +287,19 @@ def schema_to_zarr(
     attrs: Optional[Dict[str, str]] = None,
     consolidated_metadata: Optional[bool] = True,
     encoding: Optional[Dict] = None,
+    append_dim: Optional[str] = None,
```
norlandrhagen, line 290: Is it worth adding some validation to see if this append_dim (str) is a valid dataset dimension?

pangeo_forge_recipes/transforms.py
```diff
@@ -657,17 +665,31 @@ class StoreToZarr(beam.PTransform, ZarrWriterMixin):
...
+    append_dim: Optional[str] = None
 
     def __post_init__(self):
         if self.target_chunks and self.dynamic_chunking_fn:
             raise ValueError("Passing both `target_chunks` and `dynamic_chunking_fn` not allowed.")
 
+        self._append_offset = 0
+        if self.append_dim:
+            logger.warn(
+                "When `append_dim` is given, StoreToZarr is NOT idempotent. Successive deployment "
```
norlandrhagen, line 677: In future PR's to this, do you see a good path forward to making append safe/idempotent?

norlandrhagen (approved):
Awesome work Charles! It's great to see this feature added. Do you plan to add a section to the docs?

pangeo_forge_recipes/transforms.py
cisaacstern: Good question! For future readers, we had a good discussion of this in today's Coordination meeting, [minutes here](url).

pangeo_forge_recipes/aggregation.py
cisaacstern: Yes! I do that in StoreToZarr.__post_init__. Also, xarray.Dataset.to_zarr does that internally, I believe.

cisaacstern:
> Do you plan to add a section to the docs?

@norlandrhagen how does https://github.com/pangeo-forge/pangeo-forge-recipes/pull/721/commits/aaa5a694ef3371dd92d2dbf6f96846a6fa199834 seem? Rendered version [here](https://pangeo-forge--721.org.readthedocs.build/en/721/composition/styles.html)

pangeo_forge_recipes/aggregation.py
```diff
@@ -287,13 +287,19 @@ def schema_to_zarr(
...
 ) -> zarr.storage.FSStore:
     """Initialize a zarr group based on a schema."""
+    if append_dim:
+        # if appending, only keep schema for coordinate to append. if we don't drop other
+        # coords, we may end up overwriting existing data on the `ds.to_zarr` call below.
+        schema["coords"] = {k: v for k, v in schema["coords"].items() if k == append_dim}
```
cisaacstern, line 296: Just realizing that this, while seemingly necessary to avoid overwriting non-`append_dim` coords, may prevent us from benefiting from the dimension consistency checks offered by `ds.to_zarr`... we may need to re-implement such checks in `StoreToZarr.__post_init__` if we want them.

cisaacstern: Thinking about this further... in `StoreToZarr.__post_init__` we actually don't know what the coordinates in the new data are (aside from the user-provided `append_dim`), because we haven't actually fetched and opened any source files yet.

So if we want to check for consistency, we could actually do it right here. Before dropping non-append coords from the schema, we would just need to open the target dataset and check that all coords match first. Could be a good idea.

cisaacstern: Having just reviewed [xarray's checks](https://github.com/pydata/xarray/blob/6af547cdd9beac3b18420ccb204f801603e11519/xarray/backends/api.py#L1828-L1842) again, I actually don't see that this is done there, so may be overly strict. I think we can move forward with what we have for now.

PR #723, opened 2024-04-02
Move logic from schema accumulator object to combinefn

moradology:
The schema accumulator object is basically a tuple. Its methods duplicate the intent of CombineFn's methods and this doesn't seem necessary.

These changes surface the reduce behavior of a combineFn in terms of pure functions. The tests will reflect this as now results from prior combination operations have to be threaded through manually (beam handles this in the normal context)


tests/test_aggregation.py
```diff
@@ -108,85 +106,3 @@ def test_schema_to_template_ds_attrs():
 
     assert dst.attrs["pangeo-forge:test_attr_key"] == "test_attr_value"
     assert dst.attrs["original_attrs_key"] == "original_attrs_value"
-
-
-def test_concat_accumulator():
```
ranchodeluxe, line 113: Your changes above make sense but I see you are removing tests here. Should we be keeping some of these?

moradology: Yeah, that's doable directly via the combinefn. Should be fairly easy...

PR #726, opened 2024-04-04
Organize, document, and simplify transforms.py

moradology:
What I've attempted to do here:
1. [Unified a type union here](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/main/pangeo_forge_recipes/transforms.py#L693-L701). This runtime bug is actually what got me started on this PR. The fix for this is turning compile-time values into beam.pvalues which can then be used as side inputs as though they were, themselves, computed by the cluster. Where sensible, I've attempted to logically separate the provisioning of compile-time and runtime arguments. This takes the form of pushing side-inputs (runtime) into the `expand` method and constructing classes with compile-time values.
2. Broke out some transforms that were little more than beam.Map calls into their functions. We can simply call beam.Map in the higher-level transform. Should make reuse more flexible and testing easier (not that I think we need more tests honestly)
3. Added documentation where I thought it would be helpful. renamed functions to better clarify their intent.
4. Reduced function surface area in storetozarr and documented how a user can write equivalent dynamic chunking functions

### DIFF
--- a/pangeo_forge_recipes/patterns.py
+++ b/pangeo_forge_recipes/patterns.py
@@ -63,12 +63,13 @@ class MergeDim(CombineDim):
     operation: ClassVar[CombineOp] = CombineOp.MERGE
 
 
-def augment_index_with_start_stop(
+def augment_index_with_byte_range(
     position: Position,
     item_lens: List[int],
     append_offset: int = 0,
 ) -> IndexedPosition:
-    """Take an index _without_ start / stop and add them based on the lens defined in sequence_lens.
+    """Take an index _without_ start / stop (byte range) and add them based on the lens defined in
+        sequence_lens.
 
     :param index: The ``DimIndex`` instance to augment.
     :param item_lens: A list of integer lengths for all items in the sequence.
--- a/pangeo_forge_recipes/transforms.py
+++ b/pangeo_forge_recipes/transforms.py
@@ -20,10 +20,10 @@ import xarray as xr
 import zarr
 from kerchunk.combine import MultiZarrToZarr
 
-from .aggregation import XarraySchema, dataset_to_schema, schema_to_template_ds, schema_to_zarr
+from .aggregation import dataset_to_schema, schema_to_template_ds, schema_to_zarr
 from .combiners import CombineXarraySchemas, MinMaxCountCombineFn
 from .openers import open_url, open_with_kerchunk, open_with_xarray
-from .patterns import CombineOp, Dimension, FileType, Index, augment_index_with_start_stop
+from .patterns import CombineOp, Dimension, FileType, Index, augment_index_with_byte_range
 from .rechunking import combine_fragments, consolidate_dimension_coordinates, split_fragment
 from .storage import CacheFSSpecTarget, FSSpecTarget
 from .types import Indexed
@@ -241,31 +241,18 @@ class OpenWithXarray(beam.PTransform):
         )
 
 
-def _nest_dim(item: Indexed[T], dimension: Dimension) -> Indexed[Indexed[T]]:
-    index, value = item
-    inner_index = Index({dimension: index[dimension]})
-    outer_index = Index({dk: index[dk] for dk in index if dk != dimension})
-    return outer_index, (inner_index, value)
-
-
-@dataclass
-class _NestDim(beam.PTransform):
-    """Prepare a collection for grouping by transforming an Index into a nested
-    Tuple of Indexes.
-
-    :param dimension: The dimension to nest
-    """
-
-    dimension: Dimension
-
-    def expand(self, pcoll):
-        return pcoll | beam.Map(_nest_dim, dimension=self.dimension)
-
-
 @dataclass
 class DatasetToSchema(beam.PTransform):
     def expand(self, pcoll: beam.PCollection) -> beam.PCollection:
         return pcoll | beam.MapTuple(lambda k, v: (k, dataset_to_schema(v)))
+
+
+def _nest_dim(item: Indexed[T], dimension: Dimension) -> Indexed[Indexed[T]]:
+    """Nest dimensions to support multiple combine dimensions"""
+    index, value = item
+    inner_index = Index({dimension: index[dimension]})
+    outer_index = Index({dk: index[dk] for dk in index if dk != dimension})
+    return outer_index, (inner_index, value)
 
 
 @dataclass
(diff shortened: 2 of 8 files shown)


pangeo_forge_recipes/patterns.py
```diff
@@ -63,12 +63,13 @@ class MergeDim(CombineDim):
     operation: ClassVar[CombineOp] = CombineOp.MERGE
 
 
-def augment_index_with_start_stop(
+def augment_index_with_byte_range(
```
cisaacstern, line 66: I believe this is indexing logical dimensions in the dataset, not bytes.

pangeo_forge_recipes/patterns.py
```diff
@@ -63,12 +63,13 @@ class MergeDim(CombineDim):
...
     position: Position,
     item_lens: List[int],
     append_offset: int = 0,
 ) -> IndexedPosition:
-    """Take an index _without_ start / stop and add them based on the lens defined in sequence_lens.
+    """Take an index _without_ start / stop (byte range) and add them based on the lens defined in
```
cisaacstern, line 71: Same comment as above re: bytes.

pangeo_forge_recipes/transforms.py
```diff
@@ -394,22 +356,24 @@ def expand(self, pcoll: beam.PCollection) -> beam.PCollection:
...
+            >> beam.FlatMap(
                 split_fragment,
-                target_chunks=self.target_chunks,
-                schema=beam.pvalue.AsSingleton(self.schema),
+                target_chunks=target_chunks,
+                schema=schema,
             )
-            | beam.GroupByKey()  # this has major performance implication
-            | beam.MapTuple(combine_fragments)
+            | "group by write chunk key" >> beam.GroupByKey()  # group by key ensures locality
```
cisaacstern, line 374: Nice! this is very helpful developer commentary 😃

pangeo_forge_recipes/transforms.py
```diff
@@ -647,22 +611,37 @@ class StoreToZarr(beam.PTransform, ZarrWriterMixin):
...
+            target_chunk_size = 1024 * 1024 * 10
+            dim_name = 'time'
+            return calculate_dynamic_chunks(ds, target_chunk_size, dim_name)
+
+        StoreToZarr(..., dynamic_chunking_fn=dynamic_chunking_wrapper, ...)
+        ```
     """
 
-    # TODO: make it so we don't have to explicitly specify combine_dims
-    # Could be inferred from the pattern instead
```
cisaacstern, line 657: I guess we're abandoning this idea? I have had my doubts that it's possible, since it would be a meta-operation over the pipeline. But i've been wrong about what's possible in beam before.

pangeo_forge_recipes/transforms.py
```diff
@@ -688,31 +667,51 @@ def expand(
...
+        target_chunks_pcoll = (
+            pipeline | "Create target_chunks pcoll" >> beam.Create([self.target_chunks])
             if not self.dynamic_chunking_fn
-            else beam.pvalue.AsSingleton(
+            else (
                 schema
-                | beam.Map(schema_to_template_ds)
-                | beam.Map(self.dynamic_chunking_fn, **self.dynamic_chunking_fn_kwargs)
+                | "make template dataset" >> beam.Map(schema_to_template_ds)
+                | "generate chunks dynamically" >> beam.Map(self.dynamic_chunking_fn)
```
cisaacstern, line 691: these labels really help readability! tysm!

pangeo_forge_recipes/transforms.py
```diff
@@ -688,31 +667,51 @@ def expand(
...
+        ).expand(schema, beam.pvalue.AsSingleton(target_chunks_pcoll))
+
+        # Actually attempt to write datasets to their target bytes/files
+        rechunked_datasets | "write chunks" >> beam.Map(
+            store_dataset_fragment, target_store=beam.pvalue.AsSingleton(target_store)
         )
 
-        return singleton_target_store
+        # return the target store pcollection (a singleton of the fsspec target)
+        return target_store
```
cisaacstern, line 717: The reason we did this `n_target_stores` -> `FixedSizeGlobally(1)` nonsense before (and sorry for not documenting this in comments or otherwise!) ... was that without doing this, Beam will not wait on all chunks to be finished writing before yielding back the target store... and in this case, I'm pretty sure we do want to enforce blocking behavior because if we don't, then if we try to do:

```python
... | StoreToZarr() | ConsolidateMetadata()
```

```python
... | StoreToZarr() | ValidateStoredZarr()
```

or what-have-you, then the transform that is chained onto `StoreToZarr` will receive it's input before chunks have finished writing. Which will make it impossible to consolidate or validate, etc.

So basically, `n_target_stores` -> `FixedSizeGlobally(1)` ensures we block, and don't let the output of `StoreToZarr` be emitted downstream until all chunks are written.

Does that make sense?

pangeo_forge_recipes/transforms.py
```diff
@@ -647,22 +611,37 @@ class StoreToZarr(beam.PTransform, ZarrWriterMixin):
...
-    # TODO: make it so we don't have to explicitly specify combine_dims
-    # Could be inferred from the pattern instead
     combine_dims: List[Dimension]
     store_name: str
     target_root: Union[str, FSSpecTarget, RequiredAtRuntimeDefault] = field(
         default_factory=RequiredAtRuntimeDefault
     )
     target_chunks: Dict[str, int] = field(default_factory=dict)
     dynamic_chunking_fn: Optional[Callable[[xr.Dataset], dict]] = None
-    dynamic_chunking_fn_kwargs: Optional[dict] = field(default_factory=dict)
```
cisaacstern, line 665: Is there an advantage to lower arity here aside from an aesthetically tidier signature?

cisaacstern (commented):
@moradology very awesome insight re: type union, and elegant fixes! I'll freely confess some of the fixes fly above my current level of understanding of Beam composition, but generally feel quite elegant, and if the tests pass, they pass!

Made a few comments, one of which is just a gut check on function naming (the bytes thing)... and the only other of real importance I think is the question about blocking behavior in `StoreToZarr`.

Really inspiring work! I always learn something from reading your code! Thanks so much.

pangeo_forge_recipes/transforms.py
```diff
@@ -647,22 +611,37 @@ class StoreToZarr(beam.PTransform, ZarrWriterMixin):
...
+        def calculate_dynamic_chunks(ds, target_chunk_size, dim_name) -> Dict[str, int]:
+            ...
+
+        def dynamic_chunking_wrapper(ds: xarray.Dataset) -> Dict[str, int]:
+            target_chunk_size = 1024 * 1024 * 10
+            dim_name = 'time'
+            return calculate_dynamic_chunks(ds, target_chunk_size, dim_name)
+
+        StoreToZarr(..., dynamic_chunking_fn=dynamic_chunking_wrapper, ...)
+        ```
```
cisaacstern, line 635: Looks like this not rendering correctly. Not sure what your intended design is here, but maybe the paragraph starting with `Suppose..` should be dedented and the ```python fence is not necessary if you're indented?

See https://pangeo-forge--726.org.readthedocs.build/en/726/api_reference.html#ptransforms
Screen Shot 2024-04-04 at 4 05 58 PM

moradology:
I should have called this out in the PR description, but I somehow failed to include it. The one piece I'm not yet sure about is this small change to tests - the argument works as expected but it isn't obvious to me why that's now required. Before +1, I'll need to double-check expected behavior

https://github.com/pangeo-forge/pangeo-forge-recipes/pull/726/files#diff-a4a614c9bc88758962ff69a4056481f9f9ae8930d92e4ce1ad8daab54a452b30R277

pangeo_forge_recipes/patterns.py
moradology: It looks to me as though this function uses the logical position of the provided index to calculate start/stop in terms of bytes, no? Like, the sum of item lengths calculated prior to it is determined based on logical position but the resulting integer is supposed to be the byte-range start is how i read it

pangeo_forge_recipes/transforms.py
moradology: The way this reads to me right now, we'd need to pass the same argument to a couple of different transforms (which isn't a bad pattern afaict). To reuse such a value, we'd need to thread the input through as some kind of computed result and I doubt that juice is worth the squeeze

I'm not opposed to keeping the TODO open, if that's desirable. I just couldn't quite imagine how this could be facilitated at the level of recipes

pangeo_forge_recipes/transforms.py
moradology: Great catch! Makes total sense

pangeo_forge_recipes/transforms.py
moradology: The tidy signature is much of the benefit I see but also hopefully passing responsibility downstream to pipeline writers/users in a way that can't hide any magic we might be applying. On this conception of things, the user has one thing to worry about: a function that produces the chunking they desire given a template dataset

PR #742, opened 2024-05-01
Add pattern_from_file_sequence to docs

maxrjones:
Adds this function and updates a few other references based on recent changes to the code base


ranchodeluxe (approved):
Look good @maxrjones 🥳

ranchodeluxe:
I think pre-commit stuff can be ignored since it has to be fixed in a bot bump anyhow, will get to that later

ranchodeluxe:
@maxrjones flake8 issues are now fixed in `main` so if you rebase you won't have anymore pre-commit errors or you can just merge it as is!

PR #748, opened 2024-05-22
Optional support for synchronous HTTP filesystem

moradology:
This PR adds a flag that enables the experimental support for a synchronous alternative to the HTTP/S filesystems provided by FSSpec. Fairly extensive profiling with xarray reveals that this is quite likely slower than the default case but enabling it allows us to reduce the async/concurrency surface area of the code which may avoid or at least help debug potential deadlocking issues.

### DIFF
--- a/pangeo_forge_recipes/openers.py
+++ b/pangeo_forge_recipes/openers.py
@@ -18,6 +18,7 @@ def open_url(
     cache: Optional[CacheFSSpecTarget] = None,
     secrets: Optional[Dict] = None,
     open_kwargs: Optional[Dict] = None,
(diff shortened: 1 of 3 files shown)


norlandrhagen:
@moradology I just tried this locally with the pyspark runner and it worked great! 

One minor nit. Would it be possible to add some ImportError if `httpsfs_sync` is not installed and the `fsspec_sync_patch` is set to True. I got this message: `httpfs_sync could not be imported. Falling back to async http implementation.`, stopped the job and reinstalled.

ranchodeluxe (approved):
we need the juice!

PR #755, opened 2024-06-11
Remove log we don't need

### LINKED ISSUE #720: Revert changes made in #657
Some more deep digging into logs lately has convinced me that the changes I made in #657 are not working. 

I think we should remove this line:

https://github.com/pangeo-forge/pangeo-forge-recipes/blob/fe036507d728d10cf4c926fedcc5d0ca5234f7f3/pangeo_forge_recipes/transforms.py#L680

The target_chunks information is actually already being logged in [`split_fragments`](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/fe036507d728d10cf4c926fedcc5d0ca5234f7f3/pangeo_forge_recipes/rechunking.py#L35), but having the target store logged would still be valuable. Particularly if injections are used to modify this, the user should have a way to confirm **exactly where** data is written to.

moradology:
Closes #720

### DIFF
--- a/pangeo_forge_recipes/transforms.py
+++ b/pangeo_forge_recipes/transforms.py
@@ -704,7 +704,6 @@ class StoreToZarr(beam.PTransform, ZarrWriterMixin):
                 | beam.Map(self.dynamic_chunking_fn, **self.dynamic_chunking_fn_kwargs)
             )
         )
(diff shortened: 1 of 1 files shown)


jbusecke (approved):
LGTM

PR #761, opened 2024-06-27
Adds open zarr + rechunk example to feedstock

norlandrhagen:


norlandrhagen:
The new integration tests are passing, but the normal tests are still failing which should be failing on main as well.

PR #775, opened 2024-08-27
bump beam

norlandrhagen:
Fixes the version CI failures on main. cc @jbusecke 

Thanks to @keewis

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -49,7 +49,7 @@ jobs:
       - name: 🏄‍♂️ Run Tests
         shell: bash -l {0}
         run: |
-          py.test tests -v \
+          py.test tests -v -n auto \
             --cov=pangeo_forge_recipes --cov-config .coveragerc \
             --cov-report term-missing \
             --cov-report xml \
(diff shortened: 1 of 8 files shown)


.github/workflows/main.yaml
```diff
@@ -32,28 +32,25 @@ jobs:
...
-          cache-dependency-path: pyproject.toml
-      - name: 🎯 Check cache hit
-        run: echo '${{ steps.setup-python.outputs.cache-hit }}'
+      #     cache-dependency-path: pyproject.toml
+      # - name: 🎯 Check cache hit
+      #   run: echo '${{ steps.setup-python.outputs.cache-hit }}'
       - name: 🌈 Install pangeo-forge-recipes package
         shell: bash -l {0}
         run: |
           python -m pip install -e  ".[test]"
```
keewis, line 41:
```suggestion
          python -m pip install -e  ".[test]" "apache-beam>=2.54"
```

keewis:
that seems to have done it. So as a summary, we have to help `pip` to decide which beam version it should use by introducing a lower bound, and we're also incompatible with `kerchunk=0.2.6` (I tried `0.2.5` instead of `0.2.4`, which worked locally, as well)

norlandrhagen:
Credit for @keewis for fixing this!

PR #784, opened 2024-11-24
modernize the `pre-commit` hooks

keewis:
I'd nowadays use:
- `ruff` instead of `flake8`, `isort` (and `seed-isort-config`)
- the `black-pre-commit-mirror` repository to be a little bit faster (could use `ruff-format` as well, but I personally don't)
- `prettier` to format json, yaml, and markdown (and other web formats, but we don't seem to have any of those here)
- `taplo` to lint / format toml
- `validate-pyproject` to validate the `project` / `build-system` table in `pyproject.toml`

I didn't enable the `pyupgrade` / `relative-import` rules that I usually would, because this would mean a lot of additional changes. Do tell me if you'd be fine with those, though, and I'll switch them back on.

Oh, and I didn't upgrade `mypy` because there's a two errors and a bunch of notes – which I guess is what causes #764 to fail (and I don't know enough about type checking to fix those errors myself):
```
pangeo_forge_recipes/serialization.py:69: error: Argument 1 to "asdict" has incompatible type "DataclassInstance | type[DataclassInstance]"; expected "DataclassInstance"  [arg-type]
tests/test_serialization.py:99: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
pangeo_forge_recipes/openers.py:117: error: Incompatible types in assignment (expression has type "Any | IOBase | str", variable has type "str")  [assignment]
tests/test_transforms.py:263: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
tests/test_transforms.py:366: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
tests/test_end_to_end.py:224: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
tests/test_rechunking.py:67: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
```


keewis:
the test failures are caused by the upgrade of `kerchunk` from `0.2.5` to `0.2.7`, but otherwise this should be fine to merge (cc @norlandrhagen)

PR #786, opened 2025-02-12
[Draft] Removes Kerchunk recipe pipeline tools

norlandrhagen:
Optional PR that should simplify the test suite and maintenance burden. With Kerchunk's `auto-dask` feature and the work being done in VirtualiZarr this functionality seems much less useful. 

Wondering your thoughts @keewis


keewis:
for the record, the only concern I have here is backwards compatibility: if anyone was using this we'd be breaking them by removing this feature without warning.

PR #788, opened 2025-02-12
Updates to Zarr V3

norlandrhagen:
Branch to update pangeo-forge-recipes to Zarr v3. 
- [x] pinned Zarr and Xarray to comply with Zarr V3. 
- [x] fixed first of the zarr v2->v3 api changes.  (`zarr.storage.FSStore` -> `zarr.storage.FSSpecStore`)

cc @keewis

### DIFF
--- a/pangeo_forge_recipes/aggregation.py
+++ b/pangeo_forge_recipes/aggregation.py
@@ -253,13 +253,13 @@ def schema_to_template_ds(
 
 def schema_to_zarr(
     schema: XarraySchema,
-    target_store: zarr.storage.FSStore,
+    target_store: zarr.storage.FsspecStore,
     target_chunks: Optional[Dict[str, int]] = None,
     attrs: Optional[Dict[str, str]] = None,
     consolidated_metadata: Optional[bool] = True,
     encoding: Optional[Dict] = None,
     append_dim: Optional[str] = None,
-) -> zarr.storage.FSStore:
+) -> zarr.storage.FsspecStore:
     """Initialize a zarr group based on a schema."""
     if append_dim:
         # if appending, only keep schema for coordinate to append. if we don't drop other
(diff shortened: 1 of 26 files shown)


norlandrhagen:
WIP note: @keewis and I pushed on this. We're close, but currently blocked by: `TypeError: Filesystem needs to support async operations.`

keewis:
it looks like `"file"` and `"memory"` filesystems don't have a async version, so passing that as `fs` won't work.

I've tried using `AsyncFileSystemWrapper` to work around this, but that doesn't appear to properly create the files (I get `FileNotFoundError`s for various `zarr.json` files)

norlandrhagen:
Went down the exact same rabbit hole!

keewis:
I'm pretty close now, I think: only this weird encoding issue left

keewis:
it looks like the the source of the encoding value is `xarray`, which encodes floats as base64-strings. I would just remove the update of `encoding` here: https://github.com/pangeo-forge/pangeo-forge-recipes/blob/76e2a5b8d0e3ad724081ed9bf915dfaf671df611/pangeo_forge_recipes/writers.py#L36 but I'm not sure what its purpose is so I may very well be missing something.

pangeo_forge_recipes/aggregation.py
```diff
@@ -253,26 +253,29 @@ def schema_to_template_ds(
...
-    ds.to_zarr(
-        target_store,
-        append_dim=append_dim,
-        mode=("a" if append_dim else "w"),
-        compute=False,
-        consolidated=consolidated_metadata,
-        encoding=encoding,
-    )
+
+    with zarr.config.set({"array.write_empty_chunks": True}):
```
keewis, line 271: we'll probably have to check if `encoding` contains `write_empty_chunks`, but for now this makes the tests pass

keewis:
as discussed in the meeting just now I'll go ahead and merge this. This will mean that we're breaking backwards compatibility with people trying to install `zarr-python<3` together with PFR, so we'll have to figure out whether the reduced maintenance burden (i.e. no compat code) is worth the disruption.

PR #819, opened 2025-06-04
include warning in index.md

norlandrhagen:


keewis:
it looks like `prettier` does not like a admonition followed directly by the content, so we have to put a empty quote line in between.

I've also added a couple more warnings to hopefully make it clear that only `pangeo-forge-recipes` will be maintained, but ideally we'd also add a banner to the top of all pages.

